PR #3, opened 2020-02-10
Stats dashboard initial implementation

hakkane84:
This is an initial implementation of a dashboard for the stats page, working on top of the format of the mock stats API. The dashboard can be easily readapted to whichever format and information the final API has. It adds the `<Dashboard>` React component (`/src/components/Dashboard.tsx`) which is a responsive grid comprised of two types of elements:

* `StatBox` that showcase one specific stat, like number of pinned files or amount of stored data. Up to 3 boxes can be added per row, and more rows can be added if necessary
* `StatBarChart` that renders a single-series or multi-series bar chart rendered with the `D3.js` library (open source and free to use). Two charts fit per row, and as many rows with more charts can be added if required.

The color scheme is taken from 	`/theme/theme.ts` where I added a new color for the borders of the elements. The colors of the bar charts are taken from variating the brightness of the `primary` color automatically.

I attach an example image of how the default dashboard looks like, and a live example can be found on https://sialoop.net/stats (this uses a custom color set to fit the rest of the site).

This MR adds these additional dependencies to the portal:

* `react-grid-system`, which creates a responsive layout grid similar to Boostrap
* `recharts`, that allows using the `D3.js` library in React
* `@fortawesome/react-fontawesome` which enables the “Font Awesome” icons on React

dashboard

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,10 @@
     "lerna": "3.16.5"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^1.2.27",
+    "@fortawesome/free-solid-svg-icons": "^5.12.1",
+    "@fortawesome/react-fontawesome": "^0.1.8",
     "concurrently": "5.0.0",
-    "react-bootstrap": "^1.0.0-beta.16",
     "react-grid-system": "^6.1.0",
     "recharts": "^2.0.0-beta.1"
   }
--- a/package.json
+++ b/package.json
@@ -1,22 +1,36 @@
 {
-  "private": true,
+  "name": "siaviewnode-client",
+  "version": "0.0.1",
+  "main": "index.js",
+  "license": "MIT",
   "scripts": {
-    "client-dev": "yarn workspace siaviewnode-client dev -p 3000",
-    "server-dev": "yarn workspace siaviewnode-server dev",
-    "client-prod": "yarn workspace siaviewnode-client build && yarn workspace siaviewnode-client start -p 3000",
-    "server-prod": "yarn workspace siaviewnode-server build && yarn workspace siaviewnode-server start",
-    "dev": "concurrently \"yarn client-dev\" \"yarn server-dev\"",
-    "start": "NODE_ENV=production concurrently \"yarn client-prod\" \"yarn server-prod\""
-  },
-  "workspaces": {
-    "packages": [
-      "packages/*"
-    ]
-  },
-  "devDependencies": {
-    "lerna": "3.16.5"
+    "dev": "next",
+    "build": "next build",
+    "start": "next start"
   },
   "dependencies": {
-    "concurrently": "5.0.0"
+    "@emotion/core": "10.0.21",
+    "@fortawesome/fontawesome-svg-core": "^1.2.27",
+    "@fortawesome/free-solid-svg-icons": "^5.12.1",
+    "@fortawesome/react-fontawesome": "^0.1.8",
+    "@material-ui/core": "4.5.1",
+    "@mdx-js/react": "1.5.1",
+    "axios": "0.19.0",
+    "file-saver": "2.0.2",
+    "isomorphic-unfetch": "3.0.0",
+    "next": "9.1.1",
+    "ramda": "0.26.1",
+    "react": "16.10.2",
+    "react-dom": "16.10.2",
+    "react-dropzone": "10.2.1",
+    "react-grid-system": "^6.0.7",
+    "streamsaver": "2.0.3",
+    "theme-ui": "0.2.44",
+    "typescript": "3.6.4"
+  },
+  "devDependencies": {
+    "@types/node": "12.11.1",
+    "@types/ramda": "0.26.36",
+    "@types/theme-ui": "0.2.3"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,36 +1,25 @@
 {
-  "name": "siaviewnode-client",
-  "version": "0.0.1",
-  "main": "index.js",
-  "license": "MIT",
+  "private": true,
   "scripts": {
-    "dev": "next",
-    "build": "next build",
-    "start": "next start"
+    "client-dev": "yarn workspace siaviewnode-client dev -p 3000",
+    "server-dev": "yarn workspace siaviewnode-server dev",
+    "client-prod": "yarn workspace siaviewnode-client build && yarn workspace siaviewnode-client start -p 3000",
+    "server-prod": "yarn workspace siaviewnode-server build && yarn workspace siaviewnode-server start",
+    "dev": "concurrently \"yarn client-dev\" \"yarn server-dev\"",
+    "start": "NODE_ENV=production concurrently \"yarn client-prod\" \"yarn server-prod\""
   },
-  "dependencies": {
-    "@emotion/core": "10.0.21",
-    "@fortawesome/fontawesome-svg-core": "^1.2.27",
-    "@fortawesome/free-solid-svg-icons": "^5.12.1",
-    "@fortawesome/react-fontawesome": "^0.1.8",
-    "@material-ui/core": "4.5.1",
-    "@mdx-js/react": "1.5.1",
-    "axios": "0.19.0",
-    "file-saver": "2.0.2",
-    "isomorphic-unfetch": "3.0.0",
-    "next": "9.1.1",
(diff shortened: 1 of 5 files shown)


packages/siaviewnode-client/pages/stats.tsx
```diff
@@ -60,11 +61,7 @@ const Stats = () => {
                 <Container>
                     <Flex sx={{ alignItems: "center", height: 120 }}>
                         <Box>
-                            <Typography sx={{ fontWeight: 700 }}>
-                                <Link href="/">
-                                    Sia Skynet
-                                </Link>
-                            </Typography>
+                            <Typography sx={{ fontWeight: 700 }}>Sia Skynet</Typography>
```
peterjanbrone: any reason you threw out the link?

packages/siaviewnode-client/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,161 @@
...
+                </Container>
+            </Row>
+        </Container>
+    </div> )
+}
+
+
+// Custom component: StatBox
+function StatBox(props) {
+    return <Col md={5} lg={3} style={{border: '2px solid ' + theme.colors.statsElementsBorder, borderRadius: "15px", paddingTop: "20px", paddingBottom: "20px", justifyContent: 'center', textAlign: 'center', marginBottom: 50}}>
```
peterjanbrone: Quite a lot of inline styling going on here, perhaps it makes sense to add styled components (https://styled-components.com/docs/basics). Definitely not blocking this PR or anything like that though, just sharing in case you're not familiar with styled components

packages/siaviewnode-client/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,161 @@
...
+    } catch (e) {
+        return "#000000"
+    }
+}
+
+function componentToHex(c) {
+    // From an rgb component value to HEX (double digit)
+    var hex = c.toString(16);
+    return hex.length == 1 ? "0" + hex : hex;
+}
```
peterjanbrone: add newline

packages/siaviewnode-client/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,161 @@
...
+        var r = parseInt(rgb[1], 16) + rgbAddition
+        if (r > 255) {r = 255}
+        if (r < 0) {r = 0}
+        var g = parseInt(rgb[2], 16) + rgbAddition
+        if (g > 255) {g = 255}
+        if (g < 0) {g = 0}
+        var b = parseInt(rgb[3], 16) + rgbAddition
+        if (b > 255) {b = 255}
+        if (b < 0) {b = 0}
+        var newHex = "#" + componentToHex(r) + componentToHex(g) + componentToHex(b);
```
peterjanbrone: You can do string interpolation like this:
```
var newHex = `#{componentToHex(r)}{componentToHex(g)}{componentToHex(b)}`
```

packages/siaviewnode-client/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,161 @@
...
+    )
+}
+
+function colorChanger(hex, rgbAddition) {
+    // For the charts we create variations of the primary color
+    // This function transforms the HEX value into rgb, adds or substracts "rgbAddition"
+    // to eeach color component and transforms again into a new HEX value. By default, I 
+    // recommend a value of "rgbAddotion" = 50
+    try {
+        var rgb = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
```
peterjanbrone: ES6 introduced 2 new ways to create variables `const` and `let`, const being an immutable variable. Generally, it's best practice to use those.

packages/siaviewnode-client/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,161 @@
...
+        </span>
+    </Col>
+}
+
+// Custom component: StatBarChart
+function StatBarChart(props) {
+    // "series" is an object containing the name of the data serie and the color for the chart. It is
+    // mapped to a <Bar> component. Its format as an input prop is, as an example:
+    // {"p80": "#8884d8", "p95": "#82ca9d", "p99": "#666"}
+    const series = props.series
```
peterjanbrone: Just FYI you can destruct stuff like this:
```
const {series} = props
```

peterjanbrone:
This looks great :) left some general comments but most of it is just FYI. MR is definitely mergeable, I'll leave it open for a bit and give others a chance to review it.

package.json
```diff
@@ -17,6 +17,11 @@
     "lerna": "3.16.5"
   },
   "dependencies": {
-    "concurrently": "5.0.0"
+    "@fortawesome/fontawesome-svg-core": "^1.2.27",
+    "@fortawesome/free-solid-svg-icons": "^5.12.1",
+    "@fortawesome/react-fontawesome": "^0.1.8",
+    "concurrently": "5.0.0",
+    "react-grid-system": "^6.1.0",
+    "recharts": "^2.0.0-beta.1"
```
peterjanbrone: pretty heavy library dependency wise ^^ but it looks good! don't think we care too much about bundle size atm

packages/siaviewnode-client/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,161 @@
...
+export const Dashboard = (props) => {
+
+    const stats = props.api
+
+    return (<div style={{ width: "100%", paddingTop: 50}}>
+        <Container>
+            <Row justify="between">
+                <StatBox title="Pinned files" icon={faThumbtack} iconTransform="rotate-45" stat={stats.total_files_pinned} suffix=""/>
+                <StatBox title="Stored data" icon={faServer} iconTransform="" stat="0.145" suffix="TB"/>
+                <StatBox title="Files served this week" icon={faUpload} iconTransform="" stat={stats.total_files_served[168]} suffix=""/>
```
ro-tex: The 168 number is slightly magical. I get that it's hours and it's OK to use it like that but it might be a little easier on the reader if we moved all those (1, 24, 168, 720, 2160) to a set of constants with descriptive names. Just to remove a little cognitive load.

packages/siaviewnode-client/src/components/Dashboard.tsx
```diff
@@ -0,0 +1,161 @@
...
+        var rgb = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+        var r = parseInt(rgb[1], 16) + rgbAddition
+        if (r > 255) {r = 255}
+        if (r < 0) {r = 0}
+        var g = parseInt(rgb[2], 16) + rgbAddition
+        if (g > 255) {g = 255}
+        if (g < 0) {g = 0}
+        var b = parseInt(rgb[3], 16) + rgbAddition
+        if (b > 255) {b = 255}
+        if (b < 0) {b = 0}
```
ro-tex: This works well and can definitely stay as it is. If we want to DRY it up a little we can try something like this as well:
```
const normalise = (x) => {
  if (x < 0) { return 0 }
  if (x > 255) { return 255 }
  return x
}
let r = normalise(parseInt(rgb[1], 16) + rgbAddition)
...
```

ro-tex:
Looks good to me! 👍

kwypchlo:
@hakkane84 thank you for your contribution! During the last month some things changed in terms of the data sources we have at our disposal and also the webapp architecture. I will get back to you with a proper review this week.

hakkane84:
Thanks a lot. As the Portals have changed so deeply, I am not sure about how to integrate an stats panel in the current webportal, or even if it is worth at all.

Something very positive is that the stats API is finally exposed, so external website aggregators (like SiaStats) can now present stats per portal and aggregated too :-)

kwypchlo:
Sorry for taking so long to get back to this @hakkane84. Indeed we could use a stats page but the architecture of the portal implementation changed since the time this PR has been created (total ui redesign and then rewrite to Gatsby) so this page would also need to be rewritten. There are also exposed stats endpoints now so we could pick additional data from them. 
We are currently mainly focused on SiaStream but if you would be interested in contributing, please open a new issue and let's briefly discuss the data, designs and library choices :) Thanks!

PR #10, opened 2020-02-14
remove files that look like they were accidentally added

tbenz9:
Looks like someone added a bunch of files they probably didn't mean to.  Removing them.


kwypchlo:
Thank you for taking a look. Those were added initially as a part of a setup on our server. I will be working on a better way to manage that but for the time being we will hold on to those.

kwypchlo:
These files help out with initial portal setup and we're keeping them up to date as we develop new features.

PR #25, opened 2020-02-18
Fix a missing ; at end of line 11

rontus:
./setup.sh runs into a nginx error if that line does not end in ";"


kwypchlo:
Good catch, thank you!

PR #31, opened 2020-02-19
fixes overflowing of long unbreakable file names on mobile

kwypchlo:
### Before
Screen Shot 2020-02-19 at 17 22 26

### After
Screen Shot 2020-02-19 at 17 23 48


peterjanbrone:
Looks good, I'm wondering though if it would look better with a full-width button, did you try that out?

kwypchlo:
> Looks good, I'm wondering though if it would look better with a full-width button, did you try that out?

Yeah it looked a bit too heavy.

PR #56, opened 2020-03-03
Add health check script

marcinja:


setup-scripts/health-checker.py
```diff
@@ -0,0 +1,150 @@
...
+        opener = urllib.request.build_opener(handler)
+        opener.addheaders = [('User-agent', 'Sia-Agent')]
+
+        # Install the opener.
+        # Now all calls to urllib.request.urlopen use our opener.
+        urllib.request.install_opener(opener)
+
+    @staticmethod
+    def get_password():
+        password_file = open("/home/marcin/.sia/apipassword")
```
kwypchlo: that path should be either env variable or default to `~/.sia/apipassword`

PR #60, opened 2020-03-06
Directory Upload & Auto UUID

peterjanbrone:
This PR adds support for two new features:

**Directory Upload Support:**
In order to support directory upload, which is part of the Sia v1.4.4 release, there are some small changes required to the Nginx configuration. These changes include small updates to the regexes and properly passing the query string arguments on both download routes.

**Auto UUID support**
This small feature enables users to upload a Skyfile to Skynet without having to specify a UUID themselves, but rather have Nginx generate one. This is achieved by using the `$request_id` Nginx variable. It is commented out by default.

### DIFF
--- a/setup-scripts/skynet-nginx.conf
+++ b/setup-scripts/skynet-nginx.conf
@@ -13,7 +13,7 @@ server {
 	# Enable the following line if you want to have auto uuid support. This
 	# means users are able to upload Skyfiles without having to provide a uuid
 	# themselves.
-	# rewrite ^/skynet/skyfile(\?.*)?$ /skynet/skyfile/$request_time$1;
+	# rewrite ^/skynet/skyfile/?$ /skynet/skyfile/$request_id$1;
 
 	# NOTE: make sure to enable any additional configuration you might need like gzip
 
--- a/setup-scripts/skynet-nginx.conf
+++ b/setup-scripts/skynet-nginx.conf
@@ -10,6 +10,11 @@ server {
 	listen [::]:443 ssl http2;
 	server_name siasky.net www.siasky.net; # replace with actual server names
 
+	# Enable the following line if you want to have auto uuid support. This
+	# means users are able to upload Skyfiles without having to provide a uuid
+	# themselves.
+	# rewrite ^/skynet/skyfile(\?.*)?$ /skynet/skyfile/$request_time$1;
+
 	# NOTE: make sure to enable any additional configuration you might need like gzip
 
 	location / {
@@ -32,10 +37,11 @@ server {
 		proxy_set_header Authorization "Basic BASE64_AUTHENTICATION";
 	}
 
-	location ~ "^/([a-zA-Z0-9-_]{46})$" {
+	location ~ "^/([a-zA-Z0-9-_]{46}(/.*)?)$" {
 		proxy_read_timeout 600;
-		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is correct)
-		proxy_pass http://127.0.0.1:9980/skynet/skylink/$1;
+		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is
+		# correct)
+		proxy_pass http://127.0.0.1:9980/skynet/skylink/$1$is_args$args;
 		proxy_set_header Access-Control-Allow-Origin: *;
 		# make sure to override user agent header - siad requirement
 		proxy_set_header User-Agent: Sia-Agent;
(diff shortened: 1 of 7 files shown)


setup-scripts/skynet-nginx.conf
```diff
@@ -10,6 +10,11 @@ server {
 	listen [::]:443 ssl http2;
 	server_name siasky.net www.siasky.net; # replace with actual server names
 
+	# Enable the following line if you want to have auto uuid support. This
+	# means users are able to upload Skyfiles without having to provide a uuid
+	# themselves.
+	# rewrite ^/skynet/skyfile(\?.*)?$ /skynet/skyfile/$request_time$1;
```
kwypchlo: This would match both `/skynet/skyfile` and `/skynet/skyfile/nsai90a0s2` - I think it would be safe to enable it by default if user posts to `/skynet/skyfile` directly but leave an option to manually set it if clients wants to.

Could you test this configuration? (the trailing slash is optional)
```suggestion
	# rewrite ^/skynet/skyfile/?$ /skynet/skyfile/$request_time$1;
```

peterjanbrone: Oof thanks for catching it. It was actually just a copy paste from old config.
It should  be `rewrite ^/skynet/skyfile/?$ /skynet/skyfile/$request_id$1;`

src/components/HomeUpload/HomeUpload.js
```diff
@@ -31,22 +56,49 @@ export default function HomeUpload() {
       });
     };
 
+    const upload = async (formData, directory) => {
+      const uuid = shortid.generate();
+      const uploadUrl = `${apiUrl}/skynet/skyfile/${uuid}${
```
peterjanbrone: we should drop uuid here

src/components/HomeUpload/HomeUpload.js
```diff
@@ -11,8 +11,33 @@ import AppContext from "../../AppContext";
 export default function HomeUpload() {
   const [files, setFiles] = useState([]);
   const { apiUrl } = useContext(AppContext);
+  const [directoryMode, setDirectoryMode] = useState(false);
+
+  useEffect(() => {
+    if (directoryMode) {
+      inputRef.current.setAttribute("webkitdirectory", "true");
```
peterjanbrone: Don't we also need `multiple` and ``mozdirectory`?

peterjanbrone: Is there a race condition where `inputRef.current` is undefined?

src/components/HomeUpload/HomeUpload.js
```diff
@@ -11,8 +11,33 @@ import AppContext from "../../AppContext";
...
+    } else {
+      inputRef.current.removeAttribute("webkitdirectory");
+    }
+  }, [directoryMode]);
+
+  const getFilePath = (file) => {
+    const filePath = file.path || file.webkitRelativePath || file.name;
```
peterjanbrone: Is it safer to do `file.webkitRelativePath || file.path || file.name;`?

src/components/HomeUpload/HomeUpload.js
```diff
@@ -11,8 +11,33 @@ import AppContext from "../../AppContext";
...
+
+  useEffect(() => {
+    if (directoryMode) {
+      inputRef.current.setAttribute("webkitdirectory", "true");
+    } else {
+      inputRef.current.removeAttribute("webkitdirectory");
+    }
+  }, [directoryMode]);
+
+  const getFilePath = (file) => {
```
peterjanbrone: If this function gets called on a file that's not uploaded through a dir upload, I believe it fails.
Could we add sanity checks here or make that apparent either through a comment or variable name?

src/components/HomeUpload/HomeUpload.js
peterjanbrone: don't forget to drop `shortid` dependency while you're at it

src/components/HomeUpload/HomeUpload.js
```diff
@@ -31,22 +56,49 @@ export default function HomeUpload() {
...
+          });
+        } else {
+          formData.append("file", file);
+        }
+
+        const skylink = await upload(formData, directoryMode && file.name);
 
         onComplete(file, "complete", skylink);
       } catch (error) {
+        console.error(error);
```
peterjanbrone: debug logging? do we want this in?

src/components/HomeUpload/HomeUpload.js
kwypchlo: no, the ref is assigned before the useEffect is ran

src/components/HomeUpload/HomeUpload.js
kwypchlo: good good

src/components/HomeUpload/HomeUpload.js
kwypchlo: We don't need `multiple` because it's already set. 
We don't need `mozdirectory` because mozilla supports `webkitdirectory` https://caniuse.com/#feat=mdn-api_htmlinputelement_webkitdirectory

src/components/HomeUpload/HomeUpload.js
kwypchlo:
```suggestion
    const filePath = file.webkitRelativePath || file.path || file.name;
```

src/components/HomeUpload/HomeUpload.js
kwypchlo:
```suggestion
```

src/components/HomeUpload/HomeUpload.js
kwypchlo: It's not going to fail, we are actually calling it for every file, regardless of whether it was directory or files upload - https://github.com/NebulousLabs/skynet-webportal/pull/60/files#diff-4f89bf10d6d343f05c6bdb8c1edb1135R72

peterjanbrone: hm I don't understand how that works because of the `slice(1)` in case of plain files...

peterjanbrone: nevermind! :)

PR #73, opened 2020-03-09
upload progress status

kwypchlo:
- added uploading percentage indicator
- added support for "processing" state between last uploaded byte and api response
- resolved (at least temporarily) cors issue with axios

case study:

case 1:
- no `add_header Access-Control-Allow-Origin: *;`
- no `proxy_hide_header Access-Control-Allow-Origin: *;`
- no `proxy_set_header Access-Control-Allow-Origin: *;`

`No 'Access-Control-Allow-Origin' header is present on the requested resource.`

case 2: 
- added `add_header Access-Control-Allow-Origin: *;`
- no `proxy_hide_header Access-Control-Allow-Origin: *;`
- no `proxy_set_header Access-Control-Allow-Origin: *;`

`The 'Access-Control-Allow-Origin' header contains multiple values '*, *', but only one is allowed.`

case 3:
- no `add_header Access-Control-Allow-Origin: *;`
- no `proxy_hide_header Access-Control-Allow-Origin: *;`
- added `proxy_set_header Access-Control-Allow-Origin: *;`

`No 'Access-Control-Allow-Origin' header is present on the requested resource.`

case 4: 
- added `add_header Access-Control-Allow-Origin: *;`
- no `proxy_hide_header Access-Control-Allow-Origin: *;`
- added `proxy_set_header Access-Control-Allow-Origin: *;`

`The 'Access-Control-Allow-Origin' header contains multiple values '*, *', but only one is allowed.`

case 5: 
- added `add_header Access-Control-Allow-Origin: *;`
- added `proxy_hide_header Access-Control-Allow-Origin: *;`
- no `proxy_set_header Access-Control-Allow-Origin: *;`

works 🎉

### DIFF
--- a/setup-scripts/skynet-nginx.conf
+++ b/setup-scripts/skynet-nginx.conf
@@ -22,7 +22,8 @@ server {
 		# proxy this call to siad endpoint (make sure the ip is correct)
 		proxy_pass http://127.0.0.1:9980;
 		proxy_set_header Expect $http_expect;
-		proxy_set_header Access-Control-Allow-Origin: *;
+		add_header Access-Control-Allow-Origin *;
+		proxy_hide_header Access-Control-Allow-Origin;
 		# make sure to override user agent header - siad requirement
 		proxy_set_header User-Agent: Sia-Agent; 
 		# replace BASE64_AUTHENTICATION with base64 encoded <user>:<password>
--- a/src/components/HomeUpload/HomeUpload.js
+++ b/src/components/HomeUpload/HomeUpload.js
@@ -7,6 +7,7 @@ import { Button, UploadFile } from "../";
 import { Deco3, Deco4, Deco5, Folder, DownArrow } from "../../svg";
 import "./HomeUpload.scss";
 import AppContext from "../../AppContext";
+import axios from "axios";
 
 export default function HomeUpload() {
   const [files, setFiles] = useState([]);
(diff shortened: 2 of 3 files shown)


setup-scripts/skynet-nginx.conf
```diff
@@ -17,12 +17,13 @@ server {
...
 
 	location /skynet/skyfile/ {
+		add_header Access-Control-Allow-Origin: *;
 		client_max_body_size 1000M; # make sure to limit the size of upload to a sane value
 		proxy_read_timeout 600;
 		# proxy this call to siad endpoint (make sure the ip is correct)
 		proxy_pass http://127.0.0.1:9980;
 		proxy_set_header Expect $http_expect;
-		proxy_set_header Access-Control-Allow-Origin: *;
+		proxy_hide_header Access-Control-Allow-Origin: *;
```
peterjanbrone: I'm confused with this syntax. Why would `hide_header` take a value, I'd expect it'd take a key.
Also what's the reasoning, to me `add_header` in combination with `proxy_hide_header` is not really redundant, but in this case it is because the proxy header is set to the wildcard as well..

peterjanbrone: What about doing `proxy_pass_header Access-Control-Allow-Origin`

kwypchlo: > What about doing `proxy_pass_header Access-Control-Allow-Origin`

`No 'Access-Control-Allow-Origin' header is present on the requested resource.`

kwypchlo:
```suggestion
		proxy_hide_header Access-Control-Allow-Origin;
```

kwypchlo: > I'm confused with this syntax. Why would `hide_header` take a value, I'd expect it'd take a key.

my mistake, fixed

setup-scripts/skynet-nginx.conf
```diff
@@ -17,12 +17,13 @@ server {
 	}
 
 	location /skynet/skyfile/ {
+		add_header Access-Control-Allow-Origin: *;
```
peterjanbrone: maybe group both CORS changes and document it a little, no need to go into it at length though I think

PR #80, opened 2020-03-10
[WIP] Add /version endpoint

peterjanbrone:
This MR adds a `/version` route which was suggested by someone in the community.
Please note this returns the version of the "download-siad".
It's live on https://siasky.dev/version, you can see the output there.


peterjanbrone:
On hold until confirmed we want to go ahead with this. With regards to CORS. We should fix that once and for all. Could we do something like this: https://enable-cors.org/server_nginx.html @kwypchlo ?

peterjanbrone:
We will expose this information through `/stats`

PR #97, opened 2020-03-19
Update Nginx config to use double siad setup

peterjanbrone:
Point upload route to port '9970' and add documentation.

### DIFF
--- a/setup-scripts/skynet-nginx.conf
+++ b/setup-scripts/skynet-nginx.conf
@@ -46,8 +46,18 @@ server {
 	location /skynet/skyfile/ {
 		client_max_body_size 1000M; # make sure to limit the size of upload to a sane value
 		proxy_read_timeout 600;
+
 		# proxy this call to siad endpoint (make sure the ip is correct)
-		proxy_pass http://127.0.0.1:9980;
+		#
+		# note that we point uploads to port '9970', do this when you want to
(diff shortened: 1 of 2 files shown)


setup-scripts/skynet-nginx.conf
```diff
@@ -46,8 +46,18 @@ server {
...
-		proxy_pass http://127.0.0.1:9980;
+		#
+		# note that we point uploads to port '9970', do this when you want to
+		# run in a configuration where you have two siad instances, one for
+		# downloads and one for uploads. This drastically improves the up - and
+		# download speed of your portal. When running your portal in this double
+		# siad setup, make sure only the download portal runs in 'portal mode'.
+		# The upload siad can be run in normal mode. Set the port to '9980' if
+		# you do not want to run your portal in the double siad setup.
+		proxy_pass http://127.0.0.1:9970;
```
kwypchlo: should this be a default though?

kwypchlo:
@peterjanbrone we should also update https://github.com/NebulousLabs/skynet-webportal/blob/master/setup-scripts/README.md to point that 2 siads is the actual recommended method instead of "in some cases". Either you could do that or I can later today.

peterjanbrone:
oh yeah that makes sense! I'll add that

PR #148, opened 2020-04-06
include basic ddos protection

kwypchlo:
- limit /stats and /statsdown requests to 10 per minute
- close slow connections
- limit 10 download connections from one ip at a time
- limit 10 uploads from one ip at a time

### DIFF
--- a/setup-scripts/skynet-nginx.conf
+++ b/setup-scripts/skynet-nginx.conf
@@ -1,3 +1,7 @@
+limit_req_zone $binary_remote_addr zone=stats:10m rate=10r/m;
+limit_conn_zone $binary_remote_addr zone=uploads:10m;
+limit_conn_zone $binary_remote_addr zone=downloads:10m;
+
 server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
@@ -9,6 +13,10 @@ server {
    	listen 443 ssl http2;
 	listen [::]:443 ssl http2;
 	server_name siasky.net www.siasky.net; # replace with actual server names
+
+	# ddos protection: closing slow connections
+	client_body_timeout 5s;
+	client_header_timeout 5s;
 
 	# Enable the following line if you want to have auto uuid support. This
 	# means users are able to upload Skyfiles without having to provide a uuid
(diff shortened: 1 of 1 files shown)


setup-scripts/skynet-nginx.conf
```diff
@@ -10,6 +14,10 @@ server {
 	listen [::]:443 ssl http2;
 	server_name siasky.net www.siasky.net; # replace with actual server names
 
+	# ddos protection: closing slow connections
```
peterjanbrone, line 17: These are good additions. We should keep these.

setup-scripts/skynet-nginx.conf
```diff
@@ -1,3 +1,7 @@
+limit_req_zone $binary_remote_addr zone=stats:10m rate=10r/m;
```
peterjanbrone, line 1: 10r/m  feels like a very constricting rate, is there any reason we're being so conservative? Could we bump this up to 1r/s?

setup-scripts/skynet-nginx.conf
```diff
@@ -22,28 +30,33 @@ server {
...
+		# replace BASE64_AUTHENTICATION with base64 encoded <user>:<password>
 		# for sia user is empty so it's just :<password>
 		# to generate the passcode use https://www.base64encode.org or any other base64 encoder
 		proxy_set_header Authorization "Basic BASE64_AUTHENTICATION";
 		proxy_pass http://127.0.0.1:9980/skynet/stats;
-  }
+	}
 
 	location /skynet/skyfile/ {
+		limit_conn uploads 10; # ddos protection: max 10 uploads at a time
```
peterjanbrone, line 59: Is this defined per zone? Which I believe in our case is per unique IP? Perhaps a good idea to include it in the comment. If it's per IP I feel these are extremely limiting. Also by default it returns an error most probably, but which one? I think I would probably expect 429 but I'm unsure if that is the default.

setup-scripts/skynet-nginx.conf
```diff
@@ -70,6 +83,8 @@ server {
 	}
 
 	location ~ "^/([a-zA-Z0-9-_]{46}(/.*)?)$" {
+		limit_conn downloads 10; # ddos protection: max 10 downloads at a time
```
peterjanbrone, line 86: Same remark

setup-scripts/skynet-nginx.conf
peterjanbrone: oh I'm sorry I missed the zone, I think I would drop this down to 1r/m in that case :D

setup-scripts/skynet-nginx.conf
```diff
@@ -1,3 +1,7 @@
+limit_req_zone $binary_remote_addr zone=stats:10m rate=10r/m;
+limit_conn_zone $binary_remote_addr zone=uploads:10m;
+limit_conn_zone $binary_remote_addr zone=downloads:10m;
```
peterjanbrone, line 3: PP: I think it makes more sense to name the zones by the aggregator. 
So I would've probably defined a zone `byIPAddress`
And then in every location define the rate through the `limit_conn` directive.

setup-scripts/skynet-nginx.conf
peterjanbrone: I think 10 might be a little low. I think Ivo implemented parallel uploads in siac which defaulted to 8, unsure at what he capped that but we should probably make it so the max value there is the default value here? In any case 10 feels a bit restrictive

PR #149, opened 2020-04-12
Changed nginx config to generate random file uuid by default

Delivator:


kwypchlo:
Thank you, I've been meaning to update that myself. Can you also add a note that it requires at least nginx 1.11.0 (probably worth mentioning in https://github.com/NebulousLabs/skynet-webportal/blob/master/setup-scripts/README.md too).

kwypchlo:
Perfect, thank you :)

PR #192, opened 2020-04-28
fix whitespace encoding for directories

### LINKED ISSUE #190: Directory Upload will not generate accessible links for filenames containing spaces
When attempting to upload a folder via the webportal, the upload will complete successfully but files that would ordinarily require percent-encoding do not seem to be accessible even when attempting to access the proper percent-encoded paths.

Example:

Given a folder containing /folder/file.txt and /folder/my file.txt, the former is accessible at the URL http:// skynet-webportal/skylink/file.txt but the latter is not accessible at the URL http:// skynet-webportal/skylink/my%20file.txt

Please fix the upload of files inside directories that would require percent-encoded paths for retrieval when using the proper percent-encoded HTTP path.

A list of encoded characters and relevant RFCs can be found here: https://en.wikipedia.org/wiki/Percent-encoding

kwypchlo:
https://stackoverflow.com/questions/8838905/how-to-avoid-nginx-to-replace-20-by-whitespace-when-using-as-a-proxy-proxy-pas

closes #190

### DIFF
--- a/setup-scripts/skynet-nginx.conf
+++ b/setup-scripts/skynet-nginx.conf
@@ -86,10 +86,13 @@ server {
 	location ~ "^/([a-zA-Z0-9-_]{46}(/.*)?)$" {
 		limit_conn downloads_by_ip 10; # ddos protection: max 10 downloads at a time
 
+		# we need to explicitly use set directive here because $1 will contain the skylink with
+		# decoded whitespaces and set will re-encode it for us before passing it to proxy_pass
+		set $skylink $1;
+
 		proxy_read_timeout 600;
-		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is
-		# correct)
-		proxy_pass http://127.0.0.1:9980/skynet/skylink/$1$is_args$args;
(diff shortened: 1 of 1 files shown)


setup-scripts/skynet-nginx.conf
```diff
@@ -111,11 +114,14 @@ server {
...
+		# we need to explicitly use set directive here because $1 will contain the skylink with
+		# decoded whitespaces and set will re-encode it for us before passing it to proxy_pass
+		set $skylink $1;
+
 		proxy_read_timeout 600;
-		# proxy this call to siad /skunet/skylink/ endpoint (make sure the ip is
-		# correct) this alias also adds attachment=true url param to force
-		# download the file
-		proxy_pass http://127.0.0.1:9980/skynet/skylink/$1?attachment=true&$args;
+		# proxy this call to siad /skunet/skylink/ endpoint (make sure the ip is correct)
```
peterjanbrone, line 122: typo: `proxy this call to siad /skynet/skylink/`

kwypchlo, line 122:
```suggestion
		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is correct)
```

PR #213, opened 2020-05-21
Remove guild role requirement from bot_utils.

themailman05:
I was getting a Nonetype error when force_notify was being called, because "role.mention" was None.

I'm a Discord n00b, so forgive me and ignore if there is anything grievously wrong I am missing here.

### DIFF
--- a/setup-scripts/bot_utils.py
+++ b/setup-scripts/bot_utils.py
@@ -10,7 +10,7 @@ import urllib, json, os, traceback, discord, sys
 sc_precision = 10 ** 24
 
 CHANNEL_NAME = "skynet-portal-health-check"
-ROLE_NAME = "skynet-prod"
+ROLE_NAME = "everyone"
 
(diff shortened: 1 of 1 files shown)


setup-scripts/bot_utils.py
```diff
@@ -10,7 +10,7 @@
 sc_precision = 10 ** 24
 
 CHANNEL_NAME = "skynet-portal-health-check"
-ROLE_NAME = "skynet-prod"
+ROLE_NAME = "everyone"
```
marcinja, line 13: `ROLE_NAME` should be another environment variable so it can be set however you'd like. We use `skynet-prod` so only a subset of people on our discord get notified instead of everyone

kwypchlo:
made role optional in https://github.com/SkynetLabs/skynet-webportal/commit/fde760ec860138d0b01199a8a4878af55e4a866e

PR #215, opened 2020-05-21
Added skynet cache to nginx config, #214 fix

### LINKED ISSUE #214: Nginx proxy_cache zone "skynet" is unknown
The nginx config features a cache that has not been configured, thus nginx throws "unknown zone"
https://github.com/NebulousLabs/skynet-webportal/blob/468d2b036e768f7f64cc7e288dedc6062ec6eca9/setup-scripts/skynet-nginx.conf#L117

Delivator:
Should fix #214

### DIFF
--- a/setup-scripts/skynet-nginx.conf
+++ b/setup-scripts/skynet-nginx.conf
@@ -4,7 +4,7 @@ limit_conn_zone $binary_remote_addr zone=downloads_by_ip:10m;
 limit_req_status 429;
 limit_conn_status 429;
 
-proxy_cache_path /temp/nginx keys_zone=skynet:10m inactive=1d;
+proxy_cache_path /tmp/nginx keys_zone=skynet:10m inactive=1d;
 
 server {
 	listen 80 default_server;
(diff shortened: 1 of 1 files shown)


DaWe35:
I think we also need to add
```
sudo mkdir /data
sudo mkdir /data/nginx
sudo mkdir /data/nginx/cache
```
to setup.sh

Delivator:
Woops got an typto, /tmp/ instead of /temp/

marcinja:
Thanks for catching this.

On `siasky.net` we actually are using:
```
proxy_cache_path /tmp/nginx levels=1:2 keys_zone=skynet:10m max_size=10g inactive=24h use_temp_path=off;
```

I think we grabbed that from https://www.nginx.com/blog/nginx-caching-guide/ which has some justification.

Delivator:
I mean max_size is up to the user, isn't it?
levels=1:2 and use_temp_path=off are probably a good idea, ill add these.

Delivator:
Should be ready to be merged

kwypchlo:
Thanks for the contribution @Delivator !

PR #296, opened 2020-07-17
Containerised sia setup

kwypchlo:
_work in progress_


docker-compose.yml
```diff
@@ -15,6 +15,23 @@ services:
     networks:
       - shared
 
+  sia:
+    image: nebulouslabs/sia:dev
```
ro-tex, line 19: This needs to be the stable image. I guess we're gonna switch to a tag as soon as we have one.

docker/sia/Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM golang AS builder
+LABEL maintainer="NebulousLabs <devs@nebulous.tech>"
+
+ENV GOOS linux
+ENV GOARCH amd64
+
+RUN git clone https://gitlab.com/NebulousLabs/Sia.git && \
+    cd Sia && \
+    git checkout siad-v150-pre-rc && \
+    make release
```
ro-tex, line 10: This whole file should be gone - we should tag the RC, publish an image and use that one.

m-cat:
Reminder to close https://gitlab.com/NebulousLabs/Sia/-/issues/3768 once this is merged.

docker-compose.yml
kwypchlo: using a builder here with v1.5.0 branch

PR #302, opened 2020-07-23
Handshake integration

kwypchlo:
- added handshake and handshake-api services to docker-compose
- exposed `/hsn/<domain>` endpoint to resolve and redirect to a domain that contains skylink
- expose `/hsnres/<domain>` endpoint that returns the resolved skylink from handshake

### DIFF
--- a/docker/handshake-api/Dockerfile
+++ b/docker/handshake-api/Dockerfile
@@ -2,15 +2,13 @@ FROM node:14.5
 
 WORKDIR /usr/app
 
-RUN yarn init -y &&
+RUN yarn init -y && \
     yarn add express express-http-proxy hs-client
 
 COPY handshake-api/index.js ./
 
 ENV HOST="localhost"
 ENV PORT=3100
-
-ENV PORTAL="https://siasky.net"
 
 ENV HSD_NETWORK="main"
 ENV HSD_HOST="0.0.0.0"
--- a/docker/handshake-api/Dockerfile
+++ b/docker/handshake-api/Dockerfile
@@ -3,7 +3,7 @@ FROM node:14.5
 WORKDIR /usr/app
 
 RUN yarn init -y
-RUN yarn add body-parser express hs-client
+RUN yarn add express express-http-proxy hs-client
 
 COPY handshake-api/index.js ./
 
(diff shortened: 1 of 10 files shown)


docker/handshake-api/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM node:14.5
+
+WORKDIR /usr/app
+
+RUN yarn init -y
+RUN yarn add body-parser express hs-client
```
ro-tex, line 6: These two should be run together, there is no need to create two layers.
`yarn init -y && yarn add body-parser express hs-client`

ro-tex: Also... are you sure we should skip `package.json`? We can but we won't have the peace of mind that we're running the same code each time - if a new major version of one of the libraries comes out we'll automatically get it and it might not work with the rest and all that stuff.

It would e safer to have `package.json` and `package-lock.json` and use `yarn install --frozen-lockfile`, so we really know what we're running.

docker/handshake/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:14.5
+
+WORKDIR /opt/hsd
+
+RUN git clone https://github.com/handshake-org/hsd.git /opt/hsd
+RUN npm install --production
```
ro-tex, line 6: Same here - should be run as a single command.

handshake-api/index.js
```diff
@@ -0,0 +1,75 @@
...
+  apiKey: hsdApiKey,
+};
+const client = new NodeClient(clientOptions);
+
+const resolveDomain = async (name) => {
+  const response = await client.execute("getnameresource", [name]);
+
+  if (!response) throw new Error("API not responding");
+
+  console.log(`${name} => ${JSON.stringify(response.records)}`);
```
ro-tex, line 26: Potential nullpointer on empty response. We are accessing them directly without checking their existence on line 34 again, so we might just as well verify them here and throw a "no records found" exception.

docker/handshake-api/Dockerfile
kwypchlo: I will fill that gap when I migrate this to monorepo, it's temporary for now.

ro-tex: OK, we'll leave it as a F/U then.

handshake-api/index.js
ro-tex: It's not a problem here because we check `response` on line 24.

PR #308, opened 2020-07-27
WIP: drop caddy, step one

kwypchlo:
- drop caddy redirects, only leave caddy for ssl 
- fix cors from nginx
- add route53 support


docker-compose.yml
```diff
@@ -57,6 +61,7 @@ services:
       dockerfile: Dockerfile
     container_name: handshake
     restart: on-failure
+    command: hnd --log-console=false
```
ro-tex, line 64: Shouldn't this be `command: hsd --log-console=false`

docker/caddy/Dockerfile
```diff
@@ -1,20 +1,7 @@
...
-COPY yarn.lock .
-
-ENV CYPRESS_INSTALL_BINARY 0
-RUN yarn --frozen-lockfile
-RUN yarn build
-
 FROM caddy:2.1.1-builder AS caddy-builder
 
-RUN caddy-builder github.com/caddy-dns/cloudflare
+RUN caddy-builder github.com/caddy-dns/cloudflare github.com/caddy-dns/route53
```
ro-tex, line 3: Do we still need `github.com/caddy-dns/cloudflare`?

docker/nginx/conf.d/client.conf
```diff
@@ -41,39 +41,102 @@ server {
...
 		proxy_cache skynet;
 		proxy_cache_valid any 1m; # cache stats for 1 minute
+		proxy_set_header User-Agent: Sia-Agent;
 		proxy_pass http://siad-download/skynet/stats;
 	}
 
+	location /health-check {
+		include /etc/nginx/conf.d/include/cors;
+
+		proxy_pass http://health-check:3100;
```
ro-tex, line 116: Heathcheck mapped on the same port as Handshake?

kwypchlo: yeah, it's different ip so it's ok

docker/caddy/Dockerfile
kwypchlo: not really, leaving it there for anyone else to use for now

docker-compose.yml
kwypchlo: nope, handshake docker container uses `hsd` (handshake daemon) `ENTRYPOINT ["hsd"]`

PR #309, opened 2020-07-28
clean up double siad setup

kwypchlo:
with 1.5.0 coming up, we don't need to preserver double-siad setup any more


docker/nginx/conf.d/client.conf
```diff
@@ -107,7 +100,7 @@ server {
 		proxy_cache skynet;
 		proxy_cache_valid any 1m; # cache stats for 1 minute
 		proxy_set_header User-Agent: Sia-Agent;
-		proxy_pass http://siad-download/skynet/stats;
+		proxy_pass http://siad/skynet/stats; # serve upload node stats temporarily
```
ro-tex, line 103: I'm not sure I follow how this points to the upload instance.

setup-scripts/blacklist-skylink.sh
```diff
@@ -11,7 +11,6 @@ do
     echo "⌁ Blacklisting on ${server}"
     ssh -q -t user@${server} 'curl -A Sia-Agent --user "":$(cat /home/user/.sia/apipassword) --data '"'"'{"add":["'$1'"]}'"'"' "localhost:9980/skynet/blacklist"'
     ssh -q -t user@${server} 'rm -rf /home/user/skynet_webportal/docker/data/nginx/cache' # remove cache from docker-managed portals
-    ssh -q -t user@${server} 'sudo rm -rf /tmp/nginx' # remove cache from legacy non-docker portals
```
ro-tex, line 14: Ha ha, bye-bye non-docker portals! :D

setup-scripts/setup-health-check-scripts.sh
```diff
@@ -9,13 +9,7 @@ pip3 install discord.py
 pip3 install python-dotenv
 
 downloadCheck="0 0,8,16 * * * /home/user/skynet-webportal/setup-scripts/funds-checker.py /home/user/.sia/sia.env"
```
ro-tex, line 11: We should probably rename this to `siadCheck` or something similar.

PR #310, opened 2020-07-31
fix hns header overflow and redirects

kwypchlo:


kwypchlo:
@ro-tex please re-check my latest commits, merged in #322 and made sure the redirect works with `sia://` prefix

PR #322, opened 2020-08-05
HNS sia prefix

ro-tex:
Support only full skylinks, i.e. prefixed with `sia://`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
       - name: Use Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: 12.x
+          node-version: 14.6
 
       - name: Install dependencies
         run: yarn
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,4 @@ docker/data
 # Cache files
 __pycache__
 /.idea/
+/venv/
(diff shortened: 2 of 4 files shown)


m-cat:
Also, in `skynet-js` we made the double slash option so that `sia:<skylink>` works as well

ro-tex:
> Also, in `skynet-js` we made the double slash optional so that `sia:<skylink>` works as well

I think it would be better to be a little more restrictive here. Ideally, we also wouldn't allow bare skylinks but it would be nice to support them, at least temporarily, because of stuff people might have built since the Medium article.

Hmm.. thinking about this now, people wouldn't have had a chance to build much and if we keep supporting it now we might be forced to support it forever. I'll revert that and make the `sia://` prefix obligatory.

ro-tex:
I don't want to merge this until @kwypchlo has had a chance to take a look.

Also, there is still the question of the trailing slash redirect - we need it for the `hns` endpoint as well because skapps will be executed right under it and the same problem with relative links will haunt us. You can see in the history of commits my naiive attempt to add it, which unfortunately didn't work - I got redirected to the root page (e.g. `https://siasky.net`) instead. I'm sure @kwypchlo will know why that's happening.

ro-tex:
> Does the standard `?` work here for "0 or 1"? Otherwise looks good

Yes, you are right, @m-cat. Updated.

PR #328, opened 2020-08-12
Replace broken SDK link with link to SDK docs

m-cat:


src/components/HomeSamples/HomeSamples.js
```diff
@@ -23,7 +23,7 @@ export default function HomeSamples() {
         <p>
           Skynet includes{" "}
           <a
-            href="https://support.siasky.net/article/hrshqsn9wz-integrating-skynet"
+            href="https://nebulouslabs.github.io/skynet-docs/?shell--curl#introduction"
```
MSevey, line 26: why does this include `?shell--curl#introduction`?

PR #329, opened 2020-08-13
Wildcard Subdomains

peterjanbrone:
This PR adds a `server` block to the nginx configuration that handles wildcard subdomains. Those wildcards are base32 encoded skylinks and should be redirected. This is part of the changes required to have Skapps use cookies, since those depend on the domain that, prior to this change, were shared across all skapps.


docker/nginx/conf.d/client.conf
```diff
@@ -35,6 +39,17 @@ server {
...
+		# or rewrite, since we need to proxy_pass we have to work our way around
+		# using a custom error code.
+		#
+		# See https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/
+		error_page 418 = @subdomain;
+		recursive_error_pages on;
+		if ($subdomain  != "") {
+			return 418;
+		}
+
```
kwypchlo, line 52: why do we need that, what is this for?

docker/nginx/conf.d/client.conf
```diff
@@ -25,6 +25,10 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
+	# parse subdomain into custom variable
+	server_name ~^(.*)?\.siasky\.dev|siasky\.dev$;
+	set $subdomain $1;
```
kwypchlo, line 30: What happens if I'll try to access [https://germany.siasky.net](https://germany.siasky.net) ? Will it assume `germany` is the skylink?

ro-tex: We can fix that by making the regex match exactly N characters.

docker/nginx/conf.d/client.conf
peterjanbrone: I've linked the explanation in the comment. This is the suggested implementation if you want to use the 'if' directive in a `location` block. I think it's a pretty neat way to handle the proxy actually. The fact that it's the proposed solution in the nginx docs makes it even more ok to be using imo.

kwypchlo: what I mean is why do we need that whole if at all ? Am I not getting the logic here? the `location /` block is for the main webpage and it has to be accessible without the subdomain and this `if` block basically will return `418` in that case?

docker/nginx/conf.d/client.conf
kwypchlo: Is it possible that the `server_name` wouldn't contain the `siasky` domain reference? Maybe some wildcard as a domain name would do the trick? I did my best not to hardcode our domain name (unless it was completely necessary) and instead used environment variables.

peterjanbrone: https://regex101.com/r/If5280/1

docker/nginx/conf.d/client.conf
peterjanbrone: I think we have to add it to this `location` block because the urls won't contain a skylink no more, but rather a base32 encoded skylink as subdomain. Could be that I am mistaken but I tested this config already and it worked perfectly.

To clarify:
this `https://200dd950uidil1uqqpdem6qaf6s3l46pu1h74ff1v5s4fh8ag3dece0.siasky.dev/` needs to hit the `if`
this `https://siasky.dev/` does not,
and I don't think it will.

docker/nginx/conf.d/client.conf
peterjanbrone: @kwypchlo we could also make it `^([a-z0-9]{55}).*?$` wdyt? would that work for you?

docker/nginx/conf.d/client.conf
peterjanbrone: To clarify further, defining `error_page` causes an **internal** redirect to happen. So it's not like the user is seeing that error code, it'll just be redirected internally to a named location. http://nginx.org/en/docs/http/ngx_http_core_module.html#error_page

docker/nginx/conf.d/client.conf
```diff
@@ -25,6 +25,10 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
+	# parse subdomain into custom variable
+	server_name ~^([a-z0-9]{55}).*?\.siasky\.(dev|net)$;
```
kwypchlo, line 29: Would this work ? A little bit more generic.

```suggestion
	server_name ~^([a-z0-9]{55}).*?;
```

kwypchlo (changes requested):
Can you merge master to this PR? The download block code that has been copied has been changed a little bit - it would actually make sense to extract it instead of copy-pasting it for the third time.

docker/nginx/conf.d/client.conf
```diff
@@ -197,6 +212,31 @@ server {
 		proxy_cache_bypass $cookie_nocache $arg_nocache; # add cache bypass option
 	}
 
+	location @subdomain {
+		include /etc/nginx/conf.d/include/cors;
+
+		limit_conn downloads_by_ip 100; # ddos protection: max 100 downloads at a time
+		add_header Cache-Control "public, max-age=86400"; # allow consumer to cache response
+
+		proxy_read_timeout 600;
```
ro-tex, line 221: 10 minutes is a long time. Is there a reason not to use the default 60s?

peterjanbrone: no reason to be extra tight with timeouts here - not sure why we decided originally to have it at 600s but if it aint broken don't change it :) we can always revise this later but no need to go ahead and change this in this PR

ro-tex (approved):

docker/nginx/conf.d/client.conf
```diff
@@ -21,6 +21,10 @@ server {
 	listen 80 default_server;
 	listen [::]:80 default_server;
 
+	# parse subdomain into custom variable
+	server_name ~^([a-z0-9]{55}).*?;
+	set $subdomain $1;
```
kwypchlo, line 26: Can you add a comment on what is the subdomain going to be? that is can be base32 encoded skylink

kwypchlo (approved):

PR #330, opened 2020-08-14
Extend web portal health check

MSevey:
This MR is the first step towards #348  

The scope of this MR is the following:
1: Break out checks into basic and verbose. Basic checks are the existing upload/download checks that run every 5 minutes and the verbose checks are to be run daily at 3am
2. Add a number of verbose checks that verify known skyfiles. These skylink verification tests are a way of managing compatibility during upgrades. 
3. Added `critical` flag to the db entries for the health check results. 

**NOTE:** None of the verbose checks are currently listed as critical. This can be changed in the future.


health-check/schedule.js
```diff
@@ -1,16 +1,24 @@
...
-  entry.checks = await Promise.all(checks.map((check) => new Promise(check)));
+  entry.checks = await Promise.all(basicChecks.map((check) => new Promise(check)));
 
   db.get("entries").push(entry).write();
 });
 
+// execute the verbose health-check script once a day at 3am
+const verboseJob = schedule.scheduleJob("* * 3 * * *", async () => {
+  // TODO: Need to fill in based on health check
+  await Promise(verboseCheck);
```
MSevey, line 17: update to array of checks

packages/health-check/src/verboseChecks.js
```diff
@@ -0,0 +1,283 @@
...
+        // Check if the metadata is valid by checking against the known
+        // hash
+        let validMetadata = hash(res.header["skynet-file-metadata"]) === linkInfo.metadataHash;
+        // Redetermine if the Skylink is up based on the results from the body
+        // and metadata hash checks
+        up = up && validBody && validMetadata;
+      }
+
+      // Return the entry information
+      done({
```
MSevey, line 256: @kwypchlo currently none of these are critical checks. How do we get the results of them so that we can verify the results when starting a server and testing code such as on the dev server?

package.json
```diff
@@ -1,6 +1,6 @@
 {
-    "private": true,
-    "workspaces": [
-        "packages/*"
-    ]
+  "private": true,
+  "workspaces": [
+    "packages/*"
+  ]
```
kwypchlo, line 5: do not reformat that please, it's excluded from prettier because npm manages that file

packages/health-check/src/schedule.js
```diff
@@ -1,16 +1,28 @@
...
   const entry = { date: new Date().toISOString(), checks: [] };
 
-  entry.checks = await Promise.all(checks.map((check) => new Promise(check)));
+  entry.checks = await Promise.all(basicChecks.map((check) => new Promise(check)));
 
   db.get("entries").push(entry).write();
 });
 
+// execute the verbose health-check script once a day at 3am
+const verboseJob = schedule.scheduleJob("* * 3 * * *", async () => {
```
kwypchlo, line 16:
```suggestion
const verboseJob = schedule.scheduleJob("0 3 * * *", async () => {
```

packages/health-check/src/verboseChecks.js
```diff
@@ -0,0 +1,283 @@
...
+  skylinkVerification(done, linkInfo);
+}
+
+// skylinkVerification verifies a skylink against known information provided in
+// the linkInfo.
+function skylinkVerification(done, linkInfo) {
+  const time = process.hrtime();
+
+  // Create the query for the skylink
+  let query = `http://${process.env.PORTAL_URL}/${linkInfo.skylink}?nocache=true`;
```
kwypchlo, line 230:
```suggestion
  const query = `http://${process.env.PORTAL_URL}/${linkInfo.skylink}?nocache=true`;
```

packages/health-check/src/verboseChecks.js
```diff
@@ -0,0 +1,283 @@
...
+      // Determine if the skylink is up. Start with checking if there was an
+      // error in the request.
+      let up = err === null;
+      if (up) {
+        // Check if the response body is valid by checking against the known
+        // hash
+        let validBody = hash(res.body) === linkInfo.bodyHash;
+        // Check if the metadata is valid by checking against the known
+        // hash
+        let validMetadata = hash(res.header["skynet-file-metadata"]) === linkInfo.metadataHash;
```
kwypchlo, line 249:
```suggestion
        const validMetadata = hash(res.header["skynet-file-metadata"]) === linkInfo.metadataHash;
```

packages/health-check/src/verboseChecks.js
```diff
@@ -0,0 +1,283 @@
...
+      // Record the statusCode
+      const statusCode = (res && res.statusCode) || (err && err.statusCode) || null;
+
+      // Determine if the skylink is up. Start with checking if there was an
+      // error in the request.
+      let up = err === null;
+      if (up) {
+        // Check if the response body is valid by checking against the known
+        // hash
+        let validBody = hash(res.body) === linkInfo.bodyHash;
```
kwypchlo, line 246:
```suggestion
        const validBody = hash(res.body) === linkInfo.bodyHash;
```

kwypchlo (changes requested):

packages/health-check/src/schedule.js
kwypchlo: https://crontab.guru/#0_3_*_*_*

packages/health-check/src/verboseChecks.js
kwypchlo: those will be printed to `/health-check`

MSevey: I'll add it to the issue :+1:

package.json
MSevey: I didn't format it directly, must have had an old prettier config. I'll revert.

kwypchlo (approved):

PR #346, opened 2020-08-18
Gollum will now check docker logs instead of the journal.

ro-tex:
After we switched to dockerized `siad` we no longer push the logs into the journal from which Gollum is trying to read them. In order to get them we need to use `docker logs` in some shape or form.

Implementation details:
- since `docker logs` works off an image id, we need to get that dynamically and the best way to do that is to use the container name, e.g. `docker logs $(docker ps -q --filter "name=^sia$")`. This means that we cannot change the container name in the future and this needs to be reflected in the `docker-compose.yml` as a comment.

## Note
Upgrading to this code will require manually editing the crontab. The only change is to remove the "d" in "siad" because we're changing from a service name ("siad") to a container name ("sia").

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,7 +14,6 @@ services:
       dockerfile: Dockerfile
       args:
         branch: v1.5.0
-    # Changing the container name will break the log-checker.py script, so please update that accordingly.
     container_name: sia
     restart: unless-stopped
     environment:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,7 @@ services:
       dockerfile: Dockerfile
       args:
         branch: v1.5.0
+    # Changing the container name will break the log-checker.py script, so please update that accordingly.
     container_name: sia
     restart: unless-stopped
     environment:
--- a/setup-scripts/README.md
+++ b/setup-scripts/README.md
@@ -104,9 +104,7 @@ At this point we have almost everything running, we just need to set up your wal
   > `docker exec caddy caddy reload --config /etc/caddy/Caddyfile`
 - Restarting nginx gracefully after making changes to nginx configs
   > `docker exec nginx openresty -s reload`
-- Checking siad service logs (follow last 50 lines) in non-dockerized environments
-  > `journalctl -f -n 50 --user-unit siad`
-- Checking siad service logs (last hour) in dockerized environments
+- Checking siad service logs (last hour)
   > `docker logs --since 1h $(docker ps -q --filter "name=^sia$")`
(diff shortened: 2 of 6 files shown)


MSevey:
Question: The docker image contains the entire portal set up right? So why would we just call it `sia`? Wouldn't `SkynetPortal` be a better name for the container? 

And if the docker is just `siad` why would we update the name to `sia`? What is the reasoning for it being called just `sia`?

ro-tex:
> Question: The docker image contains the entire portal set up right? So why would we just call it `sia`? Wouldn't `SkynetPortal` be a better name for the container?
> 
> And if the docker is just `siad` why would we update the name to `sia`? What is the reasoning for it being called just `sia`?

The portal is not a single docker image, it's a collection of them. If you take a look at `docker-compose.yml`, each entry under `services` is a separate docker container. The `sia` container contains only `siad` and `siac`.

kwypchlo (changes requested):

kwypchlo:
I wonder whether we could actually include that in a separate container in docker-compose instead of running it in crontab of the host machine - what do you think @ro-tex ? I think that's the last thing that is required to be done on the host beside installing docker, docker-compose and changing the file handles limit.

ro-tex:
> I wonder whether we could actually include that in a separate container in docker-compose instead of running it in crontab of the host machine - what do you think @ro-tex ? I think that's the last thing that is required to be done on the host beside installing docker, docker-compose and changing the file handles limit.

That's a good idea. The crontab is an OK quick-and-dirty solution but it's super fragile. I'll add it.

ro-tex:
> > I wonder whether we could actually include that in a separate container in docker-compose instead of running it in crontab of the host machine - what do you think @ro-tex ? I think that's the last thing that is required to be done on the host beside installing docker, docker-compose and changing the file handles limit.
> 
> That's a good idea. The crontab is an OK quick-and-dirty solution but it's super fragile. I'll add it.

So, several hours later it turned out that in order to have these scripts in a container we'd need to access the log files in a different way because we don't have access to `docker` from within a container (makes sense, that would break the sandboxing). We can potentially read the raw files and manually cut the last X hours of those but since `siad` logs don't have timestamps that will be somewhat challenging.

All in all, this is still a good idea and we should do it at some point but for the moment there are a few obstacles and we can spend this time elsewhere. We talked about it and decided to postpone the dockerization of these scripts. We'll merge the branch as it is and I'll push my current progress after that, so we can continue working on this when we decide it's time.

kwypchlo (approved):

PR #358, opened 2020-08-26
Add uncensored library skylinks to health check

MSevey:
This MR adds some more simple known Skylinks.

### DIFF
--- a/packages/health-check/src/verboseChecks.js
+++ b/packages/health-check/src/verboseChecks.js
@@ -17,6 +17,41 @@ function audioExampleCheck(done) {
   skylinkVerification(done, linkInfo);
 }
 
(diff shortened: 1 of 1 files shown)


MSevey:
@kwypchlo the `Cyprus Run` part of the pipeline I think failed on me last time. What is that? How can I test that locally before pushing?

kwypchlo (approved):

kwypchlo:
> @kwypchlo the `Cyprus Run` part of the pipeline I think failed on me last time. What is that? How can I test that locally before pushing?

We were using siasky.dev as a backend and it seems we play around and break it often. I switched to using siasky.net as a backend for these tests and now they work 👍

PR #360, opened 2020-08-28
Check system load

ro-tex:
Extend the `log-checker` script to also keep an eye on the system load and report in Discord if it exceeds 10.

The value 10 was picked more or less at random. If anyone has a good reason to use another value, I'm fully open to that.

I'm using the 15 minute average, so short spikes won't trigger this. The script is run once per 8 hours, so we can't claim to monitor short spikes anyway. This is only intended to detect long-running problems.

### DIFF
--- a/setup-scripts/log-checker.py
+++ b/setup-scripts/log-checker.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-import discord, sys, traceback, io, os, asyncio
+import discord, sys, traceback, io, os, asyncio, re
 from bot_utils import setup, send_msg
 from datetime import datetime, timedelta
 from subprocess import Popen, PIPE
@@ -25,6 +25,7 @@ bot_token = setup()
 client = discord.Client()
 
 
+# exit_after kills the script if it hasn't exited on its own after `delay` seconds
 async def exit_after(delay):
     await asyncio.sleep(delay)
     exit(0)
@@ -40,12 +41,26 @@ async def on_ready():
 async def run_checks():
     print("Running Skynet portal log checks")
     try:
+        await check_load_average()
         await check_docker_logs()
 
     except: # catch all exceptions
         trace = traceback.format_exc()
         await send_msg(client, "```\n{}\n```".format(trace), force_notify=False)
 
+
+# check_load_average monitors the system's load average value and issues a
+# warning message if it exceeds 10.
+async def check_load_average():
+    uptime_string = os.popen("uptime").read().strip()
+    # pattern = ""
+    if sys.platform == "Darwin":
+        pattern = "^.*load averages: (\d*\.\d*) \d*\.\d* \d*\.\d*$"
+    else:
+        pattern = "^.*load average: (\d*\.\d*), \d*\.\d*, \d*\.\d*$"
+    load_av = re.match(pattern, uptime_string).group(1)
+    if float(load_av) > 10:
+        await send_msg(client, "High system load detected: `{}`".format(uptime_string), force_notify=True)
 
 # check_docker_logs checks the docker logs by filtering on the docker image name
 async def check_docker_logs():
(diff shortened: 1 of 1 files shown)


setup-scripts/log-checker.py
```diff
@@ -40,13 +41,27 @@ async def on_ready():
...
 
+# check_load_average monitors the system's load average value and issues a
+# warning message if it exceeds 10.
+async def check_load_average():
+    uptime_string = os.popen("uptime").read().strip()
+    # pattern = ""
+    if sys.platform == "Darwin":
+        pattern = "^.*load averages: \d*\.\d* \d*\.\d* (\d*\.\d*)$"
+    else:
+        pattern = "^.*load average: \d*\.\d*, \d*\.\d*, (\d*\.\d*)$"
```
kwypchlo, line 60: what are those numbers exactly?

ro-tex: The average number of processes either using CPU or waiting for it over the last 1, 5, and 15 minutes.

Good article: http://www.brendangregg.com/blog/2017-08-08/linux-load-averages.html
Also Wikipedia: https://en.wikipedia.org/wiki/Load_(computing)

kwypchlo: Can we indicate that in the message?

setup-scripts/log-checker.py
```diff
@@ -40,13 +41,27 @@ async def on_ready():
...
+    load_av = re.match(pattern, uptime_string).group(1)
```
m-cat, line 61:
```suggestion
    load_av = re.match(pattern, uptime_string).group(3)
```

I think it should be match 3 if we want to get the 15 minute load average

m-cat (approved):

setup-scripts/log-checker.py
kwypchlo: Can we add some context to those in the message @ro-tex ?
Maybe `1.73 (1 min), 0.60 (5 min), 7.98 (15 min)` or sth like that?

ro-tex: That would be quite redundant. It's a little like adding context to a time string: `duration: 1:34:00` -> `duration: 1 (hour) 34 (minutes) 0 (seconds)`.

m-cat: Seems better to report the output of `uptime` verbatim, however I would make it clear in the message that this string comes from `uptime`.

kwypchlo: I wasn't aware of the `uptime` syntax up until just now compared to a common thing like a clock.

setup-scripts/log-checker.py
```diff
@@ -40,13 +41,27 @@ async def on_ready():
...
 
 
+# check_load_average monitors the system's load average value and issues a
+# warning message if it exceeds 10.
+async def check_load_average():
+    uptime_string = os.popen("uptime").read().strip()
+    # pattern = ""
+    if sys.platform == "Darwin":
+        pattern = "^.*load averages: \d*\.\d* \d*\.\d* (\d*\.\d*)$"
+    else:
```
MSevey, line 59: Are all other OS the same syntax, and it is really just Darwin that is different? 
If this is trying to distinguish between just Mac and Linux I would change this to specifically check those two OS and error otherwise.

ro-tex: I *think* all Linux distros use the same format. They basically share the same code. But Macs are not Linux, they are FreeBSD (or they were), so they use a different implementation.

setup-scripts/log-checker.py
ro-tex: You are right, it isn't the same thing. It requires some arcane knowledge to understand what it means and it also requires some quite specific knowledge about the system in order to assess whether it's a problem, e.g. a load average of 5 on a single core system is a bad thing but it's perfectly fine on a quad-core hyper-threaded CPU.

I think I'm with Marcin on this - providing the `uptime` output verbatim should be fine. I also agree on the note that we need to make it clear that this is coming from that specific command. I'll make those changes.

m-cat (approved):

kwypchlo (approved):

PR #362, opened 2020-08-31
fix byte range caching

ChrisSchinnerl:
This fixes video streaming again.
Instead of trying to download the whole file to put it in the cache, nginx will just download ranges of 1MB now.
If multiple requests try to access an uncached slice at the same time, only the first one will fetch the whole sector. The remaining ones will be forwarded to siad directly.


ChrisSchinnerl (commented):
lgtm

PR #363, opened 2020-08-31
Add Skapps to health check by setting the max http header size

MSevey:
This MR sets the `--max-http-header-size` to the same as the handshake integration and then adds the uniswap and develop momentum skylinks to the health checks.


packages/health-check/src/verboseChecks.js
```diff
@@ -551,6 +370,59 @@ function uniswapIndexFileCheck(done) {
...
+
+// uniswapHNSResolverRedirectCheck returns the result of trying to download the skylink
+// for the Uniswap Application via the HNS resolver endpoint without the
+// trailing slash
+function uniswapHNSResolverRedirectCheck(done) {
+  const linkInfo = {
+    description: "Uniswap HNS Resolver Redirect",
+    skylink: "hnsres/doesn",
+    bodyHash: "44a3f0f56861ae841a6cb19cb0b3edf98ad610f8",
+    metadata: null,
```
MSevey, line 420: @kwypchlo these don't have skyfile metadata defined so the response is `undefined`. Will setting this to `null` work with the `lodash.IsEqual`?

kwypchlo: Nope, but you might want to set `metadata: undefined` and it should work in theory :)

kwypchlo: Why doesn't it have the metadata though? Is this correct?

packages/health-check/src/metadata.js
```diff
@@ -0,0 +1,1043 @@
+const developMomentumMetadata = {
```
kwypchlo, line 1: Can you convert this file to multiple `.json` files and put them in `fixtures` directory?

packages/health-check/Dockerfile
```diff
@@ -9,3 +9,4 @@ COPY src/* src/
 EXPOSE 3100
 ENV NODE_ENV production
 CMD [ "node", "src/index.js" ]
+CMD [ "node", "--max-http-header-size=64000", "src/index.js" ]
```
kwypchlo, line 12: remove the old `CMD` from the previous line

kwypchlo (changes requested):

MSevey:
update `/doesn` url to `uniswap-dex`

packages/health-check/src/verboseChecks.js
```diff
@@ -69,13 +70,42 @@ function dappExampleCheck(done) {
   skylinkVerification(done, linkInfo);
 }
 
+const developMomentumBodyHash = "53b44a9d3cfa9b3d66ce5c29976f4383725d3652";
+const developMomentumMetadata = JSON.parse(fs.readFileSync("./fixtures/developMomentumMetadata.json"));
```
kwypchlo, line 74:
```suggestion
const developMomentumMetadata = require("./fixtures/developMomentumMetadata.json");
```

packages/health-check/src/verboseChecks.js
```diff
@@ -228,134 +258,16 @@ function skyBinRedirectCheck(done) {
   skylinkVerification(done, linkInfo);
 }
 
+const skyGalleryBodyHash = "077e54054748d278114f1870f8045a162eb73641";
+const skyGalleryMetadata = JSON.parse(fs.readFileSync("./fixtures/skygalleryMetadata.json"));
```
kwypchlo, line 262:
```suggestion
const skyGalleryMetadata = require("./fixtures/skygalleryMetadata.json");
```

kwypchlo (changes requested):

packages/health-check/src/verboseChecks.js
MSevey: well isn't that fancy. I assumed it was going to be something neat like this but I didn't know what to google haha.

kwypchlo (approved):

kwypchlo (approved):

PR #371, opened 2020-09-04
add safe stop and restart sia scripts

kwypchlo:
- move scripts to `/scripts` directory out of `/setup-scripts`
- new scripts:
  - `portal-down.sh` - stops the portal by turning off the healthcheck and counts down 5 minutes
  - `portal-up.sh` - starts the portal again by turning on the healthcheck again
  - `portal-upgrade.sh` - stops the portal, rebuilds all container (no-cache) and starts it again
- removed a log checker script that was not used any more


scripts/portal-upgrade.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e # exit on first error
+
+# put the server down for maintenance
+. ${cwd}/portal-down.sh
```
ro-tex, line 6: If I hadn't read the code I'd be more than likely to either call this script from the home dir or just alias something to its full path. Using `${cwd}` is not safe.

scripts/portal-upgrade.sh
```diff
@@ -0,0 +1,14 @@
...
+
+# rebuild and restart all docker containers
+docker-compose build --no-cache
+docker-compose down
+docker-compose up -d
+
+# enable the server again
+. ${cwd}/portal-up.sh
```
ro-tex, line 14: The same about all of these. Can we use a path, relative to the position of the script rather than the CWD? And then pass it to `docker-compose` or just `cd` to it and then `cd -`?

scripts/portal-upgrade.sh
kwypchlo: you are right! I totally forgot to add this 👇  when refactoring, adding it back

```
# get current working directory (pwd doesn't cut it)
cwd=$(cd -P -- "$(dirname -- "$0")" && pwd -P)```

scripts/portal-upgrade.sh
kwypchlo: responded in https://github.com/NebulousLabs/skynet-webportal/pull/371/files#r488060349

peterjanbrone (approved):
LGTM 👍  probably worth dropping a link to this README or something as a pin in our Discord as well

scripts/portal-upgrade.sh
ro-tex: I have no clue what this does but I'll trust that you know what you're doing. 👍

ro-tex (approved):

PR #372, opened 2020-09-04
Add a health checker script to Gollum.

ro-tex:
Add a script that would read the local portal's `/health-check` endpoint and would report any issues that occurred since the last run (by scheduled execution time).

Also adjust some names in `funds-checker.py`, so we don't mistake the output/code for `health-checker.py`.

### DIFF
--- a/setup-scripts/health-checker.py
+++ b/setup-scripts/health-checker.py
@@ -17,8 +17,13 @@ health-checker reads the /health-check endpoint of the portal and dispatches
 messages to a Discord channel.
 """
 
-# The default check interval in hours.
-DEFAULT_CHECK_INTERVAL = 1
+# Get the number of hours to look back in the logs or use 1 as default.
+CHECK_HOURS = 1
+if len(sys.argv) > 3:
+    CHECK_HOURS = int(sys.argv[3])
+
+# Discord messages have a limit on their length set at 2000 bytes. We use
+# a lower limit in order to leave some space for additional message text.
 DISCORD_MAX_MESSAGE_LENGTH = 1900
 
 bot_token = setup()
@@ -69,17 +74,12 @@ async def check_health():
                            force_notify=True)
         return
 
-    # Get the number of hours to look back in the logs or use 1 as default.
-    check_hours = DEFAULT_CHECK_INTERVAL
-    if len(sys.argv) > 3:
-        check_hours = int(sys.argv[3])
-
     # Check the health records.
     failed_records = []
     failed_checks = 0
     failed_critical = 0
     passed_checks_counter = 0
-    time_limit = datetime.now() - timedelta(hours=check_hours)
+    time_limit = datetime.now() - timedelta(hours=CHECK_HOURS)
     for rec in res.json():
         time = datetime.strptime(rec['date'], '%Y-%m-%dT%H:%M:%S.%fZ')
         if time < time_limit:
--- a/setup-scripts/health-checker.py
+++ b/setup-scripts/health-checker.py
@@ -1,8 +1,16 @@
 #!/usr/bin/env python3
 
-import discord, requests, traceback, asyncio, os, sys, json, io
+import asyncio
+import io
+import json
+import os
+import sys
+import traceback
 from datetime import datetime, timedelta
-from bot_utils import setup, send_msg, siad, sc_precision
+
+import discord
+import requests
+from bot_utils import setup, send_msg
 
 """
 health-checker reads the /health-check endpoint of the portal and dispatches
@@ -11,6 +19,7 @@ messages to a Discord channel.
 
 # The default check interval in hours.
 DEFAULT_CHECK_INTERVAL = 1
+DISCORD_MAX_MESSAGE_LENGTH = 1900
 
 bot_token = setup()
 client = discord.Client()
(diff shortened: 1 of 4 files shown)


setup-scripts/health-checker.py
```diff
@@ -0,0 +1,102 @@
...
+        print("[DEBUG] run_checks() failed.")
+        if len(trace) < 1900:
+            await send_msg(client, "```\n{}\n```".format(trace), force_notify=False)
+        else:
+            await send_msg(client, "Failed to run the checks!", file=discord.File(io.BytesIO(trace), filename="failed_checks.log"), force_notify=True)
+
+
+# check_health checks /health-check endpoint and reports recent issues
+async def check_health():
+    print("\nChecking wallet/funds health...")
```
kwypchlo, line 46:
```suggestion
    print("\nChecking portal health status...")
```

setup-scripts/health-checker.py
```diff
@@ -0,0 +1,102 @@
...
+
+    try:
+        hostname = await get_hostname()
+        res = requests.get("http://"+hostname+"/health-check")
+    except: # catch all exceptions
+        trace = traceback.format_exc()
+        print("[DEBUG] check_health() failed.")
+        if len(trace) < 1900:
```
kwypchlo, line 53: Can we set that `1900` as a constant like `DISCORD_MAX_MESSAGE_LENGTH` or sth similar?

setup-scripts/health-checker.py
```diff
@@ -0,0 +1,102 @@
...
+            await send_msg(client, "```\n{}\n```".format(trace), force_notify=False)
+        else:
+            await send_msg(client, "Failed to run the checks!", file=discord.File(io.BytesIO(trace), filename="failed_checks.log"), force_notify=True)
```
kwypchlo:
```suggestion
            await send_msg(client, "Failed to run portal health checks!", file=discord.File(io.BytesIO(trace), filename="failed_checks.log"), force_notify=True)
```

kwypchlo: Also, do we want to put those failures in a different file or is it a universal file for failures?

setup-scripts/health-checker.py
```diff
@@ -0,0 +1,102 @@
...
+        return
+
+    # Get the number of hours to look back in the logs or use 1 as default.
+    check_hours = DEFAULT_CHECK_INTERVAL
+    if len(sys.argv) > 3:
+        check_hours = int(sys.argv[3])
```
kwypchlo, line 62: It would be more readable if we'd have all the input args assignments on top of the file, what do you think?

setup-scripts/health-checker.py
```diff
@@ -0,0 +1,102 @@
...
+    for rec in res.json():
+        time = datetime.strptime(rec['date'], '%Y-%m-%dT%H:%M:%S.%fZ')
+        if time < time_limit:
+            continue
+        for check in rec['checks']:
+            if check['up'] == False:
+                # We append the entire record, so we can get the full context.
+                failed_checks.append(rec)
+                break
+        passed_checks_counter += 1
```
kwypchlo, line 77: Can we also store a name (`name` key) and whether it is critical (`critical` key) of the failed checks alongside and add those to the message? Currently we only have 2 critical checks: "upload_file" and "download_file" and it would be great to have a message distinguish based on whether any of the failures was critical or not.

Also, it would be great if we could also include whether the portal is currently down or not (you can check the status code of the /health-check response) - it's 200 (up) or >= 400 (down).

setup-scripts/health-checker.py
```diff
@@ -0,0 +1,102 @@
...
+            await send_msg(client, "Failed to run the checks!", file=discord.File(io.BytesIO(trace), filename="failed_checks.log"), force_notify=True)
+
+
+# check_health checks /health-check endpoint and reports recent issues
+async def check_health():
+    print("\nChecking wallet/funds health...")
+
+    try:
+        hostname = await get_hostname()
+        res = requests.get("http://"+hostname+"/health-check")
```
kwypchlo: Will this throw on status code >= 400 ? because if portal is down, it will return an error status code but still include the json response and we want to parse it anyway.

setup-scripts/health-checker.py
```diff
@@ -0,0 +1,102 @@
...
+async def get_hostname():
+    if len(sys.argv) > 1:
+        env_file = sys.argv[1]
+    with open(env_file, 'r') as file:
+        for line in file.read().split('\n'):
+            pair = line.split("=")
+            if pair[0] == "HOSTNAME":
+                return pair[1]
+    await send_msg(client, "HOSTNAME not found, cannot check health status", force_notify=True)
+    os.exit(0)
```
kwypchlo: In case `HOSTNAME` is not defined, we could return `localhost` and it should work just fine.

kwypchlo (changes requested):

setup-scripts/health-checker.py
ro-tex: It's just the name of the file you upload on Discord. The name of the file doesn't matter, it's just how it's going to be saved on your disk when you download it.

setup-scripts/health-checker.py
ro-tex: I'm not sure. Here I'm just following the same style MarcinJa used in the original scripts.

If you really prefer it, I can move it up in all scripts or move them to a function that handles that.

setup-scripts/health-checker.py
ro-tex: Adding whether the portal is up or down is really easy, no problem.

Regarding the criticals - maybe we only want to report if critical checks have failed?

But it's not a problem to report on how many criticals have been there.

setup-scripts/health-checker.py
ro-tex: No, it won't throw. It will only throw if the host is not found at all, if the SSL is broken (fixed now with the `verify=False` param) and similar.

setup-scripts/health-checker.py
kwypchlo: We want to know when any of the checks fail but probably want to tag `@skynet-prod` only when critical checks have failed.

setup-scripts/health-checker.py
ro-tex: Sure thing.

setup-scripts/health-checker.py
ro-tex: That's a good idea. Implemented. :)

kwypchlo (approved):

PR #381, opened 2020-09-07
Update code samples for API v2

m-cat:


packages/webapp/src/components/CodeExamples/Code.js
```diff
@@ -39,21 +38,21 @@ import (
...
+var client = skynet.New()
+
 func main() {
 	// upload
-	skylink, err := skynet.UploadFile("./src.jpg", skynet.DefaultUploadOptions)
+	skylink, err := client.UploadFile("./src.jpg", skynet.DefaultUploadOptions)
 	if err != nil {
-		fmt.Printf("Unable to upload: %v", err.Error())
-		return
+		panic("Unable to upload: %v", err.Error())
```
ro-tex, line 48: req: Panic takes a single parameter.

```suggestion
		panic(fmt.Sprintf("Unable to upload: %v", err.Error()))
```

packages/webapp/src/components/CodeExamples/Code.js
```diff
@@ -39,21 +38,21 @@ import (
...
 	}
 	fmt.Printf("Upload successful, skylink: %v\\n", skylink)
 
 	// download
-	err = skynet.DownloadFile("./dst.jpg", skylink, skynet.DefaultDownloadOptions)
+	err = client.DownloadFile("./dst.jpg", skylink, skynet.DefaultDownloadOptions)
 	if err != nil {
-		fmt.Printf("Something went wrong, please try again.\\nError: %v", err.Error())
-		return
+		panic("Something went wrong, please try again.\\nError: %v", err.Error())
```
ro-tex, line 55:
```suggestion
		panic(fmt.Sprintf("Something went wrong, please try again.\nError: %v", err.Error()))
```

ro-tex (changes requested):
Looks good! Just two syntax errors in the `panic` calls in Go.

Btw, I see we have placeholders for PHP and Ruby. I'd be happy to write  a Ruby lib. Just ping me if you think that's a good idea.

m-cat:
Thanks, fixed it so it matches the panics in the docs. The double backward-slashes are needed here.

> Btw, I see we have placeholders for PHP and Ruby. I'd be happy to write a Ruby lib. Just ping me if you think that's a good idea.

It's too much work for us to write, maintain and document for now. The priority going forwards is the skynet-js SDK.

ro-tex (approved):

PR #382, opened 2020-09-07
Add free disk space check to health-checker.py.

ro-tex:
Add free disk space check to health-checker.py.
Move load-average check to health-checker.py.

### DIFF
--- a/setup-scripts/health-checker.py
+++ b/setup-scripts/health-checker.py
@@ -4,6 +4,7 @@ import asyncio
 import io
 import json
 import os
+import re
 import sys
 import traceback
 from datetime import datetime, timedelta
@@ -26,6 +27,12 @@ if len(sys.argv) > 3:
 # a lower limit in order to leave some space for additional message text.
 DISCORD_MAX_MESSAGE_LENGTH = 1900
 
+GB = 1 << 20 # converts from KiB to GiB
+# We are going to issue Discord warnings if the free space on a server falls
+# under this threshold.
+FREE_DISK_SPACE_THRESHOLD = 50 * GB  # 50 GiB
+
+
 bot_token = setup()
 client = discord.Client()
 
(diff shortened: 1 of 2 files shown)


setup-scripts/health-checker.py
```diff
@@ -57,6 +66,48 @@ async def run_checks():
...
+        await send_msg(client, "High system load detected: `uptime: {}`".format(uptime_string), force_notify=True)
+
+
+# check_disk checks the amount of free space on the /home partition and issues
+# a warning message if it's under FREE_DISK_SPACE_THRESHOLD GB.
+async def check_disk():
+    # We check free disk space in 1024 byte units, so it's easy to convert.
+    df = os.popen("df --block-size=1024").read().strip()
+    volumes = {}
+    for line in df.split("\n")[1:]:
```
m-cat, line 88:
```suggestion
    for line in df.splitlines()[1:]:
```

setup-scripts/health-checker.py
```diff
@@ -26,6 +27,12 @@
 # a lower limit in order to leave some space for additional message text.
 DISCORD_MAX_MESSAGE_LENGTH = 1900
 
+GB = 1 << 20 # converts from KiB to GiB
```
m-cat, line 30: I found this line confusing, particularly why does a variable called `GB` convert from KiB to GiB. It made sense after I read `check_disk` but I would rename this variable and/or clarify the comment. Probably just the comment so the following line `50 * GB` continues to make sense.

m-cat (approved):
Minor non-blockers

setup-scripts/health-checker.py
ro-tex: The Python scripts are turning out to be problematic, so we're most probably going to rewrite them in something easier to maintain such as TypeScript. So, for the moment I'll merge this as it and I'll leave some local notes to improve the documentation.

PR #383, opened 2020-09-08
assign static ips to containers

kwypchlo:
We want to assign static ips to services because nginx does not allow dynamic upstream aliases and would require us to remember to restart it every time we restart services to ensure that the alias resolves to a valid (current) ip.


peterjanbrone:
LGTM!

PR #388, opened 2020-09-09
improve health checks

kwypchlo:
- removed `critical: true` from verbose checks
- added new endpoints
  - `/health-check/all` - returns all health check entries
  - `/health-check/recent` - returns health check entries from last 24 hours
- added cli methods that can disable or enable the health check so we don't have to stop the service to fail it on purpose
  - `docker exec health-check cli/disable` - it will write `disabled: true` into state file making `/health-check` return error code but not stopping the service (useful for when we want to turn specific portal off for some time)
  - `docker exec health-check cli/enable` - it will write `disabled: false` into state file to enable the actual health checks again
- replaced uniswap v1 link in verbose checks with new link (old one expired)
- added `max-http-header-size` so we don't fail on uniswap link check due to headers overflow
- changed the `downloadCheck` and verbose checks to use promise api (`async/await` and `.then(...)`) instead of `end(...)` - it was improper usage and threw errors even if it worked
- added a check to ensure that `state` directory exists before initialising lowdb file adapter - it worked on docker because the volume mounting created the directory if it didn't exist but when running locally, since `state` directory doesn't exist, it threw error


packages/health-check/cli/disable
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+
+process.env.NODE_ENV = process.env.NODE_ENV || "production";
```
peterjanbrone, line 3: why?

packages/health-check/cli/enable
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env node
+
+process.env.NODE_ENV = process.env.NODE_ENV || "production";
```
peterjanbrone, line 3: why?

packages/health-check/src/index.js
```diff
@@ -17,7 +17,9 @@ const server = express();
 server.use(bodyparser.urlencoded({ extended: false }));
 server.use(bodyparser.json());
 
-server.get("/health-check", require("./endpointHealthCheck"));
+server.get("/health-check", require("./api/index"));
```
peterjanbrone, line 20: `/health-check/disabled` might be useful

packages/health-check/src/utils.js
```diff
@@ -0,0 +1,8 @@
+// return the time between start and now in milliseconds
+function getTimeDiff(start) {
```
peterjanbrone, line 2: `elapsed` is a better name for this

packages/health-check/cli/disable
kwypchlo: Because otherwise it doesn't make sense, server will only persist the setting if it's `production` environment - check `db.js`, otherwise it's in-memory (for development) and since the memory is not shared, it will not affect the running development process otherwise.

packages/health-check/cli/enable
kwypchlo: Because otherwise it doesn't make sense, server will only persist the setting if it's `production` environment - check `db.js`, otherwise it's in-memory (for development) and since the memory is not shared, it will not affect the running development process otherwise.

packages/health-check/src/checks/verbose.js
```diff
@@ -551,53 +551,47 @@ function uniswapIndexFileCheck(done) {
...
+
+        // Check if the metadata is valid by deep comparing expected value with response
+        const expectedMetadata =
+          response.header["skynet-file-metadata"] && JSON.parse(response.header["skynet-file-metadata"]);
+        if (!isEqual(expectedMetadata, metadata)) {
+          entry.up = false;
+          info.metadata = { valid: false, diff: detailedDiff(expectedMetadata, metadata) };
+        }
+
+        if (!isEmpty(info)) entry.info = info; // add info only if it exists
```
peterjanbrone, line 584: lodash's `isEmpty` is evil 😛  no need to change it though but ideally we get rid of that stupid dependency

packages/health-check/src/checks/verbose.js
```diff
@@ -551,53 +551,47 @@ function uniswapIndexFileCheck(done) {
...
+        const info = {};
+
+        // Check if the response body is valid by checking against the known hash
+        if (hash(response.body) !== bodyHash) {
+          entry.up = false;
+          info.body = { valid: false, hash: { expected: bodyHash, current: hash(response.body) } };
+        }
+
+        // Check if the metadata is valid by deep comparing expected value with response
+        const expectedMetadata =
```
peterjanbrone, line 577: (nit) `JSON.parse(response.header["skynet-file-metadata"] || "")`

packages/health-check/src/checks/verbose.js
```diff
@@ -551,53 +551,47 @@ function uniswapIndexFileCheck(done) {
...
-      }
+    .then(
+      (response) => {
+        const entry = { name, up: true, statusCode: response.statusCode, time: getTimeDiff(time) };
+        const info = {};
+
+        // Check if the response body is valid by checking against the known hash
+        if (hash(response.body) !== bodyHash) {
+          entry.up = false;
+          info.body = { valid: false, hash: { expected: bodyHash, current: hash(response.body) } };
```
peterjanbrone, line 573: we're needlessly hashing it twice here

packages/health-check/src/checks/verbose.js
peterjanbrone, line 577: this is confusing, it's not the `expectedMetadata` but the `actualMetadata`

kwypchlo: that would obscure whether the header is `undefined` or `""` (empty string) and we would want that knowledge in the detailed error info

packages/health-check/src/checks/verbose.js
peterjanbrone, line 573: not sure why `info` has a `valid` field, seems redundant as it's equal to `entry.up` for all cases

packages/health-check/src/checks/verbose.js
kwypchlo: my bad

packages/health-check/src/checks/verbose.js
kwypchlo: yeah doesn't make sense any more

packages/health-check/src/checks/critical.js
```diff
@@ -0,0 +1,48 @@
...
+        critical: true,
+      });
+    });
+}
+
+// downloadCheck returns the result of downloading the hard coded link
+async function downloadCheck(done) {
+  const time = process.hrtime();
+  const skylink = "AACogzrAimYPG42tDOKhS3lXZD8YvlF8Q8R17afe95iV2Q";
+  let statusCode;
```
peterjanbrone, line 29: (nit) more tidy
```
let up;

  try {
        const response = await superagent.get(`http://${process.env.PORTAL_URL}/${skylink}?nocache=true`);
        up = response.statusCode === StatusCodes.OK;
  } catch (error) {
        up = false'
  }

  done({
    name: "download_file",
    up,
    time: getTimeDiff(time),
    critical: true,
  });
```

kwypchlo: we still need `statusCode` in the response

peterjanbrone: it would also be nice to rewrite the download to match the structure of the upload, why isn't that possible? would be much cleaner

peterjanbrone (approved):

packages/health-check/src/checks/critical.js
peterjanbrone: > we still need statusCode in the response

I missed that!

kwypchlo: > it would also be nice to rewrite the download to match the structure of the upload, why isn't that possible? would be much cleaner

because upload is a stream and download is a promise

PR #389, opened 2020-09-10
restructure health-check architecture

kwypchlo:
### Current state of health-check

We push all the health-check entries into `entries` property in the health-check db. Each entry contains a date and a list of checks. Each check can define whether it is critical or not.

We provide a health-check endpoint that gets hit couple of times a minute by AWS load balancer and based on the response status code it either keeps the server as healthy or marks as unhealthy and pulls it our of the load balancer.

### Issue with the current health-check endpoint implementation

We have 2 types of checks that run:
- **basic** checks - they are basic upload and download checks, they run every 5 minutes and if any of those fail we should pull the server out of load balancer
- **verbose** checks - they check all the "popular" skylinks on every hour (probably could be just once per day), any one of them failing should not trigger the load balancer pull the server but we want status of them reported anyway (on discord)

We put all those entries in `entries` array but we don't distinguish between them. They all contain a date when they were taken and list of checks (array or checks).

The idea behind the health-check is:
> if the recent critical health checks fail, we want the endpoint to respond with 503 response code

But since now, any check can be marked as critical and some checks are ran once per 5 minutes and some once an hour, the "most recent check" can be either the basic check or the verbose check. This now has become an issue because on production, once an hour, the health-check would pick the verbose checks result as "most recent". This is wrong on many levels:
- recent basic check gets ignored on every full hour because the health-check endpoint picks the verbose checks
- verbose checks may or may not contain some critical checks, we might get false positives/negatives
- even if we make sure that this endpoint would only pick the basic checks, the idea behind marking a check as critical is broken because then we would ignore the verbose checks that are marked as critical

### A way to fix that

Instead of checks going all to one "entries" prop, we separate them by their purpose:
- `critical: [ ... ]` will contain only the entries critical to health-check, those are the basic upload and download ones ran every 5 minutes
- `verbose: [ ... ]` will contain the checks that are ran every hour, those should not affect the server health-check but should be reported to devs nevertheless (ie to discord)
- drop the concept of being able to mark any check with `critical: true`

Exposed endpoints:
- `/health-check` - it contains a `disabled` flag and `entry` which is just the most recent critical health-check entry
- `/health-check/critical` - all critical entries
- `/health-check/verbose` - all verbose entries
- `/health-check/disabled` - just the `disabled` flag, it was already introduced but it's just here FYI

### TODO

Adjust the python script that reports health state.


packages/health-check/src/api/critical.js
```diff
@@ -0,0 +1,8 @@
+const db = require("../db");
+
+// returns all critical health check entries
+module.exports = (req, res) => {
+  const entries = db.get("critical").orderBy("date", "desc").value();
```
ro-tex, line 5: Let's limit this to the last 24 hours. We don't really care about the health status from a month ago.

Also, let's purge old records from the DB. They won't take a lot of space but we don't need anything older than, say, a month.

packages/health-check/src/api/index.js
```diff
@@ -2,61 +2,75 @@ const { StatusCodes } = require("http-status-codes");
...
+  // check whether the portal has been manually disabled
+  const disabled = getDisabled();
 
   if (disabled) {
     return StatusCodes.SERVICE_UNAVAILABLE;
   }
 
-  // Grab entry element from DB
-  const entry = db.get("entries").orderBy("date", "desc").head().value();
+  // grab one most recent critical entry element from DB
```
ro-tex, line 18:
```suggestion
  // grab the most recent critical entry element from DB
```

packages/health-check/src/api/index.js
```diff
@@ -2,61 +2,75 @@ const { StatusCodes } = require("http-status-codes");
...
-  // Check that every critical check entry is up
-  if (entry && entry.checks.every(({ up, critical }) => up && critical)) {
+  // find out whether every check in the entry is up
+  if (entry && entry.checks.every(({ up }) => up)) {
     return StatusCodes.OK;
   }
 
-  // At least one check failed
+  // in case at least one check failed
   return StatusCodes.SERVICE_UNAVAILABLE;
```
ro-tex, line 27: PP: I would typically flip these around and list all failure modes first and then return OK at the end. This allows you to focus on all the ways in which the negative outcome of the processing can be reached and then, having taken care of all the negative cases, you can safely return the positive one at the end.

```suggestion
// find out if any of the checks failed
if (!entry || entry.checks.any(({ up }) => !up)) {
  return StatusCodes.SERVICE_UNAVAILABLE;
}

// all checks passed!
return StatusCodes.OK;
```

packages/health-check/src/schedule.js
```diff
@@ -18,11 +18,11 @@ const verboseJob = schedule.scheduleJob("0 * * * *", async () => {
...
+  db.get("verbose").push(entry).write();
 });
 
 // Launch Health check jobs
 setTimeout(() => {
-  basicJob.invoke();
+  criticalJob.invoke();
   verboseJob.invoke();
-}, 60 * 1000); // delay for 60s to give other services time to start up
+}, 60 * 0); // delay for 60s to give other services time to start up
```
ro-tex, line 28: Leftover debug changes?

packages/health-check/src/api/index.js
```diff
@@ -2,61 +2,75 @@ const { StatusCodes } = require("http-status-codes");
...
 }
 
 module.exports = (req, res) => {
+  const statusCode = getStatusCode();
+  const timeout = statusCode === StatusCodes.OK ? getAvarageResponseTime() : 0;
+
+  // We want to delay the response for the load balancer to be able to prioritize
+  // servers based on the successful response time of thid endpoint. Load balancer
+  // will pull the server if the response is an error so there is no point in delaying
+  // failures, hence 0 timeout on those.
```
ro-tex, line 68: This is a great approach. Also, thank you for the clear and informative comment! I was staring at the `setTimeout` for a while before I read the comment and a lightbulb lit up. :)

ro-tex (approved):
Some non-blocking comments and suggestions, otherwise LGTM.

ro-tex (changes requested):
Oh, the timeout in `schedule.js` is actually blocking.

packages/health-check/src/api/index.js
```diff
@@ -2,61 +2,75 @@ const { StatusCodes } = require("http-status-codes");
...
-    return 0;
-  }
-
-  // Grab 10 entries from the database as a sample to determine the average
-  // timeout for the server.
+/**
+ * Get the sample of most recent critical entries and
+ * calculate the avarage response time of all of them
+ */
+function getAvarageResponseTime() {
```
peterjanbrone, line 34: typo `getAvarageResponseTime`

packages/health-check/src/api/index.js
```diff
@@ -2,61 +2,75 @@ const { StatusCodes } = require("http-status-codes");
...
-function getTimeout() {
-  if (getStatus() === StatusCodes.SERVICE_UNAVAILABLE) {
-    return 0;
-  }
-
-  // Grab 10 entries from the database as a sample to determine the average
-  // timeout for the server.
+/**
+ * Get the sample of most recent critical entries and
+ * calculate the avarage response time of all of them
```
peterjanbrone, line 32: typo `avarage`

packages/health-check/src/api/index.js
```diff
@@ -2,61 +2,75 @@ const { StatusCodes } = require("http-status-codes");
...
+function getDisabled() {
+  return db.get("disabled").value();
 }
 
 module.exports = (req, res) => {
+  const statusCode = getStatusCode();
+  const timeout = statusCode === StatusCodes.OK ? getAvarageResponseTime() : 0;
+
+  // We want to delay the response for the load balancer to be able to prioritize
+  // servers based on the successful response time of thid endpoint. Load balancer
```
kwypchlo, line 66:
```suggestion
  // servers based on the successful response time of this endpoint. Load balancer
```

packages/health-check/src/api/index.js
```diff
@@ -2,61 +2,75 @@ const { StatusCodes } = require("http-status-codes");
...
-  // timeout for the server.
+/**
+ * Get the sample of most recent critical entries and
+ * calculate the avarage response time of all of them
+ */
+function getAvarageResponseTime() {
+  // get most recent 10 successfull checks for the calculation
   const sample = db
-    .get("entries")
+    .get("critical")
```
peterjanbrone, line 37: not really sure I'm a big fan of the split between `critical` entries and `verbose` ones, why didn't we just `filter` on `critical: true`

packages/health-check/src/api/index.js
```diff
@@ -2,61 +2,75 @@ const { StatusCodes } = require("http-status-codes");
...
 
-// getEntriesSinceYesterday gets the health check entries since yesterday
-function getEntriesSinceYesterday() {
-  const yesterday = new Date();
-
-  yesterday.setDate(yesterday.getDate() - 1);
+/**
+ * Get one, most current critical entry
+ */
+function getCurrentCriticalEntry() {
```
peterjanbrone, line 50: (nit) `getCurrent` -> `getMostRecent` or `getLatest`?

packages/health-check/src/api/index.js
peterjanbrone:
```suggestion
  // servers based on the successful response time of this endpoint. Load balancer
```

packages/health-check/src/schedule.js
kwypchlo: ah yes, good catch
```suggestion
}, 60 * 1000); // delay for 60s to give other services time to start up
```

packages/health-check/src/db.js
```diff
@@ -5,9 +5,9 @@ const Memory = require("lowdb/adapters/Memory");
 
 if (!fs.existsSync("state")) fs.mkdirSync("state");
 
-const adapter = process.env.NODE_ENV === "production" ? new FileSync("state/state.json") : new Memory();
+const adapter = new FileSync("state/state.json");
 const db = low(adapter);
 
-db.defaults({ disabled: false, entries: [] }).write();
+db.defaults({ disabled: false, critical: [], verbose: [] }).write();
```
peterjanbrone, line 11: I've been wondering whether it makes more sense to use `enabled` opposed to `disabled` 😬

peterjanbrone: that, or a timestamp might be nice, seeing as that would flag true/false and give us insights in how long it's been disabled for

peterjanbrone (approved):

packages/health-check/src/api/critical.js
kwypchlo: > Let's limit this to the last 24 hours. We don't really care about the health status from a month ago.

I will limit those to last 24 hours.

> Also, let's purge old records from the DB. They won't take a lot of space but we don't need anything older than, say, a month.

I'd like to keep them for now as is in case we set up a logger integrated with ELK stack that we could feed that information. Once we migrate archival data we will be able to add a purge mechanism.

packages/health-check/src/api/index.js
kwypchlo: The PR description should help with that.

packages/health-check/src/api/index.js
kwypchlo: `getMostRecent` sounds better 👍

packages/health-check/src/schedule.js
```diff
@@ -4,12 +4,12 @@ const { criticalChecks } = require("./checks/critical");
 const { verboseChecks } = require("./checks/verbose");
 
 // execute the critical health-check script every 5 minutes
-const basicJob = schedule.scheduleJob("*/5 * * * *", async () => {
+const criticalJob = schedule.scheduleJob("*/5 * * * *", async () => {
   const entry = { date: new Date().toISOString(), checks: [] };
 
   entry.checks = await Promise.all(criticalChecks.map((check) => new Promise(check)));
```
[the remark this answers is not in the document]
kwypchlo: Actually those are functions and they are getting resolved because they are getting passed to a `Promise` constructor and awaited :)

packages/health-check/src/db.js
kwypchlo: setting disabled to anything truthy (a string with a timestamp is truthy) will coalesce to true so we can definitely do that

packages/health-check/src/api/critical.js
ro-tex: That makes sense.

packages/health-check/src/schedule.js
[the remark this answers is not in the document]
ro-tex: Oh, man... Right... I need to write some JS soon before I've forgotten the rest of it. :)

ro-tex (approved):

PR #396, opened 2020-09-11
Discord notifications for health-check integration

kwypchlo:
### Features

- integrated with Discord via discord.js library, uses DISCORD_TOKEN from environment variables
- sends a hello message on startup (just to know it ran)
- sends a message every time a critical or verbose checks changed state (so for example when we had 20/20 passing and now we have 18/20 passing but also if we go back to 20/20 passing) 
- it will skip unnecessary repetitions of the same message (for example when we have 20/20 passing and nothing is changing but also when we have 18/20 passing and nothing is changing) but will send a message 24 hours from the last message sent anyway just to tell us it's alive

### TODO

- comments

### DIFF
--- a/packages/health-check/src/discord.js
+++ b/packages/health-check/src/discord.js
@@ -2,7 +2,7 @@ const Discord = require("discord.js");
 
 const client = new Discord.Client();
 
-const { DISCORD_BOT_TOKEN } = process.env;
+const { DISCORD_BOT_TOKEN, PORTAL_NAME } = process.env;
 
 const CHANNEL_IDS = {
   "skynet-portal-health-check": "684543533155352599",
--- a/packages/health-check/src/discord.js
+++ b/packages/health-check/src/discord.js
@@ -1,0 +1,39 @@
+const Discord = require("discord.js");
+
+const client = new Discord.Client();
+
+const { DISCORD_BOT_TOKEN } = process.env;
+
+const CHANNEL_IDS = {
+  "skynet-portal-health-check": "684543533155352599",
+};
+
+client.on("ready", () => {
+  sendMessageToHealthCheckChannel("hello there");
+});
+
+(async () => {
+  try {
+    await client.login(DISCORD_BOT_TOKEN);
+  } catch (error) {
+    console.log(`Could not connect to discord server: ${error.message}`);
+  }
+})();
+
+function sendMessage(message, channelName) {
+  const channel = client.channels.cache.get(CHANNEL_IDS[channelName]);
+
+  if (!channel) {
+    console.log(`Channel ${channelName} (id: ${CHANNEL_IDS[channelName]}) not found!`);
+
+    return;
+  }
+
+  channel.send({ content: message });
+}
+
+function sendMessageToHealthCheckChannel(message) {
+  return sendMessage(message, "skynet-portal-health-check");
+}
+
+module.exports = { sendMessage, sendMessageToHealthCheckChannel, CHANNEL_IDS };
--- a/packages/health-check/src/discord.js
+++ b/packages/health-check/src/discord.js
@@ -9,7 +9,7 @@ const CHANNEL_IDS = {
 };
 
 client.on("ready", () => {
-  sendMessageToHealthCheckChannel("hello there");
+  sendMessageToHealthCheckChannel(`${PORTAL_NAME}: reporting for duty!`);
 });
 
 (async () => {
--- a/packages/health-check/src/discord.js
+++ b/packages/health-check/src/discord.js
@@ -1,6 +1,6 @@
-const Discord = require("discord.js");
+const { Client, MessageMentions } = require("discord.js");
 
-const client = new Discord.Client();
+const client = new Client();
 
 const { DISCORD_BOT_TOKEN, PORTAL_NAME } = process.env;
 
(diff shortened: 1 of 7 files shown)


packages/health-check/src/notifyChanges.js
```diff
@@ -0,0 +1,47 @@
+const db = require("./db");
+const { sendMessageToHealthCheckChannel, getRoleByName } = require("./discord");
+
+const { PORTAL_NAME = "PORTAL_NAME not defined" } = process.env;
+
+const notificationTimestamps = { critical: 0, verbose: 0 };
+const notificationUnchangedInterval = 24 * 60 * 60 * 1000; // 24 hours
+
+function notifyChanges(type) {
```
peterjanbrone, line 9: The code is definitely readable but some comments would be nice. A small introduction on what it is you're trying to achieve.

packages/health-check/src/notifyChanges.js
```diff
@@ -0,0 +1,47 @@
+const db = require("./db");
+const { sendMessageToHealthCheckChannel, getRoleByName } = require("./discord");
+
+const { PORTAL_NAME = "PORTAL_NAME not defined" } = process.env;
+
+const notificationTimestamps = { critical: 0, verbose: 0 };
+const notificationUnchangedInterval = 24 * 60 * 60 * 1000; // 24 hours
```
peterjanbrone, line 7: I would make it 8.

packages/health-check/src/notifyChanges.js
```diff
@@ -0,0 +1,47 @@
...
+  const currentPassing = currentEntry.checks.filter(({ up }) => up).length;
+  const currentTotal = currentEntry.checks.length;
+  const sendMessage = () => {
+    notificationTimestamps[type] = Date.now();
+
+    const message = `${PORTAL_NAME}: ${currentPassing} / ${currentTotal} ${type} checks passing`;
+
+    if (currentPassing === currentTotal) return sendMessageToHealthCheckChannel(message);
+
+    const role = getRoleByName("skynet-prod");
```
peterjanbrone, line 23: I think we have to make all of the Skynet related things configurable. This script is not very useful to anyone besides us, unless they tweak it.

peterjanbrone (approved):
It feels a bit rushed. Don't get me wrong I think it's great and very useful, but we should try and keep up the code quality as well. This repo could use a linter for example, there's no documentation or explanation, and it contains references that are specific to our setup.

packages/health-check/src/discord.js
```diff
@@ -0,0 +1,50 @@
+const { Client, MessageMentions } = require("discord.js");
+
+const client = new Client();
+
+const { DISCORD_BOT_TOKEN, PORTAL_NAME = "PORTAL_NAME not defined" } = process.env;
+
+client.on("ready", () => {
+  sendMessageToHealthCheckChannel(`${PORTAL_NAME}: reporting for duty!`);
```
ro-tex, line 8: This happens only once, on portal startup, right?

packages/health-check/src/discord.js
```diff
@@ -0,0 +1,50 @@
...
+    console.log(`Channel ${channelName} not found!`);
+
+    return;
+  }
+
+  channel.send(message);
+}
+
+function sendMessageToHealthCheckChannel(message) {
+  return sendMessage(message, "skynet-portal-health-check");
```
ro-tex, line 37: This should be in a config file as well. I'm with PJ about https://github.com/NebulousLabs/skynet-webportal/pull/396/files#r487000246 - all things should be configurable. Maybe with a config file, maybe we env vars.

packages/health-check/src/discord.js
```diff
@@ -0,0 +1,50 @@
...
+function sendMessage(message, channelName) {
+  const channel = client.channels.cache.find((channel) => channel.name === channelName);
+
+  if (!channel) {
+    console.log(`Channel ${channelName} not found!`);
+
+    return;
+  }
+
+  channel.send(message);
```
ro-tex, line 33: Shouldn't this an async operations and therefore awaited?

packages/health-check/src/notifyChanges.js
```diff
@@ -0,0 +1,47 @@
...
+  if (!currentEntry) return;
+
+  const currentPassing = currentEntry.checks.filter(({ up }) => up).length;
+  const currentTotal = currentEntry.checks.length;
+  const sendMessage = () => {
+    notificationTimestamps[type] = Date.now();
+
+    const message = `${PORTAL_NAME}: ${currentPassing} / ${currentTotal} ${type} checks passing`;
+
+    if (currentPassing === currentTotal) return sendMessageToHealthCheckChannel(message);
```
ro-tex, line 21: We seem to always send the healthy status. Why not follow the same rules as "status unchanged"?

packages/health-check/src/discord.js
kwypchlo: No, this happens when health-check service starts.

ro-tex: Right, my question was more along the lines of is this a permanently running service that we'd start once and have running for a long time or is it going to be a short-lived one that would start up, do the checks, and then exit until it's time for it to run again.

kwypchlo:
Closing this since we are going to pursue different approach

PR #398, opened 2020-09-14
remove reveal and fade effects

kwypchlo:
- removed reveal and fade effects on the webapp to make it seem faster and more responsive
- removed `react-reveal` dependency


MSevey:
With UX changes like this we should get in the habit of submitting a recording of some kind to show that the UX is now as expected . Similar to how we submit asciinema recordings with siac changes.

kwypchlo:
> With UX changes like this we should get in the habit of submitting a recording of some kind to show that the UX is now as expected . Similar to how we submit asciinema recordings with siac changes.

@MSevey good point, let me recreate this PR with a bit of magic

PR #402, opened 2020-09-14
Remove reveal and fade effects

kwypchlo:
- removed reveal and fade effects on the webapp to make it seem faster and more responsive
- removed `react-reveal` dependency


MSevey (approved):
yay for previews!

PR #403, opened 2020-09-15
move handshake proxy to nginx

kwypchlo:
Move handshake proxy endpoint to nginx with help of lua scripts. This allows us to keep it as low level as possible, removing any potential overhead from running requests through node server.

### DIFF
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -138,6 +138,11 @@ server {
 			local json = require('cjson')
 			local hns_domain_name, request_uri_rest = string.match(ngx.var.request_uri, "/hns/([^/?]+)(.*)")
 			local hnsres_res = ngx.location.capture("/hnsres/" .. hns_domain_name)
+
+			if hnsres_res.status ~= ngx.HTTP_OK then
+				ngx.exit(ngx.HTTP_NOT_FOUND)
+			end
+
 			local hnsres_json = json.decode(hnsres_res.body)
 			local skylink_prefix, skylink, skylink_rest = string.match(hnsres_json.skylink, "(sia://)([^/?]+)(.*)")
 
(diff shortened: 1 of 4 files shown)


docker/nginx/conf.d/client.conf
```diff
@@ -127,10 +128,42 @@ server {
...
-		proxy_pass http://handshake-api:3100;
+		set $skylink ''; # placeholder for the raw 46 bit skylink
+		set $rest ''; # placeholder for the rest of the url that gets appended to skylink (path and args)
+		
+		# resolve handshake domain by requesting to /hnsres endpoint and assign correct values to $skylink and $rest
+		access_by_lua_block {
+			local json = require('cjson')
+			local hns_domain_name, request_uri_rest = string.match(ngx.var.request_uri, "/hns/([^/?]+)(.*)")
+			local hnsres_res = ngx.location.capture("/hnsres/" .. hns_domain_name)
+			local hnsres_json = json.decode(hnsres_res.body)
```
peterjanbrone, line 141: You want to handle responses with status != 200 here differently. Lua is debug hell if you don't handle errors properly

docker/nginx/conf.d/client.conf
```diff
@@ -127,10 +128,42 @@ server {
...
+
+			ngx.var.skylink = skylink
+			if request_uri_rest == "" or (request_uri_rest == "/" and skylink_rest ~= "") then
+				ngx.var.rest = skylink_rest
+			else
+				ngx.var.rest = request_uri_rest
+			end
+		}
+
+		# overwrite the Cache-Control header to only cache for 60s in case the domain gets updated
```
peterjanbrone, line 152: What does this mean exactly after we've proxy passed it to localhost again which should handle caching right? Isn't the setting of the header redundant at this point?

peterjanbrone (approved):
I like it!

docker/nginx/conf.d/client.conf
kwypchlo: this is for browser caching - it only sets the header that is consumed by browser

the cache you are referring to is proxy cache inside of nginx and it's not affected by that line

docker/nginx/conf.d/client.conf
kwypchlo: yeah I should probably 404 here

PR #405, opened 2020-09-18
update db before reading it in api

kwypchlo:
We need to make sure db is updated in memory each time we make a request because the underlying file could have been changed (due to other scripts, namely disable and enable cli scripts, modifying it). This is not ideal because most of the time it's redundant but it works for now. Better solution would be to have new private (internal) api endpoint (available only from localhost) that the cli would hit so the database would be modified only by one process.


ro-tex (approved):
LGTM. Might not be the most performant solution but we don't particularly care about performance in this case.

PR #418, opened 2020-09-23
improve nginx config

kwypchlo:
- proxy subdomain and /file locations to our general skylink location so we can drop unnecessary copy & paste logic and correctly count skylink usages from access.log
- include cache status in access log: "MISS", "HIT" or "BYPASS" (possibly "EXPIRED" but I've never seen it)
- improve cache key so whether you hit hns address or subdomain address or add weird args, it will correctly cache proxy response
    - remove `$subdomain` because proxy caching will be handled in different location
    - include `$arg_format` so we properly cache different response formats (don't include all `$args` because it's redundant and also it would mean that the location where we add `attachment=true` to args would get separate cached entries)

### DIFF
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -234,35 +234,14 @@ server {
 	}
 
 	location @subdomain {
-		include /etc/nginx/conf.d/include/cors;
 		include /etc/nginx/conf.d/include/proxy-buffer;
-		include /etc/nginx/conf.d/include/proxy-cache-downloads;
 
-		limit_conn downloads_by_ip 100; # ddos protection: max 100 downloads at a time
-		add_header Cache-Control "public, max-age=86400"; # allow consumer to cache response
-
-		proxy_read_timeout 600;
-		proxy_set_header User-Agent: Sia-Agent;
-		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is correct)
-		proxy_pass http://siad/skynet/skylink/$subdomain/$request_uri;
+		proxy_pass http://127.0.0.1/$subdomain/$request_uri;
 	}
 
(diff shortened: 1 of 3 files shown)


peterjanbrone:
> proxy subdomain and /file locations to our general skylink location so we can drop unnecessary copy & paste logic and correctly count skylink usages from access.log

this in particular is nice :)

docker/nginx/conf.d/client.conf
```diff
@@ -234,35 +234,14 @@ server {
...
-		include /etc/nginx/conf.d/include/proxy-cache-downloads;
-
-		limit_conn downloads_by_ip 100; # ddos protection: max 100 downloads at a time
-		add_header Cache-Control "public, max-age=86400"; # allow consumer to cache response
 
-		proxy_read_timeout 600;
-		proxy_set_header User-Agent: Sia-Agent;
-		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is correct)
-		proxy_pass http://siad/skynet/skylink/$subdomain/$request_uri;
+		proxy_pass http://127.0.0.1/$subdomain/$request_uri;
```
peterjanbrone, line 239: this won't work without altering the regex

docker/nginx/conf.d/client.conf
```diff
@@ -215,7 +215,7 @@ server {
 		proxy_pass http://siad/skynet/skyfile/$dir1/$dir2/$dir3/$dir4$is_args$args;
 	}
 
-	location ~ "^/([a-zA-Z0-9-_]{46}(/.*)?)$" {
+	location ~ "^/(([a-zA-Z0-9-_]{46}|[a-z0-9]{55})(/.*)?)$" {
```
peterjanbrone, line 218: Hm grouping triggered me a bit but I think it's fine.

PR #461, opened 2020-10-13
replace node-schedule with cron to fix cpu usage spikes

kwypchlo:
- [node-schedule](https://www.npmjs.com/package/node-schedule) cases unwanted cpu load for some reason, [cron](https://www.npmjs.com/package/cron) should fix that while maintaining the same functionality
- added [ms](https://www.npmjs.com/package/ms) utility package to manage milliseconds transformations

### DIFF
--- a/packages/health-check/src/schedule.js
+++ b/packages/health-check/src/schedule.js
@@ -1,30 +1,45 @@
-const schedule = require("node-schedule");
+const { CronJob } = require("cron");
+const ms = require("ms");
 const db = require("./db");
 const { criticalChecks } = require("./checks/critical");
 const { verboseChecks } = require("./checks/verbose");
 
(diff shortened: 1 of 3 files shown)


packages/health-check/src/schedule.js
```diff
@@ -1,30 +1,45 @@
-const schedule = require("node-schedule");
+const { CronJob } = require("cron");
+const ms = require("ms");
```
ro-tex, line 2: I was gonna say that adding a new package just to avoid writing `60 * 1000` is a bit too much but it's a 6KB package with some nice functionality. Nice one.

ro-tex (approved):

packages/health-check/src/schedule.js
kwypchlo: :)

PR #490, opened 2020-10-20
Update Code.js

iRyanBell:
Example should use camelCase in Node.


kwypchlo (approved):
Thank you for correcting this mistake!

PR #491, opened 2020-10-21
[PARKED] Sync Skykeys

ro-tex:
[PARKED] This PR will not be merged for the moment until we know more about our future direction.

This PR adds a new container that starts with the rest, waits for `siad`'s API to be available, adds the skykeys it finds in its `/keys/keys.json` file in `siad` and exist.


docker/synckeys/synckeys.go
```diff
@@ -0,0 +1,124 @@
...
+		if err != nil {
+			log.Println("Error while adding skykeys:", err.Error())
+			return err
+		}
+		defer resp.Body.Close()
+
+		b, _ := ioutil.ReadAll(resp.Body)
+		bodyStr := string(b)
+		if resp.StatusCode >= 300 && !strings.Contains(bodyStr, "Skykey ID already exists") {
+			log.Println("Error while adding skykeys:", bodyStr)
```
ChrisSchinnerl, line 70: Just to be safe I think we should not abort but collect the errors. To make sure we don't fail to add all keys just because the first one failed due to a typo or whatever.

Otherwise looks good to me.

docker/synckeys/synckeys.go
```diff
@@ -0,0 +1,124 @@
...
+			return err
+		}
+		fmt.Println("Successfully added!")
+	}
+	return nil
+}
+
+func waitUntilApiAvailable(siadIP, apipass string) error {
```
peterjanbrone, line 78: If this works I guess it is fine, but just FYI docker-compose has a feature called `healthcheck` that you could've used for this. It is made especially for this type of usecase. healthcheck: https://docs.docker.com/compose/compose-file/#healthcheck


You can then have containers depend on the siad service and list the health check as a condition:
```
depends_on:
  siad:
    condition: service_healthy 
```
which is nice I think, and reusable!

docker-compose.yml
```diff
@@ -160,3 +160,21 @@ services:
...
+      context: docker/synckeys
+      dockerfile: Dockerfile
+    container_name: synckeys
+    restart: on-failure
+    logging: *default-logging
+    volumes:
+      - ./docker/data/keys:/keys
+      - ./docker/data/sia:/sia-data
+    environment:
+      - SIAD_IP=10.10.10.10
```
peterjanbrone, line 175: this would also not be necessary using the `healthcheck` suggested below

docker/synckeys/synckeys.go
```diff
@@ -0,0 +1,124 @@
...
+		return nil, err
+	}
+	log.Println("Successfully read keys.")
+	return keys, nil
+}
+
+func apiPass() (string, error) {
+	b, err := ioutil.ReadFile(ApiKeyFile)
+	if err != nil {
+		log.Println("Error while api password:", err.Error())
```
peterjanbrone, line 38:
```suggestion
		log.Println("Error while reading api password:", err.Error())
```

docker/synckeys/synckeys.go
```diff
@@ -0,0 +1,124 @@
...
+	return strings.Trim(string(b), " \n"), nil
+}
+
+func registerKeys(siadIP string, keys []string, apipass string) error {
+	fmt.Printf("Will process %d keys.\n", len(keys))
+	client := http.Client{}
+	reqBody, _ := json.Marshal(map[string]string{})
+	for _, sk := range keys {
+		fmt.Println("Adding key:", sk)
+		reqUrl := fmt.Sprintf("http://%s:9980/skynet/addskykey?skykey=%s", siadIP, url.QueryEscape(sk))
```
peterjanbrone, line 51: do you need the `siadIP` here? Couldn't you just call it `sia`? I thought you can ping containers using their service name.. is that not a thing?

docker/synckeys/synckeys.go
```diff
@@ -0,0 +1,124 @@
...
+		req.SetBasicAuth("", apipass)
+
+		resp, err := client.Do(req)
+		if err != nil {
+			log.Println("Error while adding skykeys:", err.Error())
+			return err
+		}
+		defer resp.Body.Close()
+
+		b, _ := ioutil.ReadAll(resp.Body)
```
peterjanbrone, line 67: Printing the error here might be useful?

peterjanbrone (approved):

docker/synckeys/synckeys.go
ro-tex: Ah, nice! I'll use that in addition to this.

The reason to want to still use this is because I can't always be sure that `siad` is going to start immediately. Maybe this is unnecessary but I don't see how it can hurt us, so I think I'll keep it.

docker/synckeys/synckeys.go
ro-tex: You are right! I didn't know about this. Here's an article: https://docs.docker.com/compose/networking/

docker-compose.yml
kwypchlo, line 175: Can't you use `sia` alias that points to this ip?

docker-compose.yml
```diff
@@ -160,3 +160,21 @@ services:
...
+
+  synckeys:
+    build:
+      context: docker/synckeys
+      dockerfile: Dockerfile
+    container_name: synckeys
+    restart: on-failure
+    logging: *default-logging
+    volumes:
+      - ./docker/data/keys:/keys
```
kwypchlo, line 172: please call the container data directory the same name as the container name (so `synckeys` instead of `keys`) to keep consistency

docker/synckeys/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+ENV GOOS linux
+ENV GOARCH amd64
+ENV CGO_ENABLED 0
+
+ADD synckeys.go /go/src/synckeys/
+RUN go install ./...
+
+FROM scratch
+
+COPY --from=builder /go/bin/synckeys /usr/local/bin/synckeys
```
kwypchlo, line 12: if this container exits after it's done then maybe we could drop the scratch image and just run from golang image?

docker-compose.yml
```diff
@@ -160,3 +160,21 @@ services:
...
+      - ./docker/data/sia:/sia-data
```
kwypchlo, line 173:
```suggestion
      - ./docker/data/sia/apipassword:/sia-data/apipassword:ro
```

mount just the apipassword file instead of the whole sia-data directory (nginx container does that)

packages/webapp/src/components/HomeUpload/HomeUpload.js
```diff
@@ -183,7 +183,7 @@ export default function HomeUpload() {
               className="home-upload-mode-switch link"
               onClick={() => setDirectoryMode(!directoryMode)}
             >
-              {directoryMode ? "⇐ Switch back to uploading files" : "Do you want to upload entire directory?"}
+              {directoryMode ? "⇐ Switch back to uploading files" : "Do you want to upload an entire directory?"}
```
kwypchlo, line 186: already fixed on master, please merge or rebase 👍

kwypchlo (changes requested):
This script only ensures that siad will read and add keys on startup. How do we sync keys between servers or make sure that new keys added on one server would work on another server without restarting the server?

docker-compose.yml
ro-tex: Yes, PJ just told me that's possible. I'm changing it. :)

ro-tex:
> This script only ensures that siad will read and add keys on startup. How do we sync keys between servers or make sure that new keys added on one server would work on another server without restarting the server?

That wasn't the goal of this script. Its only goal was to ensure we have a given set of keys always present on all portals. Adding the keys to a given portal is a manual step. If we want to have that automated we need a  completely different solution.

docker-compose.yml
ro-tex: Nice, will do.

docker/synckeys/Dockerfile
ro-tex: The point of the scratch image is that the resulting container is about 6MB instead of 600MB. The golang image is just a builder that creates a single binary that sits alone in the resulting container.

docker-compose.yml
ro-tex: Sure, good point.

docker/synckeys/synckeys.go
ro-tex: This looks great! The only issue is this line in the docs:
```
Version 3 no longer supports the condition form of depends_on.
```
I'll keep it in mind for the future but for this ad-hoc job I think the function will do.

docker-compose.yml
ro-tex: It turns out I don't even need the `healthcheck` for this! Changed.

kwypchlo:
> > This script only ensures that siad will read and add keys on startup. How do we sync keys between servers or make sure that new keys added on one server would work on another server without restarting the server?
> 
> That wasn't the goal of this script. Its only goal was to ensure we have a given set of keys always present on all portals. Adding the keys to a given portal is a manual step. If we want to have that automated we need a completely different solution.

So the way we would do this, we would add private playdj key there right? It's not a general solution to a generic problem. I wonder whether we should merge this into this docker-compose stack then...

kwypchlo:
closing for the time being

PR #501, opened 2020-10-28
Postgres

### LINKED ISSUE #500: Postgres
## Purpose

I am setting up a Postgres DB to live on each portal node and to replicate data across all nodes with a single master node. The setup has many steps and many little tweaks that need to be done, so I'll keep track of those here.

## Steps
- ~~`docker-compose.yaml` that sets up the two nodes and allows them to communicate~~
- domain name (`pg.siasky.net` for the moment) that points to the current master node
- a mechanism for promoting a replica to master status and reflecting that in the domain address
- ~~set up two nodes in the same docker-compose that replicate~~`
- ~~this replication needs to survive restart/rebuild of the `dc` based on the data on disk~~
- ~~the nodes need to communicate based on passwords, not based on implicit trust. see [.pgpass](https://www.postgresql.org/docs/9.2/libpq-pgpass.html)~~
- ~~the `docker-compose.yaml` shouldn't contain any passwords, those should come from a `.env` file or not be needed altogether~~
- figure out a mechanism for managing the archived WAL files under `mnt/server/archivedir` because those take up a lot of disk space when not managed. Check if the `max_wal_size` option in `postgres.conf` works for those or only applies to the unarchived, current WAL.
- apparently it's not that straightforward to change master/replica via failover. Watch this and figure it out after the initial failover is set up: https://www.youtube.com/watch?v=Pn_vsQOvI_c

ro-tex:
Closes #500


ro-tex:
We're going for MongoDB.

PR #514, opened 2020-11-03
Support skydb in hns

kwypchlo:
- extend /hnsres endpoint to return either skylink or registry entry based on what has been stored in the dns entry (support for `skydb://<publickey>/<datakey>` format - **should be uri-encoded**)
- support parsing registry entry in /hns endpoint

### extending `/hnsres` payload
`/hnsres` depending on the domain entry will now return:
- skylink entry, example for domain configuration 
```
sia://GACSvPi8yHr7s0IALienivoMLnwxy787Ptg5vseNxD46BA
```
```json
{
  "skylink": "GACSvPi8yHr7s0IALienivoMLnwxy787Ptg5vseNxD46BA"
}
```
- registry entry, example for domain configuration:
```
skydb://ed25519%3Ad611373052bb124f23c3376a181e140390d9545666dae333acee2402dfc95f4a/79c05b4b67764ad99a7976a7d2fb1cfce4f196ea217ef0356af042cb5492bd5d
```
```json
{
  "registry": {
    "publickey": "ed25519:d611373052bb124f23c3376a181e140390d9545666dae333acee2402dfc95f4a",
    "datakey": "79c05b4b67764ad99a7976a7d2fb1cfce4f196ea217ef0356af042cb5492bd5d"
  }
}
```

### DIFF
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -151,12 +151,39 @@ server {
 			-- example response: '{"skylink":"sia://XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg"}'
 			local hnsres_json = json.decode(hnsres_res.body)
 
-			-- try to match the skylink with sia:// prefix
-			local skylink, skylink_rest = string.match(hnsres_json.skylink, "sia://([^/?]+)(.*)")
+			if hnsres_json.skylink then
+				-- try to match the skylink with sia:// prefix
+				local skylink, skylink_rest = string.match(hnsres_json.skylink, "sia://([^/?]+)(.*)")
 
-			-- in case the skylink did not match, assume that there is no sia:// prefix and try to match again
-			if skylink == nil then
-				skylink, skylink_rest = string.match(hnsres_json.skylink, "/?([^/?]+)(.*)")
+				-- in case the skylink did not match, assume that there is no sia:// prefix and try to match again
+				if skylink == nil then
+					skylink, skylink_rest = string.match(hnsres_json.skylink, "/?([^/?]+)(.*)")
+				end
+			end
+
+			if hnsres_json.registry then
+				local publickey = hnsres_json.registry.publickey
+				local datakey = hnsres_json.registry.datakey
+
+				-- make a get request to /skynet/registry endpoint with the credentials from text record
+				local registry_res = ngx.location.capture("/skynet/registry?publickey=" .. publickey .. "&datakey=" .. datakey)
+
+				ngx.header.content_type = 'text/plain'
+				ngx.print(registry_res.body)
+
+				-- we want to fail with a generic 404 when /skynet/registry returns anything but 200 OK
+				if registry_res.status ~= ngx.HTTP_OK then
+					ngx.exit(ngx.HTTP_NOT_FOUND)
+				end
+
+				-- since /skynet/registry endpoint response is a json, we need to decode it before we access it
+				local registry_json = json.decode(registry_res.body)
+				-- response will contain hex encoded, we need to decode it
+				local data = (registry_json.data:gsub('..', function (cc)
+					return string.char(tonumber(cc, 16))
+				end))
+
+				skylink = data
 			end
 
 			ngx.var.skylink = skylink
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -180,10 +180,6 @@ server {
 					return string.char(tonumber(cc, 16))
 				end))
 
-				ngx.header.content_type = 'text/plain'
-				ngx.print(data)
-				ngx.exit(ngx.HTTP_OK)
-
 				skylink = data
 			end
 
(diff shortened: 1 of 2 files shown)


docker/nginx/conf.d/client.conf
```diff
@@ -151,12 +151,36 @@ server {
...
+					skylink, skylink_rest = string.match(hnsres_json.skylink, "/?([^/?]+)(.*)")
+				end
+			end
+
+			if hnsres_json.registry then
+				local publickey = hnsres_json.registry.publickey
+				local datakey = hnsres_json.registry.datakey
+
+				-- make a get request to /skynet/registry endpoint with the credentials from text record
+				local registry_res = ngx.location.capture("/skynet/registry?publickey=" .. publickey .. "&datakey=" .. datakey)
```
ro-tex, line 169: Now I get it why you bolded the "should be uri-encoded". This is a little uncomfortable but I don't see any major damage that can happen from it aside from a bad request.

docker/nginx/conf.d/client.conf
```diff
@@ -151,12 +151,36 @@ server {
...
+
+				-- we want to fail with a generic 404 when /skynet/registry returns anything but 200 OK
+				if registry_res.status ~= ngx.HTTP_OK then
+					ngx.exit(ngx.HTTP_NOT_FOUND)
+				end
+
+				-- since /skynet/registry endpoint response is a json, we need to decode it before we access it
+				local registry_json = json.decode(registry_res.body)
+				-- response will contain hex encoded, we need to decode it
```
ro-tex, line 178:
```suggestion:
				-- response will contain a hex encoded skylink, we need to decode it
```

packages/handshake-api/src/index.js
```diff
@@ -22,6 +22,7 @@ const cache = new NodeCache({ stdTTL: 300 }); // cache for 5 minutes
 
 // Match both `sia://HASH` and `HASH` links.
 const startsWithSkylinkRegExp = /^(sia:\/\/)?[a-zA-Z0-9_-]{46}/;
+const registryEntryRegExp = /^skydb:\/\/(?<publickey>[a-zA-Z0-9%]+)\/(?<datakey>[a-zA-Z0-9%]+)$/;
```
ro-tex, line 25: What does the `%` do?

packages/handshake-api/src/index.js
```diff
@@ -36,18 +37,36 @@ const getDomainRecords = async (name) => {
...
-    .find(({ txt }) => txt?.some((entry) => isValidSkylink(entry)));
+    .find(({ txt }) => txt?.some((entry) => isValidSkylink(entry) || isValidRegistryEntry(entry)));
 };
 
-const getSkylinkFromRecord = (record) => {
-  return record?.txt?.find((entry) => isValidSkylink(entry));
+const createResponseFromCompatibleRecord = (record) => {
+  const skylink = record?.txt?.find((entry) => isValidSkylink(entry));
+
+  if (skylink) return { skylink };
```
ro-tex, line 53: In the nginx config we prioritize registry entries over skylinks and I think that's a good idea. Here we do the opposite. I believe we should be prioritizing registry entries but more than that we should be consistent.

packages/handshake-api/src/index.js
kwypchlo: public key looks like `"ed25519:d611373052bb124f23c3376a181e140390d9545666dae333acee2402dfc95f4a"`
uri-encoded looks like `"ed25519%3Ad611373052bb124f23c3376a181e140390d9545666dae333acee2402dfc95f4a"`

so we need to include `%` in character list

docker/nginx/conf.d/client.conf
kwypchlo: Actually I think we should revisit the portal code at some point to ensure all the user input is properly escaped, sanitised  or regex matched. It's kind of a MVP implementation while we are still experimenting with the features.

docker/nginx/conf.d/client.conf
kwypchlo:
```suggestion
				-- response will contain a hex encoded, we need to decode it
```

packages/handshake-api/src/index.js
kwypchlo: We don't prioritise registry entries over skylinks -  the prority lies in an order of domain configuration entries. We find latest domain configuration entry that matches either a skylink or skydb entry in `findSkynetCompatibleRecord` and it can only contain either a skylink or a skydb entry so the order of if-else in the code doesn't really matter (and actually in both nginx and api, we first check for skylink existence, then we check for registry entry existence).

packages/handshake-api/src/index.js
ro-tex: Oh, I see. Thanks.

packages/handshake-api/src/index.js
ro-tex: OK.

docker/nginx/conf.d/client.conf
ro-tex: Can you open a ticket for that? Otherwise it would surely be forgotten.

kwypchlo: Also, meaningful error messages would help, we now return 404 if anything is not set up correctly. That's a todo on my list at some point.

ro-tex (approved):

PR #527, opened 2020-11-18
Update the README with instructions on initialising the mongo

ro-tex:


README.md
```diff
@@ -15,6 +15,56 @@ List of available parameters:
 
 - `GATSBY_API_URL`: override api url (defaults to location origin)
 
+### MongoDB Setup
+
+Mongo needs a couple of extra steps in order to start a secure cluster.
+
+* Open port 27017 to all nodes that will take part in the cluster.
```
peterjanbrone, line 22:
```suggestion
* Open port 27017 on all nodes that will take part in the cluster.
```

README.md
```diff
@@ -15,6 +15,56 @@ List of available parameters:
...
+* Manually run an initialisation `docker run` with extra environment variables 
+that will initialise the admin user with a password (example below).
+* Manually add a `mgkey` file under `./docker/data/mongo` with the respective 
+secret (see Mongo's keyfile access control for details).
```
peterjanbrone, line 26: could link it https://docs.mongodb.com/manual/tutorial/enforce-keyfile-access-control-in-existing-replica-set/

README.md
```diff
@@ -15,6 +15,56 @@ List of available parameters:
...
+	-v /home/user/skynet-webportal/docker/data/mongo
+```
+Cluster initialisation mongo command:
+```
+rs.initiate(
+  {
+    _id : "skynet",
+    members: [
+      { _id : 0, host : "helsinki.siasky.net:27017" },
+      { _id : 1, host : "us-va-1.siasky.net:27017" },
```
peterjanbrone, line 62: not sure I would explicitly mention these

peterjanbrone (approved):

README.md
ro-tex: I actually meant `to` but I ended up doing `on`, haha. What I mean is that my original idea was to allow connections to 27017 only from the other nodes that are part of this replica set. But there was something wrong with that and the cluster didn't work, so I switched to "everybody can connect to 27017" for the moment. This is not great and we should revisit it later.

README.md
ro-tex: Good point.

ro-tex:
I'm closing this in favour of #554

PR #528, opened 2020-11-18
Add alert check to the python scripts

MSevey:
This MR Adds a check function to the python scripts that queries the siad `daemon/alerts` endpoint and checks for siafile health alerts. 

This check will send a message to discord once a day unless there are Critical or Error alerts found as those should be addressed immediately.

### DIFF
--- a/setup-scripts/health-checker.py
+++ b/setup-scripts/health-checker.py
@@ -46,6 +46,7 @@ async def run_checks():
         await check_load_average()
         await check_disk()
         await check_health()
+        await check_alerts()
     except:
         trace = traceback.format_exc()
         print("[DEBUG] run_checks() failed.")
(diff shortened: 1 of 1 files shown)


setup-scripts/health-checker.py
```diff
@@ -210,5 +211,60 @@ async def check_health():
...
+
+    message = ""
+    force_notify = False
+
+    if len(critical_alerts) > 0:
+        message += "{} CRITICAL Alerts found! ".format(len(critical_alerts))
+        force_notify = True
+    if len(error_alerts) > 0:
+        message += "{} Error Alerts found! ".format(len(error_alerts))
+        force_notify = True
```
ro-tex, line 254: If we add these there will be a lot of duplication between these alerts and what's reported based on the logs.
I'm not sure what's the best approach. Would everything interesting be reported by these alerts? If so, we can use that and stop monitoring the log file. OTOH, the log file is available even when `siad` is struggling for some reason (but not when it's completely dead because then the container is down).

Or maybe we can scan these alerts only for specific problems we want to monitor and rely on all other issues being logged as well?

setup-scripts/health-checker.py
```diff
@@ -210,5 +211,60 @@ async def check_health():
...
+    
+    message += "{} Warning Alerts found. ".format(len(warning_alerts))
+    message += "{} SiaFile Alerts found. ".format(len(siafile_alerts))
```
ro-tex, line 257: `SiaFile Alerts` is consistent with the rest of the messaging but might not be very clear in the context we want to have this running. We only need this to report the number of files with bad health, so we can monitor if they are increasing or decreasing without reading log files. We're basically optimising for "glanceability" here. :)

setup-scripts/health-checker.py
MSevey: `log.Critical` is different that Critical alerts. 

`log.Critical` is used when something terrible has happened and siad shouldn't continue.

Critical Alerts are alerts that the user should be aware of and they need to take specific action on to resolve as siad can't resolve them on their own.   

Same idea with alert errors, they shouldn't be overlap with the logs check. The logs check is scanning stdout right? not the actual `.log` files of the siad instance correct? 

Also I quickly checked, there is only 1 Error alert and 6 Critical Alerts. 5 of the critical alerts are for hosts so our portals would not see those ever.

So then the Error and Critical alerts that this check would be picking up are related to contract formation and renewal which currently don't have `log.Criticals`. They do log to the `contractor.log` file with `log.Debugln("SEVERE")` but since it is not `log.Severe` the log check wouldn't pick it up. 

All to say, this isn't duplicating anything that the log check should be notifying us about unless I'm mistaken in how the log check is functioning.

setup-scripts/health-checker.py
MSevey: yea fair, I'll update :+1:

setup-scripts/health-checker.py
ro-tex: > Same idea with alert errors, they shouldn't be overlap with the logs check. The logs check is scanning stdout right? not the actual .log files of the siad instance correct?

Yes, that's correct. Technically, I believe it scans both stdout and errout but it's all the same in this context.

Alright, great, then we're getting more visibility without overlap!

ro-tex (approved):

PR #529, opened 2020-11-19
Ratelimit

peterjanbrone:
This PR adds a mechanism to rate-limit upload requests by keeping a "blocklist" that specifies heavy uploader IP addresses.
It works by defining a `$limit_key` depending on whether the current IP is part of that list, if so, the key will be specified and the rate-limited zone, which is obvioulsy the most restrictive, will get applied.

I've chosen to apply rate limits on upload by default, but I've made it so that the default values are very forgiving. I limit both the requests per second and the number of connections severely for users that are on the block list.
E.g. 1 connection per IP and only 10 requests per minute.

By default it's 10 connections per IP, and 10 requests per second, with an allowed burst of 100 requests, which I think is fine.
Unfortunately we could not use `rate_limit`, as that only allows to limit the rate of response transmission.

This was not tested yet! So do not merge it yet.

### DIFF
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -3,7 +3,7 @@ proxy_cache_path /data/nginx/cache levels=1:2 keys_zone=skynet:10m max_size=10g 
 # ratelimit specified IPs
 geo $limit {
 	default 0;
-	include /etc/nginx/conf.d/include/ratelimited;
+	include /etc/nginx/conf.d/include/ratelimited/*;
 }
 map $limit $limit_key {
 	0 "";
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -10,8 +10,8 @@ map $limit $limit_key {
 	1 $binary_remote_addr;
 }
 
-limit_req_zone $binary_remote_addr zone=upload_req:10m rate=10r/s;
-limit_req_zone $limit_key zone=upload_req_rl:10m rate=10r/m;
+limit_req_zone $binary_remote_addr zone=uploads_by_ip:10m rate=10r/s;
+limit_req_zone $limit_key zone=uploads_by_ip_throttled:10m rate=10r/m;
 
 limit_conn_zone $binary_remote_addr zone=upload_conn:10m;
 limit_conn_zone $limit_key zone=upload_conn_rl:10m;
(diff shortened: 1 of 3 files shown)


ro-tex (approved):
Looks good, as far as I can tell.

ChrisSchinnerl:
Looking good but we should try it before merging

ChrisSchinnerl (approved):

docker/nginx/conf.d/client.conf
```diff
@@ -1,6 +1,21 @@
...
+	default 0;
+	include /etc/nginx/conf.d/include/ratelimited;
+}
+map $limit $limit_key {
+	0 "";
+	1 $binary_remote_addr;
+}
+
+limit_req_zone $binary_remote_addr zone=upload_req:10m rate=10r/s;
+limit_req_zone $limit_key zone=upload_req_rl:10m rate=10r/m;
```
kwypchlo, line 14: I would suggest using `_throttled` suffix to the zone name instead of `_rl` - it is more descriptive than an abbreviation that is not expanded anywhere else. Also I think `uploads_by_ip` name is more descriptive than `upload_req` - I think we already changed that name following your advise in #148 

```suggestion
limit_req_zone $binary_remote_addr zone=uploads_by_ip:10m rate=10r/s;
limit_req_zone $limit_key zone=uploads_by_ip_throttled:10m rate=10r/m;
```

docker/nginx/conf.d/include/ratelimited
```diff
@@ -0,0 +1,6 @@
+# Add a list of IPs here that should be severely rate limited on upload.
+# Note that it is possible to add IP ranges as well as the full IP address.
+#
+# Examples:
+# 192.168.0.0/24 1;
+# 79.85.222.247 1;
```
kwypchlo, line 6: This looks fine, I am just not super happy that we will be changing a committed file on the servers. I usually want to keep the repo file untouched on the servers if not necessary - ex. using `docker-compose.override.yml` instead of directly modifying `docker-compose.yml`.

kwypchlo (changes requested):

docker/nginx/conf.d/include/ratelimited
peterjanbrone: Hm I didn't think of this. If you include a non existing file in nginx, would it complain you think? Because if not we can just remove this file. I only added it to have an example of the ip ranges, but we can move these comments to where we include the file.

peterjanbrone: We might be able to do something like this:
```
geo $limit {
	default 0;
	include /etc/nginx/conf.d/include/blocklist/*.ips;
}
```
then rename the above file (and move it) to `/etc/nginx/conf.d/include/blocklist/example.ips;`
and then define `/etc/nginx/conf.d/include/blocklist/skynet.ips;` on the server

kwypchlo: I like it but I would actually drop the `.ips` extension, it doesn't tell much but I would definitely move that file you had with readme and add info "create a file and include ips delimitered by new lines like this [example]" and we should be fine.

kwypchlo (approved):

ChrisSchinnerl (approved):

PR #530, opened 2020-11-20
Update Siafile Health Check to parse siac

MSevey:
Original we were using the siad API which doesn't work since the API is not exposed outside of docker.

### DIFF
--- a/setup-scripts/health-checker.py
+++ b/setup-scripts/health-checker.py
@@ -47,6 +47,7 @@ async def run_checks():
         await check_disk()
         await check_health()
         await check_alerts()
+        await check_portal_size()
     except:
         trace = traceback.format_exc()
         print("[DEBUG] run_checks() failed.")
@@ -78,6 +79,7 @@ async def check_disk():
     # We check free disk space in 1024 byte units, so it's easy to convert.
     df = os.popen("df --block-size=1024").read().strip()
     volumes = {}
+    # Iterate over the output, ignoring the header line
     for line in df.split("\n")[1:]:
         fields = list(filter(None, line.split(" ")))
         # -1 is "mounted on", 3 is "available space" in KiB which we want in bytes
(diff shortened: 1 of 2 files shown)


setup-scripts/health-checker.py
```diff
@@ -211,60 +218,118 @@ async def check_health():
...
-
-    alerts =  alerts_json['alerts']
-    critical_alerts =  alerts_json['criticalalerts']
-    error_alerts =  alerts_json['erroralerts']
-    warning_alerts =  alerts_json['warningalerts']
+    # Execute siac alerts and read the response
+    # TODO: is the container name always `sia` for production servers? Is it
+    # only changed to the server name when it it is moved to Maintenance? Will
+    # this just never check the alerts on the maintenance servers?
+    cmd_string = "docker exec sia siac alert"
```
MSevey, line 235: Do we have a docker container variable name? `sia` works for the production portals, but then won't work for the maintenance servers.

I guess I'm assuming this same script is running on the maintenance servers, is that true? This is a good check to be running on both so I want to make sure we do that.

MSevey:
I don't understand why the checks are failing. Logs look like issues with set up, referencing env vars not parsing. Seems unrelated to my PR. 
Did something change in the build process?

setup-scripts/health-checker.py
ro-tex: Yes, we do. The crontab entry that's calling this script looks like this:
```
0 * * * * ~/skynet-webportal/setup-scripts/health-checker.py ~/skynet-webportal/.env sia 1
```
The first parameter is the container name.

On the maintenance server we're not running the health-check script. Maybe we should but we don't right now, so we can worry about that in a future PR.

setup-scripts/health-checker.py
```diff
@@ -211,60 +218,118 @@ async def check_health():
...
 
+    # Split the output by line and check for type of alert and siafile alerts
+    for line in siac_alert_output.split("\n"):
+        # Check for the type of alert
+        if contains_string(lin, critical):
+            num_critical_alerts++
+        if contains_string(lin, error):
+            num_error_alerts++
+        if contains_string(lin, warning):
+            num_warning_alerts++
```
ro-tex, line 260: Python doesn't support the `++` operator.

Also `lin` should be `line`.

setup-scripts/health-checker.py
ro-tex: I added that.

ro-tex (dismissed):
LGTM.

ro-tex (approved):

ro-tex (approved):

setup-scripts/health-checker.py
```diff
@@ -211,59 +232,118 @@ async def check_health():
...
+
+    if num_files > max_files:
+        message += "Portal has {} files! Consider rotating! ".format(num_files)
+        # send notification when above 40% of the limit
+        force_notify = num_files > max_files * 1.4
+    else:
+        message += "Portal has {} files. ".format(num_files)
+
+    # send a message if we force notification, or just once daily (heartbeat) on 1 AM
+    if force_notify and datetime.utcnow().hour == 1:
```
ro-tex, line 344: This `or` was intentional. For certain things we want to be pinged immediately (`force_notify`) and for others - once a day (do not force it). If we want to deviate from that logic let's change the name of the var, so it's clear what's happening.

MSevey: That was a mistake on my end. Didn't mean to change it to an `and`. Also forgot to add the Force Notify to the max files conditionally.

ro-tex (dismissed):

kwypchlo (dismissed):

setup-scripts/health-checker.py
```diff
@@ -211,59 +232,118 @@ async def check_health():
...
-        message += "{} Error Alerts found! ".format(len(error_alerts))
+    if num_critical_alerts > 0:
+        message += "{} CRITICAL Alerts found! ".format(num_critical_alerts)
         force_notify = True
-    
-    message += "{} Warning Alerts found. ".format(len(warning_alerts))
-    message += "{} SiaFiles with bad health found. ".format(len(siafile_alerts))
+    if num_error_alerts > 0:
+        message += "{} Error Alerts found! ".format(num_error_alerts)
+        force_notify = num_error_alerts > num_siafile_alerts
```
MSevey, line 293: I would remove this line. It doesn't make sense. There is currently only 1 error alert in the whole code base. Even if we add more error alerts I don't think this makes sense, we should either alert on error alerts or not. I think it is OK to not alert and just send it in a message.

setup-scripts/health-checker.py
```diff
@@ -211,59 +232,118 @@ async def check_health():
...
 
-    alerts_file = None 
-    if len(alerts) > 0:
-        alerts_file = json.dumps(alerts, indent=2)
+    message += "{} Warning Alerts found. ".format(num_warning_alerts)
```
MSevey, line 295: I think we wanted to update this to be `num_warning_alerts - num_siafile_alerts` since siafile alerts are warning level.

setup-scripts/health-checker.py
kwypchlo: removed

setup-scripts/health-checker.py
kwypchlo: I think that's fine to have the total warning alerts number here follow by how many of these are the bad health warnings

kwypchlo (approved):

ro-tex (approved):

PR #534, opened 2020-11-24
add option to pass a list of skylinks for blocklist

kwypchlo:
To improve skylink blocklist update process, we can now pass a file path as an argument to scripts/blocklist-skylink.sh

ex. `scripts/blocklist-skylink.sh all_skylinks_to_block.txt`

Script will ssh into each server, block all skylinks and then prune the nginx cache once.

### DIFF
--- a/scripts/blocklist-skylink.sh
+++ b/scripts/blocklist-skylink.sh
@@ -14,13 +14,15 @@ skylinks=()
 if test -f "$1"; then
     OLDIFS=$IFS
     IFS=','
+    line_number=1
     while read line
     do
         if [[ $line =~ ([a-zA-Z0-9_-]{46}) ]]; then
             skylinks+=("$BASH_REMATCH")
         else
-            echo "Incorrect skylink: $line" && exit 1
+            echo "Incorrect skylink at line ${line_number}: $line" && exit 1
         fi
+        let line_number+=1
     done < $1;
     IFS=$OLDIFS
 else
--- a/scripts/blocklist-skylink.sh
+++ b/scripts/blocklist-skylink.sh
@@ -44,7 +44,7 @@ do
     ######################################################
     # purge nginx cache after all the skylinks are blocked
     ######################################################
-    ssh -q -t user@${server} 'rm -rf /home/user/skynet_webportal/docker/data/nginx/cache'
+    ssh -q -t user@${server} 'docker exec nginx sh -c "rm -rf /data/nginx/cache/*"'
     echo ".... 🗑️ Pruned nginx cache on ${server}"
 done
 
--- a/scripts/blocklist-skylink.sh
+++ b/scripts/blocklist-skylink.sh
@@ -3,7 +3,7 @@
 set -e # exit on first error
 
 if [ -z "$1" ]; then
-    echo "Please provide either a skylink or file with skylinks separated by blank lines" && exit 1
+    echo "Please provide either a skylink or file with skylinks separated by new lines" && exit 1
 fi
 
 #########################################################
--- a/scripts/blocklist-skylink.sh
+++ b/scripts/blocklist-skylink.sh
@@ -3,14 +3,49 @@
 set -e # exit on first error
 
 if [ -z "$1" ]; then
-    echo "Please provide a skylink to blocklist" && exit 1
+    echo "Please provide either a skylink or file with skylinks separated by blank lines" && exit 1
 fi
 
+#########################################################
+# read either a file containing skylinks separated by new 
+# lines or a single skylink and put them in an array
+#########################################################
+skylinks=()
+if test -f "$1"; then
+    OLDIFS=$IFS
+    IFS=','
+    while read line
+    do
+        if [[ $line =~ ([a-zA-Z0-9_-]{46}) ]]; then
+            skylinks+=("$BASH_REMATCH")
+        else
+            echo "Incorrect skylink: $line" && exit 1
+        fi
+    done < $1;
+    IFS=$OLDIFS
+else
+    skylinks=("$1") # just single skylink passed as input argument
+fi
+
+####################################################
+# iterate through all servers and block the skylinks
+####################################################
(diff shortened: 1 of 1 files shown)


scripts/blocklist-skylink.sh
```diff
@@ -3,14 +3,49 @@
 set -e # exit on first error
 
 if [ -z "$1" ]; then
-    echo "Please provide a skylink to blocklist" && exit 1
+    echo "Please provide either a skylink or file with skylinks separated by blank lines" && exit 1
```
ro-tex, line 6:
```suggestion:
   echo "Please provide either a skylink or file with skylinks separated by new lines" && exit 1
```

scripts/blocklist-skylink.sh
```diff
@@ -3,14 +3,49 @@
...
+skylinks=()
+if test -f "$1"; then
+    OLDIFS=$IFS
+    IFS=','
+    while read line
+    do
+        if [[ $line =~ ([a-zA-Z0-9_-]{46}) ]]; then
+            skylinks+=("$BASH_REMATCH")
+        else
+            echo "Incorrect skylink: $line" && exit 1
```
ro-tex, line 22: Would it make sense to also print the line on which the bad skylink sits? That would help a little with vim. Although we can also just search for it... I don't know, what do you think?

scripts/blocklist-skylink.sh
peterjanbrone, line 6: (nit) "separated by blank lines" is confusing

scripts/blocklist-skylink.sh
```diff
@@ -3,14 +3,49 @@
...
+        fi
+    done < $1;
+    IFS=$OLDIFS
+else
+    skylinks=("$1") # just single skylink passed as input argument
```
peterjanbrone, line 27: no validation on single skylink?

ro-tex (approved):

scripts/blocklist-skylink.sh
```diff
@@ -3,14 +3,49 @@
...
 fi
 
+#########################################################
+# read either a file containing skylinks separated by new 
+# lines or a single skylink and put them in an array
+#########################################################
+skylinks=()
+if test -f "$1"; then
+    OLDIFS=$IFS
+    IFS=','
```
peterjanbrone, line 16: not following why you adjust `IFS` to be a comma, but never mention comma separation

scripts/blocklist-skylink.sh
```diff
@@ -3,14 +3,49 @@
...
+    # iterate throught all skylinks and add each one to blocklist
+    #############################################################
+    for skylink in "${skylinks[@]}";
+    do
+        echo ".. ⌁ Blocking skylink ${skylink} on ${server}"
+        ssh -q -t user@${server} 'docker exec sia siac skynet blocklist add '$skylink''
+    done
+
+    ######################################################
+    # purge nginx cache after all the skylinks are blocked
```
peterjanbrone, line 45: Well that's unfortunate ^^ not sure how the `nginx/cache` folder structure looks but perhaps we could eventually prune the individual skylinks. I understand though it not being a priority at all, it would be annoying though if we have to run this script several times a day, which I guess might happen in the short to mid term.

peterjanbrone (approved):

scripts/blocklist-skylink.sh
ChrisSchinnerl: This might be a solution: https://serverfault.com/questions/493411/how-to-delete-single-nginx-cache-file

So it might be possible to update the cache entry for the blocked files by just requesting them with the nocache flag.

ChrisSchinnerl (commented):
I agree with the other comments. Apart from that it looks good.

scripts/blocklist-skylink.sh
kwypchlo:
```suggestion
    echo "Please provide either a skylink or file with skylinks separated by new lines" && exit 1
```

kwypchlo: it should say new lines, my bad :)

ro-tex: Oh, I see. When typing that up I shouldn't add `:` after `suggestion`. But still, it would be easier to remember that there's a button. Thanks. :)

scripts/blocklist-skylink.sh
kwypchlo: Nah, siac will validate the input when you pass it. If it's single skylink then it will just exit with an error and you can fix the skylink. For a list of skylinks it's important to fail early because the list can be 50 skylinks and it would fail somewhere along the way and you would repeat the whole process again.

scripts/blocklist-skylink.sh
kwypchlo: me neither, it just doesn't work without it and all SO examples have it there, it can be empty fwiw and it works too but it has to be set

scripts/blocklist-skylink.sh
kwypchlo: I found this https://scene-si.org/2017/01/08/improving-nginx-lua-cache-purge/ - should be even better, I will make an issue about it

kwypchlo: https://github.com/NebulousLabs/skynet-webportal/issues/536

scripts/blocklist-skylink.sh
kwypchlo: added

scripts/blocklist-skylink.sh
peterjanbrone: (works for me)

ro-tex (approved):

scripts/blocklist-skylink.sh
peterjanbrone: fwiw I found:
```
#!/bin/bash
input="/path/to/txt/file"
while IFS= read -r line
do
  echo "$line"
done < "$input"
```
but don't mind if we have it in there - it's just confusing... perhaps set it to '' at least

peterjanbrone (approved):

PR #535, opened 2020-11-24
do not force notification on error alerts

kwypchlo:
Do not force notification on error alerts until we fix the health loop and clear the low file health errors.


kwypchlo:
I'm going to change that in https://github.com/NebulousLabs/skynet-webportal/pull/530

PR #541, opened 2020-11-26
add server override includes and redirect legacy endpoints

kwypchlo:
- added wildcard include at the end of server block in nginx client.conf to allow custom includes that are not commited to repo
- permanently redirected /stats and /portals to /skynet/stats and /skynet/portals respectively as a part of config cleanup (already did that for blocklist)

### DIFF
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -50,6 +50,8 @@ server {
 	client_max_body_size 128k;
 
 	# legacy endpoint rewrite
+	rewrite /portals /skynet/portals permanent;
+	rewrite /stats /skynet/stats permanent;
 	rewrite /skynet/blacklist /skynet/blocklist permanent;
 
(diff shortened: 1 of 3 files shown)


docker/nginx/conf.d/client.conf
```diff
@@ -308,4 +291,7 @@ server {
 
 		proxy_pass http://127.0.0.1/$uri?attachment=true&$args;
 	}
+
+	# include custom locations, specific to the server
+	include /etc/nginx/conf.d/locations/*;
```
peterjanbrone, line 296: am I missing something here? are we doing `locations` or `server-override`

kwypchlo:
```suggestion
	include /etc/nginx/conf.d/server-override/*;
```

ro-tex (approved):

peterjanbrone (approved):

PR #552, opened 2020-12-01
do not ping when server is in maintenance mode

kwypchlo:
do not add @skynet-prod when server is manually disabled - we don't need the ping when server is in maintenance mode

### DIFF
--- a/setup-scripts/health-checker.py
+++ b/setup-scripts/health-checker.py
@@ -200,7 +200,6 @@ async def check_health():
 
     if json_check["disabled"]:
         message += "__Portal manually disabled!__ "
-        force_notify = True
     elif res_check.status_code is not requests.codes["ok"]:
         message += "__Portal down!!!__ "
         force_notify = True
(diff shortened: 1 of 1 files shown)


ro-tex (approved):

setup-scripts/health-checker.py
```diff
@@ -227,7 +226,7 @@ async def check_health():
         failed_records_file = json.dumps(failed_records, indent=2)
 
     # send a message if we force notification, there is a failures dump or just once daily (heartbeat) on 1 AM
-    if force_notify or failed_records_file or datetime.utcnow().hour == 1:
+    if force_notify or json_check["disabled"] or failed_records_file or datetime.utcnow().hour == 1:
```
peterjanbrone, line 229: I don't get it. I thought we wanted to **not** send a message if the portal is manually disabled.
This code makes it so `json_check["disabled"]` is treated essentially the same as `force_notify`
and thus will notify us still. Right?

peterjanbrone (changes requested):
https://github.com/NebulousLabs/skynet-webportal/pull/552/files#r533367160

setup-scripts/health-checker.py
kwypchlo: nope, it will send the message but not add a @skynet-prod ping which is something we don't need every hour

peterjanbrone: Oh I get how `force_notify` works now. Tanks! I thought it had to do with the actual sending of the message, not the notification

peterjanbrone (approved):

PR #553, opened 2020-12-02
improve nginx skylinks cache purging

kwypchlo:
- used a lua script from https://scene-si.org/2017/01/08/improving-nginx-lua-cache-purge/ to add a PURGE method handling that is secured to only local ips, that purges cached entries that contain the skylink
- refactored block script to use the new purging endpoint to purge only the skylinks that we are blocklisting


scripts/blocklist-skylink.sh
```diff
@@ -29,25 +29,17 @@ else
...
         echo ".. ⌁ Blocking skylink ${skylink} on ${server}"
-        ssh -q -t user@${server} 'docker exec sia siac skynet blocklist add '$skylink''
-    done
 
-    ######################################################
-    # purge nginx cache after all the skylinks are blocked
-    ######################################################
-    ssh -q -t user@${server} 'docker exec nginx sh -c "rm -rf /data/nginx/cache/*"'
-    echo ".... 🗑️ Pruned nginx cache on ${server}"
+        ssh -q -t user@${server} "docker exec sia siac skynet blocklist add $skylink && docker exec nginx curl -s -i -X PURGE http://localhost/$skylink | egrep \"^(OK|HTTP|X-)\""
```
ro-tex, line 41: Have you tested this? At first glance it seems like the whole `&& docker exec nginx...` command will be passed as parameters to the `siac` call in the first docker command.

Do we have any reporting of errors? I see you're filtering out the OK status but I don't know what happens with the not-OK statuses.

ro-tex (approved):
LGTM (the `explode` function in lua is a bit unreadable, tbh) aside from the one comment I made. If that's fine then I'm OK with merging this.

scripts/blocklist-skylink.sh
kwypchlo: I did test it out.

Example of successful (one server, skylink valid, redacted):
```bash
> scripts/blocklist-skylink.sh XXX

.. ⌁ Blocking skylink XXX on siasky.xyz
Skynet Blocklist updated
HTTP/1.1 200 OK
X-Purged-Count: 0
OK
✓ All done !
```

Example of failure:  (one server, skylink invalid, redacted):
```bash
> scripts/blocklist-skylink.sh XXXINVALID

.. ⌁ Blocking skylink XXXINVALID on germany.siasky.net
Unable to update skynet blocklist: [POST request error; error parsing skylink: [unable to decode skylink; skylink has incorrect size]]
```

ro-tex: Great, I just wanted to make sure because that linking of docker commands looked very suspicious.

peterjanbrone (approved):
pretty sweet

PR #554, opened 2020-12-04
Accounts

### LINKED ISSUE #511: User accounts: Initial
## Overview

* User registration
* User login
* [optional] Password change and recovery of a lost/forgotten password
* API keys
* History and stats

One PR (at least) per point.

## Details

* These data will be stored in MongoDB. This must be done in such a way that portal operators who don't wish to use user accounts do not need to set up Mongo and still have nice and smooth experience.
   * the entire setup is now completely optional and not including the `docker-compose.accounts.yml` excludes the entire stack
* The Mongo setup needs to be streamlined, so setting up a single DB portal doesn't require manual steps. This might entail default passwords (the DB is not exposed, so the risk is somewhat limited), auto-generation of key files, etc.
* The API key generated for each user should be supplied on each request, so the user is never rate limited.
* The user history should be stored in the DB in such a way that we can easily query:
    * which skylinks are visited by the user (+ when, how many times, etc.)
    * which users have visited a given skylinlk (+ --//--)

kwypchlo:
- you need to manually add `MYSQL_ROOT_PASSWORD` to `.env` on each server (when you set up new, it gets generated with `openssl rand -base64 32`)
- in docker-compose.yml there is hardcoded BASE_URL and KRATOS_BROWSER_URL - you need to change them accordingly until I find a more suitable variable for those
- you need to change all the hardcoded https://siasky.net to the domain of your choice in docker/kratos/config/kratos.yml - I will try to come up with more generic solution for that

## TODOs

- [x] Secure CockroachDB
- [x] Add a good option for overriding the portal's domain in `kratos.yml` (https://www.ory.sh/kratos/docs/reference/configuration/)
- [x] Switch from siasky.xyz to secure/auth/login.siasky.net and set up the domain properly

## Bugs

- [x] Go to sign up, leave the email blank. You get an error that the email is invalid, but now I can't type in the input box for the email anymore. Page refresh does nothing. Have to click back to sign in then back to sign up
- [x] UX request. If we are going to have password validation, can we have a tool tip or something to explain the requirements? I don't like guessing multiple times to see if my password is going to be valid
- [x] Current plan stuck to "Free"
- [x] hide the bandwidth used for now, and in the premium page make sure the copy talks about storage not bandwidth
- [x] Make the "Stripe Customer Portal" link appear like a button and replace the "Subscribe" button
- [x] Rework the "Payments" page to be a proxy for Stripe Customer Portal, at least when the user has a running sub.
- [x] On bad login credentials, the error message shouldn't say "The provided credentials are invalid, check for spelling mistakes in your password or username, email address, or phone number.", it should only mention email address and password.
- [ ] custom email templates
- [ ] email verification - do we need it? should we enforce it?
- [x] add kratos password config (`max_breaches: 100`) to the general config.

## Ideas

- [x] Skylinks in uploads/downloads histories could have copy-to-clipboard-on-click
    - [ ] can we have some sort of notification that the link was copied?

Closes #511


ro-tex:
I think we can put a `PORTAL_NAME` or `PORTAL_DOMAIN` in the `.env` as well and use that to configure the base URLs for Kratos. We'll probably need that anyway, in order to properly set cookies that support subdomains.

Yes, it would be nice to have a move flexible solution for the config file but this setup will do for the moment.

docker-compose.yml
```diff
@@ -174,3 +174,78 @@ services:
...
+    expose:
+      - 4436
+      - 4437
+    networks:
+      shared:
+        ipv4_address: 10.10.10.85
+
+  mysqld:
+    image: mysql:5.7
+    container_name: mysqld
```
ro-tex, line 244: Note: We'll need a volume here in order to have persistence. That can come with the replication work. I'm adding this note as a reminder that until then all account data will be lost on restart.

docker/kratos/config/indentity.schema.json
```diff
@@ -0,0 +1,45 @@
+{
```
ro-tex, line 1: Quick note before I forget - there's a typo in the file name - `indentity` instead of `identity`. I'll change it later when I have the whole thing working.

docker-compose.override.yml.DISABLED
```diff
@@ -0,0 +1,92 @@
+version: "3.7"
```
ro-tex, line 1: Oops, this file shouldn't have been committed at all. It doesn't have anything sensitive in it but we don't need it in the repo.

docker/kratos/oathkeeper/id_token.jwks.json
```diff
@@ -0,0 +1,19 @@
+{
+  "keys": [
```
ro-tex, line 2: We need to override this with a secure new key.

ro-tex: Here is how to do that with Java: https://connect2id.com/products/nimbus-jose-jwt/examples/jwk-generation

docker/accounts/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM golang:1.15
+LABEL maintainer="NebulousLabs <devs@nebulous.tech>"
+
+ENV GOOS linux
+ENV GOARCH amd64
+
+WORKDIR /root
+
+RUN go get -d -u github.com/NebulousLabs/skynet-accounts/... && \
```
ro-tex, line 9: We might want to change this to fetch the latest tag or the latest release or something like that. Otherwise it would be using the latest `master` which will not be always stable.

setup-scripts/README.md
```diff
@@ -19,6 +19,7 @@ You may want to fork this repository and replace ssh keys in
   - [handshake](https://handshake.org) ([github](https://github.com/handshake-org/hsd)): full handshake node
   - [handshake-api](https://github.com/NebulousLabs/skynet-webportal/tree/master/packages/handshake-api): simple API talking to the handshake node - [read more](https://github.com/NebulousLabs/skynet-webportal/blob/master/packages/handshake-api/README.md)
   - [webapp](https://github.com/NebulousLabs/skynet-webportal/tree/master/packages/webapp): portal frontend application - [read more](https://github.com/NebulousLabs/skynet-webportal/blob/master/packages/webapp/README.md)
+  - [kratos](https://www.ory.sh/kratos/): user account management system
```
ro-tex, line 22: We should add Oathkeeper here as well. Also Cockroach and Mongo.

setup-scripts/README.md
```diff
@@ -83,10 +84,12 @@ At this point we have almost everything running, we just need to set up your wal
    - `CLOUDFLARE_AUTH_TOKEN` (optional) if using cloudflare as dns loadbalancer (need to change it in Caddyfile too)
    - `AWS_ACCESS_KEY_ID` (optional) if using route53 as a dns loadbalancer
    - `AWS_SECRET_ACCESS_KEY` (optional) if using route53 as a dns loadbalancer
+   - `MYSQL_ROOT_PASSWORD` (optional) if you're integrating accounts - it's the database password for accounts management
 1. if you have a custom domain and you configured it in `DOMAIN_NAME`, edit `/home/user/skynet-webportal/docker/caddy/Caddyfile` and uncomment `import custom.domain`
 1. only for siasky.net domain instances: edit `/home/user/skynet-webportal/docker/caddy/Caddyfile`, uncomment `import siasky.net`
 1. `docker-compose up -d` to restart the services so they pick up new env variables
 1. `docker exec caddy caddy reload --config /etc/caddy/Caddyfile` to reload Caddyfile configuration
+1. add your custom Kratos configuration to `/home/user/skynet-webportal/.kratos.yml` (in particular, the credentials for your mail server should be here, rather than in your source control).
```
ro-tex, line 92: Add "Check the `docker/kratos/config/kratos.yml` for an example of what needs to be in that file`.

Add instructions for setting up and securing mongo and cockroach.

Remove the MySQL password above.

docker/kratos/oathkeeper/id_token.jwks.json
ro-tex: Or we can just use Ory's tools:
```go
package main

import (
	"encoding/json"
	"log"
	"os"

	"github.com/ory/hydra/jwk"
)

func main() {
	gen := jwk.RS256Generator{
		KeyLength: 2048,
	}
	jwks, err := gen.Generate("", "sig")
	if err != nil {
		log.Fatal(err)
	}
	jsonbuf, err := json.MarshalIndent(jwks, "", "  ")
	if err != nil {
		log.Fatal("failed to generate JSON: %s", err)
	}
	os.Stdout.Write(jsonbuf)
}
```

docker/nginx/nginx.conf
```diff
@@ -63,4 +63,5 @@ http {
     #gzip  on;
 
     include /etc/nginx/conf.d/*.conf;
+    include /etc/nginx/conf.extra.d/*.conf;
```
ro-tex, line 66: Can we call this `conf.accounts.d` instead of `conf.extra.d`? Just so we know the relationship in the future.

docker-compose.accounts.yml
```diff
@@ -0,0 +1,163 @@
...
+    image: oryd/kratos:v0.5.4-alpha.1
+    container_name: kratos
+    restart: unless-stopped
+    logging: *default-logging
+    ports:
+      - "4433:4433" # public
+      - "4434:4434" # admin
+    expose:
+      - 4433 # public
+      - 4434 # admin
```
ro-tex, line 87: I guess we don't need both but I'm not sure which to keep. I guess we're only going to be accessing this locally, so the `expose` makes more sense, right?

docker-compose.accounts.yml
```diff
@@ -0,0 +1,163 @@
...
+      - COOKIE_HASH_KEY=${COOKIE_HASH_KEY}
+      - COOKIE_ENC_KEY=${COOKIE_ENC_KEY}
+    expose:
+      - 3000
+    networks:
+      shared:
+        ipv4_address: 10.10.10.70
+    depends_on:
+      - mongo
+      - oathkeeper
```
ro-tex, line 43: Looking at this got me thinking, we have a bunch of services that depend on each other in some way, do we need to describe that with `depends_on` instructions or are we fine with the services not functioning properly until all pieces are up?

We have this (`->` is "depends on"):
* accounts -> mongo
* accounts -\> oathkeeper (we can remove the dep)
* kratos -> kratos-migrate
* kratos-migrate -> cockroach (it can start without cockroach but it will wait and retry until that's up)
* kratos -> cockroach

I'm not sure the webapp really depends on anything - it can function all on its own, the only thing it really needs is `siad`. But the "please log in" functionality will be defunct until oathkeeper and accounts are up.

The longest chain seems to be:
webapp -> oathkeeper -> kratos -> kratos-migrate -> cockroach (slow to start because of the clustering)

Maybe the current setup is fine - all pieces start in parallel and only function after the entire stack is running.

peterjan (dismissed):

ro-tex (approved):

peterjan (approved):

PR #562, opened 2020-12-10
allow to include previous skynet stats

kwypchlo:
In a case where the server had to be reinitialised with a new siad node, you might want to add the old siad node storage stats to the current node /skynet/stats. This is handy when we move the siad node to maintenance after it reaches certain amount of files (these nodes do not expose any api) but we would want to keep the server stats nevertheless. In future https://siasky.net/skynet/stats should probably expose combined stats from all servers.

To add previous stats, create docker/data/nginx/skynet/prevstats.lua file (example below) and reload nginx config.
```lua
return { numfiles = 12345, totalsize = 123456789 }
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -72,6 +72,7 @@ services:
       - ./docker/nginx/conf.d:/etc/nginx/conf.d:ro
       - ./docker/data/nginx/cache:/data/nginx/cache
       - ./docker/data/nginx/logs:/usr/local/openresty/nginx/logs
+      - ./docker/data/nginx/skynet:/data/nginx/skynet:ro
       - ./docker/data/sia/apipassword:/data/sia/apipassword:ro
       - webapp:/var/www/webportal:ro
     networks:
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -105,6 +105,28 @@ server {
 	location /skynet/stats {
 		include /etc/nginx/conf.d/include/cors;
 
+		set $response_body ''; # we need a variable for full response body (not chuncked)
+
+		# modify the response to add numfiles and totalsize to account for node rotation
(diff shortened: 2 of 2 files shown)


ro-tex:
A couple of thoughts:
1. We will probably move a given server to maintenance several times in the long run. It would be nice for the lua script to reflect each individual maintenance copy separately. This is mostly because of point 2.
2. It would be nice to have a strategy for updating these hard-coded numbers. On one hand, some files might go unrepairable and we'll need to subtract those. On the other hand, an entire maintenance server might go bust (for whatever reason) and we should subtract its contents' stats from the total.

ro-tex (approved):
nginx magic! I don't see anything wrong with it. :)

peterjanbrone (approved):

kwypchlo:
> 1. We will probably move a given server to maintenance several times in the long run. It would be nice for the lua script to reflect each individual maintenance copy separately. This is mostly because of point 2.

Since it's lua, what I planned for is to put each individual server counts in comments and just manually sum them and return total in the returned object. I wanted to keep the returned object simple so it doesn't require too much logic in the location block.

example: 
```lua
-- m1.siasky.net(us-east:2020-03-01): 12345 / 123456789
-- m1.siasky.net(us-east:2020-05-01): 23456 / 234567890
-- m2.siasky.net(us-east:2020-06-01): 34567 / 345678901
local numfiles = 12345 + 23456 + 34567
local totalsize = 123456789 + 234567890 + 345678901
return { numfiles = numfiles, totalsize = totalsize }
```

> 2. It would be nice to have a strategy for updating these hard-coded numbers. On one hand, some files might go unrepairable and we'll need to subtract those. On the other hand, an entire maintenance server might go bust (for whatever reason) and we should subtract its contents' stats from the total.

Yeah I think we should have a way of accessing those old maintenance nodes but that's not on current roadmap yet.

PR #567, opened 2020-12-15
WIP: accounts frontend

kwypchlo:


docker/kratos/config/kratos.yml
```diff
@@ -64,7 +64,7 @@ secrets:
 
 session:
   cookie:
-    domain: siasky.xyz
+    domain: secure.siasky.xyz
```
ro-tex, line 67: We could use a comment here explaining why it should be `secure.siasky.xyz` and not `siasky.xyz`.

Hmmm... Does this mean that everyone running `accounts` will need to setup a subdomain? I guess so...

docker/nginx/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM openresty/openresty:1.19.3.1-2-bionic
+
+# RUN apt-get update -qq && apt-get install cron logrotate -qq
+RUN luarocks install luasocket
+
+# CMD ["sh", "-c", "service cron start;", "/usr/local/openresty/bin/openresty -g daemon off;"]
+CMD ["/usr/local/openresty/bin/openresty", "-g", "daemon off;"]
```
ro-tex, line 7: This part looks like WIP. Can we finalize it?

PR #570, opened 2020-12-19
Skyns

Falci:
**Proposal**: 
Add support to `/skyns/<publicKey>/<dataKey>/*`

**Why?**
* I was looking for a way to validate my `skyns` link before adding it to a HNS domain. This endpoint could facilitate a lot
* In my server (not a portal) I have a subdomain with `proxy_pass https://<key>.siasky.net;`. It would be handy if I could set the `skyns` in the URL

The majority of this code was copied from the `/hns` section. Maybe it worth a refator. 
The goal of this PR is to demonstrate the proposal, not necessarily to be merged.

### DIFF
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -24,8 +24,8 @@ limit_conn_zone $binary_remote_addr zone=downloads_by_ip:10m;
 limit_req_status 429;
 limit_conn_status 429;
 
-# since we are proxying request to nginx from caddy, access logs will contain caddy's ip address 
-# as the request address so we need to use real_ip_header module to use ip address from 
+# since we are proxying request to nginx from caddy, access logs will contain caddy's ip address
+# as the request address so we need to use real_ip_header module to use ip address from
 # X-Forwarded-For header as a real ip address of the request
 set_real_ip_from  10.0.0.0/8;
 set_real_ip_from  127.0.0.1/32;
@@ -113,11 +113,11 @@ server {
 			local file_exists = io.open("/data/nginx/skynet/prevstats.lua")
 			if file_exists then
 				file_exists.close()
-				
+
 				-- because response data is chunked, we need to concat ngx.arg[1] until
 				-- last chunk is received (when ngx.arg[2] is set to true)
 				ngx.var.response_body = ngx.var.response_body .. ngx.arg[1]
-				
+
 				if ngx.arg[2] then
 					local json = require('cjson')
 					local prevstats = require('/data/nginx/skynet/prevstats')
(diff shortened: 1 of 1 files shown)


kwypchlo:
Thank you @Falci, definitely something we could consider, I will be looking into that after holiday break 👍

Delivator:
How am I discovering this just now?
This is amazing and as far as I can tell exactly what #609 is asking for!

kwypchlo:
I admit I totally forgot about this one. I will try to review and test it out today, sorry folks!

kwypchlo:
Sorry for another delay, could not find time for that on Friday but I am going to get to this PR on a first possible occasion 🙌

Delivator:
I tested it a bit on my portal
Seemed to work in some cases but subfiles didn't work for me.
Eg. `/skyns/<publicKey>/<dataKey>/some.file`

themailman05:
Any update on this?
I am on a race against the clock to build with this feature.
Thank you.

kwypchlo:
Using url pathname for accessing the skylink directly from https://siasky.net (or any other publicly available) portal is a security risk because any uploaded app would have read and write access to the cookies set on the root domain itself. It is not much different than using a pathname as a skylink but since we discovered the issue on that, we are planning to deprecate using a pathname for accessing the skylink content in favour of using skylink in subdomain so adding this right now gets us only farther from that goal. Additionally, (the same as when using skylink in a pathname) it causes applications to be served from a non-absolute url and it breaks a lot of apps that were not really built with relative paths in mind.

One workaround could be to leverage the subdomains again for this and allow url like https://ed25519.d611373052bb124f23c3376a181e140390d9545666dae333acee2402dfc95f4a.79c05b4b67764ad99a7976a7d2fb1cfce4f196ea217ef0356af042cb5492bd5d.skyns.siasky.net in a form of https://<crypto>.<publickey>.<datakey>.skyns.siasky.net - that would make the url longer but mitigate all the issues we had with the pathname approach. 

I will talk this over today with the team on discord and will follow up here. Stay tuned!

ChrisSchinnerl:
What we can do is allow for `<subscriptionhash>.skyns.siasky.net` where `subscriptionhash` is the hash of the public key and datakey in base32 encoding. Actually both the host and renter make use of this key already to save memory.

Unfortunately fetching a registry entry this way from a portal means that we are hiding the actual public key from the portal so it is not able to verify whether the entry returned by the host is valid. The skapp would need to do this itself.

While this is not a huge task to implement, it requires a protocol upgrade. This involves a patch, rolling it out to hosts and waiting for enough hosts to upgrade to be reasonably useable. So it's doable but takes some time to roll out.

kwypchlo:
> What we can do is allow for `<subscriptionhash>.skyns.siasky.net` where `subscriptionhash` is the hash of the public key and datakey in base32 encoding. Actually both the host and renter make use of this key already to save memory.
> 
> Unfortunately fetching a registry entry this way from a portal means that we are hiding the actual public key from the portal so it is not able to verify whether the entry returned by the host is valid. The skapp would need to do this itself.
> 
> While this is not a huge task to implement, it requires a protocol upgrade. This involves a patch, rolling it out to hosts and waiting for enough hosts to upgrade to be reasonably useable. So it's doable but takes some time to roll out.

We will be following up on this over pull requests. In the meantime, it turns out that `https://<crypto>.<publickey>.<datakey>.skyns.siasky.net` is not viable since we will not be able to issue a certificate for a multi level subdomain with dynamic values like `*.*.*.skyns.siasky.net`, the best we can do is one `*` and it's not going to be enough in this case since the subdomain can only be up to 64 characters and we will not fit a full length public key and data key together.

Delivator:
To be fair, I think the biggest usecase for this endpoint isn't to access a Skapp, but rather to have a static URL that can be updated.
@themailman05's and my usecase for example would be, to use Skynet to host the JSON metadata of Ethereum tokens.
https://eips.ethereum.org/EIPS/eip-1155#metadata
That would allow us to update the metadata off-chain and gasless. No need for a subdomain in that case.

themailman05:
This seems like a fine solution, though as @Delivator mentions, I don't really need subdomain, just a unique URL that points to a mutable SkyDB record.

Happy to provide more context if necessary.

Falci:
Hi 👋
Any update here?

kwypchlo:
> Hi 👋
> Any update here?

Hey, we are finishing work on skylink v2 support which will basically let you encode a registry record into a skylink. I will make sure to leave a comment here in addition to the official announcement when it's out, shouldn't be too long any more.

Falci:
Thanks for the quick reply.
Would that cover the use case I commented in the **why**?
> In my server (not a portal) I have a subdomain with proxy_pass https://<key>.siasky.net;. It would be handy if I could set the skyns in the URL

kwypchlo:
> Thanks for the quick reply.
> Would that cover the use case I commented in the **why**?
> 
> > In my server (not a portal) I have a subdomain with proxy_pass https://.siasky.net;. It would be handy if I could set the skyns in the URL

yes it would, instead of setting publicKey and dataKey in hns entry, you will be able to set it to the v2 skylink (46 bytes) that would serve as a pointer to that registry entry

dghelm:
> Thanks for the quick reply.
> Would that cover the use case I commented in the **why**?
> 
> > In my server (not a portal) I have a subdomain with proxy_pass https://.siasky.net;. It would be handy if I could set the skyns in the URL

Hello Falci, we have most of Skylink v2 implemented, with a few loose ends before an official announcement. If you wanted to reach out on discord (dghelm#8125) or email (daniel@siasky.net), I'd love to touch base and make sure your use case has been addressed.

kwypchlo:
closing this with resolver skylinks implementing the same functionality

read more: https://docs.siasky.net/skynet-topics/resolver-skylinks

PR #578, opened 2021-01-05
Extend `nginx` log format

peterjanbrone:
This extends the log format to include the `hns` subdomain.


peterjanbrone:
Put it in `DRAFT` as I still need to check with daniel + verify it on siasky.xyz

PR #581, opened 2021-01-08
upgrade openresty image to ubuntu 18

kwypchlo:
- use ubuntu 18 openresty image instead of ubuntu 16 image that has not been maintained for quite a while now
- use Dockerfile to enable future extensions to openresty - https://github.com/NebulousLabs/skynet-webportal/pull/554 will require luasocket and we will need to fix logs rotation at some point too

### DIFF
--- a/docker/nginx/Dockerfile
+++ b/docker/nginx/Dockerfile
@@ -1,0 +1,7 @@
+FROM openresty/openresty:1.19.3.1-2-bionic
+
+# RUN apt-get update -qq && apt-get install cron logrotate -qq
+# RUN luarocks install luasocket
+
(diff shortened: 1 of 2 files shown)


docker/nginx/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM openresty/openresty:1.19.3.1-2-bionic
+
+# RUN apt-get update -qq && apt-get install cron logrotate -qq
```
peterjanbrone, line 3: are you purposefully committing this?

peterjanbrone (approved):

ro-tex (approved):

docker/nginx/Dockerfile
ro-tex: from offline: "it's for future reference - that's to support the todos I described in pr description"

PR #587, opened 2021-01-19
Cockroach backups

ro-tex:
We use S3 for simplicity. We can switch to Filebase later on, so we can stay on the Sia network.

It's OK for all nodes to try to backup every day. Only one node will succeed and all others will fail with "you need an enterprise license for incremental backups" which is more than fine for us. All we need to do it schedule the backups at different times (random hour?).

I reuse the existing AWS credentials from .env, I just added S3 access rights to that robot user.

The backup script has been added to the crontab of all servers with one hour difference between executions, so we don't get some odd race conditions situation.

**Note:**
In order for the restoration of the database to succeed, the DB shouldn't exist. Should I add that to the script? I'm kinda hesitant to drop the existing DB only to find that the backup name the user supplied is invalid and now we don't have a DB?

Also, the restore script relies on the user to supply a valid backup name (which can very easily be guessed, so you don't even need to check) instead of looking for the most recent one because the latter approach turned out to be annoying to script (I need to dig into `jq`). I can go around that by just trying the current day's backup and then backtrack one day until I find an existing backup for go before Jan 2021.


scripts/crdb_backup.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Take the current datetime:
+DT=`date +%Y-%m-%d`
+# Set the environment:
+set -o allexport
+source ../.env
+set +o allexport
+# Take the current datetime:
```
peterjanbrone, line 9: this appears twice

scripts/crdb_backup.sh
```diff
@@ -0,0 +1,16 @@
...
+DT=`date +%Y-%m-%d`
```
peterjanbrone, line 10: why only add the day, it's probably better to include time as well (`.%H.%M.%S`)

scripts/crdb_restore.sh
```diff
@@ -0,0 +1,18 @@
...
+
+BACKUP=$1
+if [[ $BACKUP == "" ]]; then
+  echo "No backup name given. It should look like '2020-01-29'."
+  exit 1
+fi
+
+# Set the environment:
+set -o allexport
+source ../.env
```
kwypchlo, line 11: Will that relative path work when running this from crontab?

You could use this
```bash
# get current working directory (pwd doesn't cut it)
cwd=$(cd -P -- "$(dirname -- "$0")" && pwd -P)
```

https://github.com/NebulousLabs/skynet-webportal/blob/master/scripts/portal-upgrade.sh#L6

scripts/crdb_restore.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+BACKUP=$1
+if [[ $BACKUP == "" ]]; then
+  echo "No backup name given. It should look like '2020-01-29'."
+  exit 1
+fi
+
+# Set the environment:
```
peterjanbrone, line 9: I am assuming you are setting the environment to ensure the access key id and stuff is set, perhaps it makes sense to explicitly check for those and exit with an informative user message.

peterjanbrone (approved):

scripts/crdb_restore.sh
ro-tex: Or maybe I can hard-code the path to the .env file? What would be better?

scripts/crdb_backup.sh
ro-tex: We want one backup per day for the moment and all servers will try to backup at different times. The first one will succeed, the rest will fail with "get an enterprise license".

scripts/crdb_backup.sh
ro-tex: Ha ha, the joys of copy-paste I guess.

scripts/crdb_restore.sh
ro-tex: Sure, I can add that.

scripts/crdb_restore.sh
ro-tex: Done, used your approach. It's less likely that we'll move the script (within the project) than that we'll deploy somewhere else.

scripts/crdb_backup.sh
peterjanbrone: hm interesting, aren't we worried all nodes will fail and we won't notice?

peterjanbrone: I'm only worried because a failed backup becomes part of the expected behaviour,
that's when mistakes or oversights happen 🤷‍♂️

PR #600, opened 2021-01-26
Healthcheck cron

kwypchlo:
replace js cron library with native cron for improved performance and garbage collection - we were having issues with this container getting heavy on the load, running one-off commands with native cron should resolve the issue


packages/health-check/src/checks/verbose.js
```diff
@@ -1037,7 +1037,7 @@ function skylinkVerification(done, { name, skylink, bodyHash, headers }) {
   const time = process.hrtime();
 
   // Create the query for the skylink
-  const query = `http://${process.env.PORTAL_URL}/${skylink}?nocache=true`;
+  const query = `${process.env.PORTAL_URL}/${skylink}?nocache=true`;
```
peterjanbrone, line 1040: (nit) strip trailing slash

peterjanbrone (approved):

ChrisSchinnerl (approved):

PR #604, opened 2021-01-27
fix hns links redirects

kwypchlo:
probably due to upgrade in nginx we introduced an issue where handshake domains accessed with https://siasky.net/hns/domain_name/ would not register the trailing slash and would redirect to https://siasky.net/hns/domain_name/domain_name/ and error out

the weird thing is that it sometimes work and sometimes it doesn't - probably due to the variable being set somewhere globally, that's why we need to explicitly declare it as local

broken commit https://github.com/NebulousLabs/skynet-webportal/commit/c946014e7c1d97f93667e3c87363204348ccca64

### DIFF
--- a/packages/health-check/src/checks/verbose.js
+++ b/packages/health-check/src/checks/verbose.js
@@ -1024,6 +1024,20 @@ function fileEndpointCheck(done) {
   skylinkVerification(done, linkInfo);
 }
 
+// check whether hns/note-to-self would properly redirect to note-to-self/
+function hnsEndpointDirectoryRedirect(done) {
+  const expected = {
+    name: "hns endpoint directory redirect",
+    skylink: "hns/note-to-self",
+    statusCode: 307,
+    headers: {
+      location: "note-to-self/",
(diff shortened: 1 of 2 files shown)


ChrisSchinnerl (approved):

packages/health-check/src/checks/verbose.js
```diff
@@ -1033,26 +1047,36 @@ function parseHeaderString(header) {
...
   const time = process.hrtime();
 
   // Create the query for the skylink
-  const query = `${process.env.PORTAL_URL}/${skylink}?nocache=true`;
+  const query = `${process.env.PORTAL_URL}/${skylink}`;
 
   // Get the Skylink
   superagent
     .get(query)
+    .set("cookie", "nocache=true")
```
kwypchlo, line 1059: move `?nocache=true` from url to cookie because it would be harder to assert proper `location` header response with `?nocache=true` in it

peterjanbrone (approved):

PR #624, opened 2021-02-18
Include Skynet-Portal-Api header

### LINKED ISSUE #423: Configure the portal domain
## Problem

`sky-bin.hns.siasky.net` wasn't working because the generated skylinks were appended to the full domain path, e.g. `sky-bin.hns.siasky.net/<skylink>`. We put in a [temporary fix](https://github.com/NebulousLabs/skybin/pull/3) that extracts the last two domain elements, e.g. `siasky.net` in this example, and uses that, but it fails in some edge cases:

- If there is a portal on skynet.somedomain.com for example where somedomain.com is not a portal this won't work
- Also on www.siacdn.com this will just generate siacdn.com/skylink instead of www.siacdn.com/skylink
- And on skyportal.uk.com I think this will return uk.com/skylink

## Proposed solution

Eventually we should include the portal domain in either a header or endpoint (I think I prefer the first) and update sky-bin to use that. siasky.net will need to be configured to have this header/endpoint return the portal domain, i.e. `siasky.net`, so that applications living at `<domain>.hns.siasky.net` and `siasky.net/hns/<domain>` can access the portal domain.

kwypchlo:
Header `skynet-portal-api` will be included on every portal response whether it's a GET, POST or HEAD.
This allows anyone to basically to write
```js
const skynetPortalApi = (await fetch("", { method: "HEAD" })).headers.get("Skynet-Portal-Api")
```
and through cheap, small and fast HEAD request acquire Skynet-Portal-Api value when calling from a portal hosted web app since the url we query `""` which is an empty relative url, will just use the current address. It is possible to instead of `""` use `"/"` but it doesn't really matter much.

In terms of naming, it's called exactly `Skynet-Portal-Api` in the nginx because the headers that start with `Skynet-` are the ones that siad returns and they are all Pascal-Kebab-Case (first letter capitalised, words separated by dash). In most cases, letter casing doesn't matter because clients are either case insensitive or transform the header into lower-kebab-case anyway - that js example with `fetch` above, you could use `"Skynet-Portal-Api"` and `"skynet-portal-api"` and both would be correctly returned.

**migration note:** portal operators will be required to add a new environment variable into the `.env` file called `SKYNET_PORTAL_API` which is the absolute url to the portal api (for us it is https://siasky.net), otherwise the header will not be included in the responses.

**verification:** run the example snippet on:
- https://siasky.dev
- https://siasky.dev/hns/marstorage/
- https://marstorage.hns.siasky.dev

closes #423

### DIFF
--- a/docker/nginx/nginx.conf
+++ b/docker/nginx/nginx.conf
@@ -25,6 +25,7 @@ worker_processes  1;
 
 #pid        logs/nginx.pid;
 
+env SKYNET_PORTAL_API; # declare env variable to use it in config
 
 events {
     worker_connections  1024;
@@ -61,6 +62,9 @@ http {
     keepalive_timeout  65;
 
     #gzip  on;
+
+    # include skynet-portal-api header on every request
+    header_filter_by_lua 'ngx.header["Skynet-Portal-Api"] = os.getenv("SKYNET_PORTAL_API")';
 
     include /etc/nginx/conf.d/*.conf;
 }
--- a/docker/nginx/conf.d/include/cors
+++ b/docker/nginx/conf.d/include/cors
@@ -11,4 +11,4 @@ if ($request_method = 'OPTIONS') {
 more_set_headers 'Access-Control-Allow-Origin: *';
 more_set_headers 'Access-Control-Allow-Methods: GET, POST, OPTIONS, PUT, DELETE';
 more_set_headers 'Access-Control-Allow-Headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range';
-more_set_headers 'Access-Control-Expose-Headers: Content-Length,Content-Range,Skynet-File-Metadata,Skynet-Skylink';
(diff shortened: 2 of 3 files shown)


docker/nginx/nginx.conf
```diff
@@ -62,5 +63,8 @@ http {
 
     #gzip  on;
 
+    # include skynet-portal-api header on every request
+    header_filter_by_lua 'ngx.header["Skynet-Portal-Api"] = os.getenv("SKYNET_PORTAL_API")';
```
peterjanbrone, line 67: what's the reason you're doing this with `header_filter_by_lua` - wouldn't you able to accomplish the same thing simply by doing `add_header  Skynet-Portal-Api  {SKYNET_PORTAL_API};`

peterjanbrone: Ok so I guess Nginx can't handle env variables like that. Lua it is then. In general though we should refrain from using `lua` if we don't absolutely have to.

peterjanbrone (approved):

m-cat (approved):

PR #635, opened 2021-03-04
only deploy on webapp changes

kwypchlo:
we should only use the deploy to skynet action if the webapp project is changed, otherwise it's a spam and unnecessary traffic


ro-tex (approved):

peterjanbrone:
LGTM

peterjanbrone (approved):

PR #637, opened 2021-03-05
blocklist batching

kwypchlo:
due to issue with the list of arguments getting too long, we are batching the skylinks

### DIFF
--- a/setup-scripts/blocklist-airtable.py
+++ b/setup-scripts/blocklist-airtable.py
@@ -75,13 +75,14 @@ async def block_skylinks_from_airtable():
         return print(message) or await send_msg(client, message, force_notify=False)
 
     print("Searching nginx cache for blocked files")
-    
-    print("Searching nginx cache for blocked files")
-    exec('printf "^KEY: .*' + '\n^KEY: .*'.join(skylinks) + '" > /tmp/blocklist-aritable.txt')
-    cached_files_command = (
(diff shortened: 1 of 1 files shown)


setup-scripts/blocklist-airtable.py
```diff
@@ -75,12 +75,14 @@ async def block_skylinks_from_airtable():
...
-        "/usr/bin/find /data/nginx/cache/ -type f | /usr/bin/xargs --no-run-if-empty -n1000 /bin/grep -Els '^KEY: .*("
-        + "|".join(skylinks)
-        + ")'"
-    )
-    cached_files_count = int(exec('docker exec -it nginx bash -c "' + cached_files_command + ' | wc -l"') or 0)
+    cached_files_count = 0
+    for i in range(0, len(skylinks), 1000):
+        cached_files_command = (
+            "/usr/bin/find /data/nginx/cache/ -type f | /usr/bin/xargs --no-run-if-empty -n1000 /bin/grep -Els '^KEY: .*("
+            + "|".join(skylinks[i:i+1000])
```
MSevey, line 82: Will python panic if you exceed the length of the array?
If so we should check that i + 1000 doesn't exceed the length. Otherwise LGTM

kwypchlo: nope, it will just return as many records as there are

ro-tex (approved):

peterjanbrone (approved):

PR #697, opened 2021-04-27
Update the hardcoded list of servers for blocking skylinks.

ro-tex:


kwypchlo (approved):

scripts/blocklist-skylink.sh
```diff
@@ -32,7 +32,13 @@ fi
 #########################################################################
 # iterate through all servers, block the skylinks and purge it from cache
 #########################################################################
-for server in "eu-ger-1.siasky.net" "eu-ger-2.siasky.net" "eu-fin-1.siasky.net" "us-or-1.siasky.net" "us-or-2.siasky.net" "us-va-1.siasky.net" "us-pa-1.siasky.net" "us-pa-2.siasky.net" "siasky.xyz";
+declare -a servers=(  "eu-ger-1.siasky.net" "eu-ger-2.siasky.net" "eu-ger-3.siasky.net" "eu-ger-4.siasky.net"
```
m-cat, line 35: Doesn't this work?
```suggestion
servers=(  "eu-ger-1.siasky.net" "eu-ger-2.siasky.net" "eu-ger-3.siasky.net" "eu-ger-4.siasky.net"
```

m-cat (approved):
Looks good, just a wee nit.

scripts/blocklist-skylink.sh
kwypchlo: Yeah it does, is there any specific reason why you wanted to use `declare -a` there? Seems redundant

ro-tex: No special reason. I googled "how to define an array in bash" and this is what I got.
Will clean up in a F/U (https://github.com/SkynetLabs/skynet-webportal/issues/702).

PR #699, opened 2021-04-27
improve health checks

kwypchlo:
- replaced superagent with got
- removed dependency on storing the file locally in a temp file
- fixed sha1 checksums - used `shasum` tool locally to verify (those existing hashes were not correct, verified with saving the file and hashing manually)
- fixed extracting error messages from failed responses
- renamed "verbose" suite into "extended"


packages/health-check/src/checks/verbose.js
```diff
@@ -1098,69 +1099,61 @@ function parseHeaderString(header) {
...
-        if (bodyHash) {
-          const currentBodyHash = hash(response.body);
-          if (currentBodyHash !== bodyHash) {
-            entry.up = false;
-            info.bodyHash = { expected: bodyHash, current: currentBodyHash };
-          }
-        }
+  try {
+    const response = await got[method](query, { followRedirect, headers: { cookie: "nocache=true" } });
+    // const [response, buffer] = await Promise.all([promise, promise.buffer()]);
```
MSevey, line 1110: Can this be deleted?

MSevey (approved):
Is there any way to add CI tests for the health checks? Would be nice to know that changes are going to break the health checks before we merge and deploy them.

packages/health-check/src/checks/verbose.js
kwypchlo: yes

kwypchlo:
```suggestion
```

kwypchlo:
> Is there any way to add CI tests for the health checks? Would be nice to know that changes are going to break the health checks before we merge and deploy them.

that's a good idea, maybe once we get some more time to add more end-to-end tests

packages/health-check/src/checks/verbose.js
```diff
@@ -1098,69 +1099,60 @@ function parseHeaderString(header) {
...
-        // Check if the response body is valid by checking against the known hash
-        if (bodyHash) {
-          const currentBodyHash = hash(response.body);
-          if (currentBodyHash !== bodyHash) {
-            entry.up = false;
-            info.bodyHash = { expected: bodyHash, current: currentBodyHash };
-          }
-        }
+  try {
+    const response = await got[method](query, { followRedirect, headers: { cookie: "nocache=true" } });
```
peterjan, line 1109: what's this `nocache=true` cookie?

peterjan (approved):

packages/health-check/src/checks/verbose.js
```diff
@@ -1086,7 +1087,7 @@ function hnsEndpointDirectoryRedirect(done) {
     },
   };
 
-  skylinkVerification(done, expected, { redirects: 0 });
+  skylinkVerification(done, expected, { followRedirect: false });
```
ro-tex, line 1090: This function is now `async`. Shouldn't we `await` here (and everywhere else)?

packages/health-check/src/checks/verbose.js
kwypchlo: we can either use a cookie or `?nocache=true` in url to omit nginx cache and ensure that the record is served directly from siad - we need that becuase otherwise we're sometimes hitting cache and not testing end-to-end

packages/health-check/src/checks/verbose.js
kwypchlo: yeah we might but not in this iteration yet, we would have to refactor all 40ish test functions

ro-tex (approved):

packages/health-check/src/checks/critical.js
```diff
@@ -36,11 +37,12 @@ async function downloadCheck(done) {
...
 
   try {
-    const response = await superagent.get(`${process.env.PORTAL_URL}/${skylink}?nocache=true`);
+    const response = await got(`${process.env.PORTAL_URL}/${skylink}?nocache=true`);
 
     statusCode = response.statusCode;
   } catch (error) {
-    statusCode = error.statusCode || error.status;
+    statusCode = error?.response?.statusCode || error.statusCode || error.status;
+    errorMessage = error.message;
```
peterjan, line 45: not used it seems

peterjan (approved):
LGTM 👍

packages/health-check/src/checks/critical.js
kwypchlo: resolved

PR #703, opened 2021-04-30
Add skyd tracing with Jaeger

ro-tex:
## Overview

Add tracing for `skyd`.

## Details

We need to be able to quickly answer questions like "Why is this call slow?" in real time. In order to be able to do that we need to know exactly what is happening with user requests going through the stack and specifically what is happening within `skyd`.

This PR introduces Jaeger - a tracing tool to which `skyd` will send information about its internal state. Jaeger will persist this data on disk with the help of Elasticsearch and will expose a nice UI at local port 16686 (local because there is no authentication).

The integration with Jaeger requires additional instrumentation in `skyd` which is available on branch `chris/tracing-sample` and will be extended and merged into `master` in the future.

The integration with Jaeger is disabled in `skyd` by default and will only be enabled if the env var `JAEGER_DISABLED` is set to `false`.

## Caveats

Elasticsearch seems to be running as a regular user in its container and therefore cannot write to regular volumes unless they are set up with specific ownership or permissions. That's why for it to work we need to specifically create and `chown` its local data directory at `./docker/data/elasticsearch/data` to `1000:1000`.

### DIFF
--- a/docker-compose.jaeger.yml
+++ b/docker-compose.jaeger.yml
@@ -21,10 +21,10 @@ services:
     command: [ "--reporter.grpc.host-port=jaeger-collector:14250", "--reporter.grpc.retry.max=1000" ]
     container_name: jaeger-agent
     restart: on-failure
-    ports:
-      - "127.0.0.1:6831:6831/udp"
-      - "127.0.0.1:6832:6832/udp"
-      - "127.0.0.1:5778:5778"
+    expose:
+      - 6831
+      - 6832
+      - 5778
     environment:
       - LOG_LEVEL=debug
     networks:
@@ -38,10 +38,10 @@ services:
     command: [ "--es.num-shards=1", "--es.num-replicas=0", "--es.server-urls=http://elasticsearch:9200" ]
     container_name: jaeger-collector
     restart: on-failure
-    ports:
-      - "127.0.0.1:14269:14269"
-      - "127.0.0.1:14268:14268"
-      - "127.0.0.1:14250:14250"
+    expose:
+      - 14269
+      - 14268
+      - 14250
     environment:
       - SPAN_STORAGE_TYPE=elasticsearch
       - LOG_LEVEL=debug
@@ -56,9 +56,9 @@ services:
     command: [ "--es.num-shards=1", "--es.num-replicas=0", "--es.server-urls=http://elasticsearch:9200" ]
     container_name: jaeger-query
     restart: on-failure
-    ports:
-      - "127.0.0.1:16686:16686"
-      - "127.0.0.1:16687:16687"
+    expose:
+      - 16686
+      - 16687
     environment:
       - SPAN_STORAGE_TYPE=elasticsearch
       - LOG_LEVEL=debug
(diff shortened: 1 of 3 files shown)


kwypchlo (commented):
there's a lot of ports being mapped to parent network - are all those necessary?

docker-compose.jaeger.yml
```diff
@@ -0,0 +1,88 @@
...
+  jaeger-agent:
+    image: jaegertracing/jaeger-agent
+    command: [ "--reporter.grpc.host-port=jaeger-collector:14250", "--reporter.grpc.retry.max=1000" ]
+    container_name: jaeger-agent
+    restart: on-failure
+    ports:
+      - "5775:5775/udp"
+      - "6831:6831/udp"
+      - "6832:6832/udp"
+      - "5778:5778"
```
ChrisSchinnerl, line 29: Shouldn't those be `localhost:5775:5775` etc. to avoid docker adding them to iptables?

Maybe even better use 
```
expose:
  - someport
```
To not export the port at all and only make it visible from within the shared network of the containers?
https://github.com/compose-spec/compose-spec/blob/master/spec.md#expose

docker-compose.jaeger.yml
```diff
@@ -0,0 +1,88 @@
+version: '3.7'
+
+services:
+  sia:
+    environment:
+      - SIA_MODULES=gctwr
```
ChrisSchinnerl, line 6: why is this in the jaeger.yml?

docker-compose.jaeger.yml
ChrisSchinnerl: Same goes for all the other exposed ports in this file

docker-compose.jaeger.yml
ro-tex: Forgotten. I'll clean it up.

ChrisSchinnerl (dismissed):

docker-compose.jaeger.yml
```diff
@@ -0,0 +1,85 @@
...
+    environment:
+      - discovery.type=single-node
+    volumes:
+      # This dir needs to be chowned to 1000:1000
+      - ./docker/data/elasticsearch/data:/usr/share/elasticsearch/data
+    expose:
+      - 9200
+    networks:
+      shared:
+        ipv4_address: 10.10.10.93
```
ChrisSchinnerl, line 85: not blocking but I think you don't need all of those now that you reference them by hostname

ro-tex: I'd rather have them in there, even if they don't do anything. We set IPs for all other containers.

kwypchlo: Yeah we need to set them manually because if one of the services restart and they would have a different address assigned, in case one service would have cached the ip with the hostname, it would render the connection unreachable. We had that issue with nginx because nginx caches hostnames and ips and needs a full restart in case those change - it's a precaution basically.

kwypchlo (approved):

ChrisSchinnerl (approved):

PR #715, opened 2021-05-04
cache /accounts internal response for 1 minute

kwypchlo:
not perfect but will help us scale up for now - we will have people under unauthorised user limits for up to a minute after they log in


peterjan (approved):
LGTM (`match` var name threw me off for 5s though)

ro-tex (approved):

PR #717, opened 2021-05-05
accounts health checks and testing improvements

kwypchlo:
- added accounts health check that polls https://account.siasky.net/health endpoint and checks whether `dbAlive` is `true`
- replaced `PORTAL_URL` that defaulted to `nginx` with an alias in /etc/hosts for the current machine ip to siasky.net and account.siasky.net so we can actually test requests to https://siasky.net and https://account.siasky.net instead of requesting from http://nginx
- added `ip` entry in health checks so we could double check whether the health check actually hit the specific server that we wanted

### DIFF
--- a/packages/health-check/src/checks/critical.js
+++ b/packages/health-check/src/checks/critical.js
@@ -19,7 +19,7 @@ async function uploadCheck(done) {
     data.up = true;
     data.ip = response.ip;
   } catch (error) {
-    data.statusCode = error?.response?.statusCode || error.statusCode || error.status;
+    data.statusCode = error.response?.statusCode || error.statusCode || error.status;
     data.errorMessage = error.message;
     data.errorResponseContent = getResponseContent(error.response);
     data.ip = error?.response?.ip ?? null;
--- a/packages/health-check/src/checks/critical.js
+++ b/packages/health-check/src/checks/critical.js
@@ -7,26 +7,28 @@ const { calculateElapsedTime, getResponseContent } = require("../utils");
 async function uploadCheck(done) {
   const time = process.hrtime();
   const form = new FormData();
-  const data = Buffer.from(new Date()); // current date to ensure data uniqueness
+  const payload = Buffer.from(new Date()); // current date to ensure data uniqueness
+  const data = { up: false };
 
-  form.append("file", data, { filename: "time.txt", contentType: "text/plain" });
-  let statusCode, errorResponseContent;
+  form.append("file", payload, { filename: "time.txt", contentType: "text/plain" });
 
   try {
-    const response = await got.post(`${process.env.PORTAL_URL}/skynet/skyfile`, { body: form });
+    const response = await got.post(`${process.env.SKYNET_PORTAL_API}/skynet/skyfile`, { body: form });
 
-    statusCode = response.statusCode;
+    data.statusCode = response.statusCode;
+    data.up = true;
+    data.ip = response.ip;
   } catch (error) {
-    statusCode = error?.response?.statusCode || error.statusCode || error.status;
-    errorResponseContent = getResponseContent(error?.response);
+    data.statusCode = error?.response?.statusCode || error.statusCode || error.status;
+    data.errorMessage = error.message;
+    data.errorResponseContent = getResponseContent(error.response);
+    data.ip = error?.response?.ip ?? null;
   }
 
   done({
     name: "upload_file",
-    up: statusCode === StatusCodes.OK,
-    statusCode,
-    errorResponseContent,
     time: calculateElapsedTime(time),
+    ...data,
   });
 }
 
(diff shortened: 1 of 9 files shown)


peterjan (dismissed):
LGTM 👏🏻

packages/health-check/src/checks/critical.js
```diff
@@ -7,53 +7,86 @@ const { calculateElapsedTime, getResponseContent } = require("../utils");
...
+  const payload = Buffer.from(new Date()); // current date to ensure data uniqueness
+  const data = { up: false };
 
-  form.append("file", data, { filename: "time.txt", contentType: "text/plain" });
-  let statusCode, errorResponseContent;
+  form.append("file", payload, { filename: "time.txt", contentType: "text/plain" });
 
   try {
-    const response = await got.post(`${process.env.PORTAL_URL}/skynet/skyfile`, { body: form });
+    const response = await got.post(`${process.env.SKYNET_PORTAL_API}/skynet/skyfile`, { body: form });
```
ro-tex, line 16: This seems to depend on the custom `/etc/hosts` that this PR is adding. Is this better than using `PORTAL_NAME`? It seems like an easy thing to mess up in the future if we decide to, for example, have the portal API be at `portal.siasky.net` and don't add that to the custom `/etc/hosts`. And then we'd end up with health checks going through the load balancer and being meaningless.

I don't mind the way it works right now but I'm afraid that this change is dangerous because it's setting us up for failure if we forget to do a seemingly unrelated change. The definition of tech debt.

packages/health-check/src/checks/critical.js
```diff
@@ -7,53 +7,86 @@ const { calculateElapsedTime, getResponseContent } = require("../utils");
...
 
-    statusCode = response.statusCode;
+    data.statusCode = response.statusCode;
+    data.up = true;
+    data.ip = response.ip;
   } catch (error) {
-    statusCode = error?.response?.statusCode || error.statusCode || error.status;
-    errorResponseContent = getResponseContent(error?.response);
+    data.statusCode = error?.response?.statusCode || error.statusCode || error.status;
```
ro-tex, line 22: nit: `error` can't be null here :)

packages/health-check/src/checks/critical.js
```diff
@@ -7,53 +7,86 @@ const { calculateElapsedTime, getResponseContent } = require("../utils");
...
   });
 }
 
-module.exports = [uploadCheck, downloadCheck];
+async function accountHealthCheck(done) {
+  const time = process.hrtime();
+  const data = { up: false };
+
+  try {
+    const response = await got(`${process.env.SKYNET_DASHBOARD_URL}/health`, { responseType: "json" });
```
ro-tex, line 66: Same comment as above. Maybe we can use the value of the env var in and bind that to the current IP? That would remove the possibility of desyncing the two. But while we're using one hardcoded value (Dockerfile) and another configurable one we will be running that risk.

ro-tex (changes requested):

packages/health-check/src/checks/critical.js
kwypchlo:
```suggestion
    data.statusCode = error.response?.statusCode || error.statusCode || error.status;
```

packages/health-check/src/checks/critical.js
kwypchlo: To ensure that we're not going through load balancer we have now the `ip` property that we can double check in case we're not sure - of course it's manual but that's additional layer of security.

PORTAL_NAME is not a good choice here because it can be anything, readme says:
> `PORTAL_NAME` a string representing name of your portal e.g. `siasky.xyz` or `my skynet portal`

I am aware of the potential issue with hardcoding account.siasky.net but there wasn't really easy way around it that wouldn't require adding a new env variable to every server. I would like to follow up at some point with a cleanup to .env file and all environment variables, this could also mean we would add specific env variable that we could use here.

packages/health-check/src/checks/critical.js
kwypchlo: `SKYNET_DASHBOARD_URL` contains `https://` prefix and it wouldn't work if we assign the ip to address with https:// prefix - tried it. We would need `SKYNET_DASHBOARD_URL_NO_HTTPS` :) Let's follow up in a env cleanup branch once there is a bit more time.

ro-tex: 1. I agree that `.env` is due for a cleanup. I would advise to do that when we stabilise the stack a little more.
2. We could trim the prefix. That should solve the issue for both `SKYNET_DASHBOARD_URL` and `SKYNET_PORTAL_API`. I would prefer to resolve this now rather than later but I won't hold the PR on it and I'll trust your judgement.

packages/health-check/src/checks/critical.js
ro-tex: Right, I forgot that was a very open var.

ro-tex (approved):

peterjan (approved):

PR #757, opened 2021-05-11
Wait for Elasticsearch to be ready

ro-tex:
Jaeger doesn't have the option to wait for Elasticsearch to be ready before starting. It also doesn't retry connecting if it has started without a valid connection to Elasticsearch. This makes it completely unreliable and forces us to look for a way to ensure Elasticsearch is up and running (see the note) before Jaeger starts.

This PR does exactly that by running a shell script that doesn't return until it either manages to successfully talk to Elasticsearch or exhausts its allotted time and exist with an error.

Note: Using `docker-compose`'s built-in dependencies does not guarantee that Elasticearch will be ready by the time Jaeger starts. All it guarantees is that the container in which ES is running is fully started before Jaeger's container is started. That doesn't tell us anything about the readiness of the service.


ChrisSchinnerl (approved):
lgtm

kwypchlo (changes requested):

kwypchlo (approved):

PR #819, opened 2021-05-28
use dc command instead of docker-compose directly

kwypchlo:
Instead of using `docker-compose` and adding compose files manually, we can now use `./dc` command that already loads compose files based on `PORTAL_MODULES` variable in .env file. By default it loads only docker-compose.yml and docker-compose.override.yml if exists but you can set also "a" for accounts and "j" for jaeger. Example `PORTAL_MODULES=aj`

### DIFF
--- a/dc
+++ b/dc
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+if [ -f .env ]; then
+  export $(cat .env | sed 's/#.*//g' | xargs)
+fi
+
+for i in $(seq 1 ${#PORTAL_MODULES}); do
+  # accounts module - alias "a"
+  if [ ${PORTAL_MODULES:i-1:1} == "a" ]; then
+    COMPOSE_FILES+=" -f docker-compose.accounts.yml"
+  fi
+
+  # jaeger module - alias "j"
+  if [ ${PORTAL_MODULES:i-1:1} == "j" ]; then
+    COMPOSE_FILES+=" -f docker-compose.jaeger.yml"
+  fi
+done
+
+# override file if exists
+if [ -f docker-compose.override.yml ]; then
+  COMPOSE_FILES+=" -f docker-compose.override.yml"
+fi
+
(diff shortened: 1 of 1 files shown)


ro-tex (approved):

dc
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
```
firyx, line 2: Aren't we missing default `docker-compose.yml` in `COMPOSE_FILES`?

dc
```diff
@@ -0,0 +1,24 @@
...
+  # accounts module - alias "a"
+  if [ ${PORTAL_MODULES:i-1:1} == "a" ]; then
+    COMPOSE_FILES+=" -f docker-compose.accounts.yml"
+  fi
+
+  # jaeger module - alias "j"
+  if [ ${PORTAL_MODULES:i-1:1} == "j" ]; then
+    COMPOSE_FILES+=" -f docker-compose.jaeger.yml"
+  fi
+done
```
firyx, line 17: Couldn't be simplified to something like?

`echo "${#PORTAL_MODULES}" | grep a && COMPOSE_FILES+=" -f docker-compose.accounts.yml"`
`echo "${#PORTAL_MODULES}" | grep j && COMPOSE_FILES+=" -f docker-compose.jaeger.yml"`

dc
kwypchlo: oh yeah we do :)

dc
kwypchlo: I wanted to make sure we include modules in an order that they are defined in the PORTAL_MODULES that's why I need a loop (sometimes one file would override some settings from the other and order is important)

firyx: we should align this in ansible

firyx (approved):

ro-tex (approved):
LGTM.

I would prefer the `grep` approach Fil proposed but this works as well.

PR #820, opened 2021-05-31
env variables are strings, fix conditions

kwypchlo:
all env variables are strings and we need to compare them accordingly - fixing issue where setting `ACCOUNTS_ENABLED=0` did not result in accounts being disabled

### DIFF
--- a/packages/health-check/src/checks/critical.js
+++ b/packages/health-check/src/checks/critical.js
@@ -108,7 +108,7 @@ async function genericAccessCheck(name, url, done) {
 
 const checks = [uploadCheck, websiteCheck, downloadCheck, skylinkSubdomainCheck, handshakeSubdomainCheck];
 
-if (process.env.ACCOUNTS_ENABLED) {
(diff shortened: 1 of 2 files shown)


ro-tex (approved):

packages/health-check/src/checks/critical.js
```diff
@@ -108,7 +108,7 @@ async function genericAccessCheck(name, url, done) {
 
 const checks = [uploadCheck, websiteCheck, downloadCheck, skylinkSubdomainCheck, handshakeSubdomainCheck];
 
-if (process.env.ACCOUNTS_ENABLED) {
+if (process.env.ACCOUNTS_ENABLED === "1") {
```
peterjan, line 111: I personally don't like using 1 and 0 for 'true' and 'false'. In an `.env` file I guess you'd usually write:
`process.env.ACCOUNTS_ENABLED=true` rather than `process.env.ACCOUNTS_ENABLED=1`.

peterjan (approved):

PR #830, opened 2021-06-04
/skynet/pin/<skylink> endpoint

kwypchlo:
- added /skynet/pin/<skylink> endpoint
- cleaned up config by moving some parts to include files


peterjan (approved):
LGTM 👍

ChrisSchinnerl (dismissed):

PR #878, opened 2021-06-15
fix portal size check reporting zero files

kwypchlo:
Adding "Lost File" into siac broke our health check since we were looking for a line that contained "Files" and it detected two lines:
```
  Files:                608388
  Lost Files:              0
```

I replaced the line matcher with one that specifically checks for the line that starts with "Files:".

### DIFF
--- a/setup-scripts/health-checker.py
+++ b/setup-scripts/health-checker.py
@@ -350,9 +350,9 @@ async def check_portal_size():
     # Initialize variables
     num_files = 0
     max_files = 1500000  # 1.5 mln
-    files_text = "Files"
+    files_text = "Files:"
     for line in siac_renter_output.split("\n"):
-        if contains_string(line, files_text):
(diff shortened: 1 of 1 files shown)


setup-scripts/health-checker.py
```diff
@@ -350,9 +350,9 @@ async def check_portal_size():
     # Initialize variables
     num_files = 0
     max_files = 1500000  # 1.5 mln
-    files_text = "Files"
+    files_text = "Files:"
     for line in siac_renter_output.split("\n"):
-        if contains_string(line, files_text):
+        if line.strip().startswith(files_text):
```
ro-tex, line 355: Can we take it a step further and match with a regex that checks specifically for `^Files:\w+\d+$`?

ro-tex (approved):

ChrisSchinnerl (approved):

Issue #660, opened 2021-04-01
Expose direct address to server via new header


m-cat:
It would be useful for the SDK to sometimes know which actual server from the load balancer is being used.

e.g. This would potentially help us debug consistency issues with the registry.

PR #889, opened 2021-06-16
Support tus with multiple servers setup

kwypchlo:
- add new SKYNET_SERVER_API env variable that like SKYNET_PORTAL_API should point to api endpoint but it should be direct and server specific so SKYNET_PORTAL_API can be https://siasky.net but SKYNET_SERVER_API should be for example https://eu-ger-5.siasky.net
- added new header Skynet-Server-Api alongside Skynet-Portal-Api that exposes SKYNET_SERVER_API
- added health check for SKYNET_SERVER_API
- support tus endpoint proxy rewrite so location header is rewritten to point to specific server directly using SKYNET_SERVER_API because we need one server to handle the upload from start to end and if we load balance, we might accidentally end up with routing user to a different server in between chunks

closes https://github.com/SkynetLabs/skynet-webportal/issues/660

### DIFF
--- a/setup-scripts/setup-docker-services.sh
+++ b/setup-scripts/setup-docker-services.sh
@@ -31,7 +31,7 @@ docker-compose --version # sanity check
 # * AWS_ACCESS_KEY_ID - (optional) if using route53 as a dns loadbalancer
 # * AWS_SECRET_ACCESS_KEY - (optional) if using route53 as a dns loadbalancer
 # * API_PORT - (optional) the port on which siad is listening, defaults to 9980
-# * PORTAL_NAME - the name of the portal
+# * PORTAL_NAME - a string representing name of your portal e.g. `siasky.xyz` or `my skynet portal` (internal use only)
 # * DISCORD_BOT_TOKEN - (optional) only required if you're using the discord notifications integration
 # * SKYNET_DB_USER - (optional) if using `accounts` this is the MongoDB username
 # * SKYNET_DB_PASS - (optional) if using `accounts` this is the MongoDB password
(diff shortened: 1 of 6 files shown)


setup-scripts/setup-docker-services.sh
```diff
@@ -22,15 +22,16 @@ docker-compose --version # sanity check
...
+# * SKYNET_SERVER_API - absolute url to the server api ie. https://eu-ger-1.siasky.net (direct server address, if this is single server portal use the same address as SKYNET_PORTAL_API)
+# * SKYNET_DASHBOARD_URL - (optional) absolute url to the portal dashboard ie. https://account.siasky.net
 # * EMAIL_ADDRESS - this is the administrator contact email you need to supply for communication regarding SSL certification
 # * HSD_API_KEY - this is auto generated secure key for your handshake service integration
 # * CLOUDFLARE_AUTH_TOKEN - (optional) if using cloudflare as dns loadbalancer (need to change it in Caddyfile too)
 # * AWS_ACCESS_KEY_ID - (optional) if using route53 as a dns loadbalancer
 # * AWS_SECRET_ACCESS_KEY - (optional) if using route53 as a dns loadbalancer
 # * API_PORT - (optional) the port on which siad is listening, defaults to 9980
-# * PORTAL_NAME - the name of the portal, required by the discord bot
+# * PORTAL_NAME - the name of the portal
```
m-cat, line 34: nit: How does this differ from SKYNET_PORTAL_API? Should add an example.

setup-scripts/setup-docker-services.sh
```diff
@@ -43,7 +44,7 @@ docker-compose --version # sanity check
 # * CR_CLUSTER_NODES - (optional) if using `accounts` the list of servers (with ports) which make up your CockroachDB cluster, e.g. `helsinki.siasky.net:26257,germany.siasky.net:26257,us-east.siasky.net:26257`
 if ! [ -f /home/user/skynet-webportal/.env ]; then
     HSD_API_KEY=$(openssl rand -base64 32) # generate safe random key for handshake
-    printf "SSL_CERTIFICATE_STRING=example.com, *.example.com, *.hns.example.com\nSKYNET_PORTAL_API=https://example.com\nSKYNET_DASHBOARD_URL=https://account.example.com\nEMAIL_ADDRESS=email@example.com\nSIA_WALLET_PASSWORD=\nHSD_API_KEY=${HSD_API_KEY}\nCLOUDFLARE_AUTH_TOKEN=\nAWS_ACCESS_KEY_ID=\nAWS_SECRET_ACCESS_KEY=\nPORTAL_NAME=\nDISCORD_BOT_TOKEN=\n" > /home/user/skynet-webportal/.env
+    printf "SSL_CERTIFICATE_STRING=siasky.net, *.siasky.net, *.hns.siasky.net\nSKYNET_PORTAL_API=https://siasky.net\nSKYNET_SERVER_API=https://siasky.net\nSKYNET_DASHBOARD_URL=https://account.example.com\nEMAIL_ADDRESS=email@example.com\nSIA_WALLET_PASSWORD=\nHSD_API_KEY=${HSD_API_KEY}\nCLOUDFLARE_AUTH_TOKEN=\nAWS_ACCESS_KEY_ID=\nAWS_SECRET_ACCESS_KEY=\nPORTAL_NAME=\nDISCORD_BOT_TOKEN=\n" > /home/user/skynet-webportal/.env
```
m-cat, line 47: Can you make SKYNET_SERVER_API contain a specific server here for purposes of clarity?

m-cat (dismissed):

ro-tex (commented):
LGTM, aside from Marcin's remarks which are valid. Once those are address I will approve this.

MSevey (dismissed):
LGTM

ro-tex (approved):

PR #891, opened 2021-06-17
clean up webapp directory and update some deps

kwypchlo:
- removes unused webapp directory that being there caused build to be significantly longer due to package.json dependencies
- updated some dependencies
- removed unused square dependency


MSevey (dismissed):
Was the original Idea behind leaving in the webapp/ was that it could be used by community webportals? If so, is there a way to leave it in and just not include it in our build?

kwypchlo:
> Was the original Idea behind leaving in the webapp/ was that it could be used by community webportals? If so, is there a way to leave it in and just not include it in our build?

I don't think it's on the table any more, we'll most likely want to prepare a version of a current portal if we want to make something custom for community

ro-tex (approved):

MSevey (approved):

PR #892, opened 2021-06-18
support dynamic max upload size on tus endpoint

kwypchlo:


ChrisSchinnerl (approved):

docker/nginx/conf.d/client.conf
```diff
@@ -337,6 +337,23 @@ server {
...
+		# set max upload size dynamically based on account limits
+		rewrite_by_lua_block {
+			-- set default limit value to 1 GB
+			ngx.req.set_header("SkynetMaxUploadSize", 1073741824)
+
+			-- this block runs only when accounts are enabled
+			if os.getenv("ACCOUNTS_ENABLED", "0") == "0" then return end
+
+			-- fetch account limits and set max upload size accordingly
+			local res = ngx.location.capture("/accounts/user/limits", { copy_all_vars = true })
```
ro-tex, line 349: This hits `accounts` before each upload. Which leads to a DB hit. That's not fast enough to accommodate this use case. We need to cache it somewhere along the line before we can deploy this. I would actually prefer to do it before we even merge this, just to avoid the possibility of accidentally pushing it to prod and melting the DB, thus killing the entire prod.

We can cache it at any point, really, but closest to nginx will be best because of the saved network chatter. If it's hard to add it here, I can add the next best thing in `accounts`.

ro-tex: Added a F/U issue here: https://github.com/SkynetLabs/skynet-webportal/issues/901

Again, I would prefer to not merge this right now in order to avoid the risks outlined above. But there is nothing wrong with the PR, so I'll approve it.

ro-tex (approved):

PR #903, opened 2021-06-22
resolve ip mismatch on some of the health checks

kwypchlo:
originally this branch was used to debug issues with dnsmasq, findings:
  - dnsmasq works correctly and all queries were logged and resolved properly
  - despite above, some requests in health-checks ended up with different ip, reason unknown

**found issue**
/etc/resolv.conf was unreliable with the resolution - sometimes it would ask dnsmasq about a route but use docker nameserver to resolve anyway

**change that solve the above:**
moved docker nameserver to dnsmasq to ensure proper resolution order of resolution
  - /etc/resolv.conf had docker nameserver, now it's replaced with dnsmasq nameserver
  - dnsmasq configured to resolve siasky.net with current server ip
  - dnsmasq configured to resolve other request than siasky.net with the origincal docker nameserver
  - this way whole resolving is contained in dnsmasq and we can rely on the ordering of resolution because /etc/resolv.conf was unreliable in that matter

**also:**
- added ip check to ALL checks just so we check on every request that the reponse ip is the actual server ip
- moved address alias from ad hoc file to cli argument just for clarity
- removed excessive "--max-http-header-size=64000" from health check (not needed since we don't return metadata)

### DIFF
--- a/packages/health-check/Dockerfile
+++ b/packages/health-check/Dockerfile
@@ -23,8 +23,8 @@ ENV NODE_ENV production
 # 3. start crond in the background
 # 4. start the health-check api service
 CMD [ "sh", "-c", \
-      "echo -e \"nameserver 127.0.0.1\n$(cat /etc/resolv.conf)\" > /etc/resolv.conf ; \
-       dnsmasq --strict-order --no-resolv --address=/siasky.net/$(node src/whatismyip.js) ; \
+      "echo -e \"nameserver 127.0.0.1\noptions ndots:0\" > /etc/resolv.conf ; \
+       dnsmasq --server=127.0.0.1 --address=/siasky.net/$(node src/whatismyip.js) ; \
        crond ; \
        node --max-http-header-size=64000 src/index.js" \
     ]
--- a/packages/health-check/Dockerfile
+++ b/packages/health-check/Dockerfile
@@ -24,9 +24,8 @@ ENV NODE_ENV production
 # 4. start crond in the background
 # 5. start the health-check api service
 CMD [ "sh", "-c", \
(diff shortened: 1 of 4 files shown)


peterjan (approved):

packages/health-check/Dockerfile
```diff
@@ -18,15 +18,15 @@ COPY cli cli
...
+#    - default docker nameserver 127.0.0.11 for any other request
+# 2. replace docker nameserver with dnsmasq nameserver in /etc/resolv.conf
+# 3. start crond in the background to schedule periodic health checks
+# 4. start the health-check api service
 CMD [ "sh", "-c", \
-      "echo address=/siasky.net/$(node src/whatismyip.js) > /etc/dnsmasq.d/siasky.net.conf ; \
-       echo -e \"nameserver 127.0.0.1\n$(cat /etc/resolv.conf)\" > /etc/resolv.conf ; \
-       dnsmasq ; \
+      "dnsmasq --no-resolv --log-facility=/var/log/dnsmasq.log --address=/siasky.net/$(node src/whatismyip.js) --server=127.0.0.11 ; \
+       echo \"$(sed 's/127.0.0.11/127.0.0.1/' /etc/resolv.conf)\" > /etc/resolv.conf ; \
```
ro-tex, line 29: This is a little curious. Why set 127.0.0.11 and then replace it with 127.0.0.1?

ro-tex (approved):

packages/health-check/Dockerfile
kwypchlo: because 127.0.0.1 is a docker nameserver and we need to use dnsmasq nameserver which is 127.0.0.1

then in dnsmasq we internally use docker nameserver 127.0.0.1 and also define override for siasky.net

PR #933, opened 2021-06-30
upgrade website to skynet-js 4 beta

kwypchlo:


packages/website/src/components/Uploader/Uploader.js
```diff
@@ -201,7 +201,7 @@ const Uploader = () => {
             response = await client.uploadFile(file, { onUploadProgress });
           }
 
-          const url = client.getSkylinkUrl(response.skylink, { subdomain: mode === "directory" });
+          const url = await client.getSkylinkUrl(response.skylink, { subdomain: mode === "directory" });
```
peterjan, line 204: I don't think you caught all occurrences. This is also called from the `Dashboard` and from an index.js file somewhere.

ro-tex (approved):

peterjan (changes requested):
think you missed some occurrences, other than that it looks good 👍

packages/website/src/components/Uploader/Uploader.js
kwypchlo: I only updated website package, dashboard is a separate one

PR #942, opened 2021-07-05
dnslink beta support

kwypchlo:
**beta** release of dnslink support for skynet
- create a CNAME entry for your domain (example.com) to siasky.net
- create a TXT entry for _dnslink prefixed domain (_dnslink.example.com) and set it to `dnslink=/skynet-ns/<skylink>` and replace `<skylink>` with the actual skylink
- it can take up to 5-10 minutes for the configuration to propagate (skynet portals cache the configuration for 5 minutes if you are updating existing entry and dns propagation can take anywhere up to 5 minutes)

improvements to be made in short term for a full release:
- [x] better configuration error reporting - misconfiguration will now return 400 http code and meaningful, user friendly error
- [ ] discussion what should be the dnslink prefix for skynet (currently it's /skynet-ns/<skylink> becuase that's what @PJ put in initial push to metamask)


packages/dnslink-api/src/index.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    const dnslinks = records.flat().filter((record) => dnslinkRegExp.test(record));
+
+    if (dnslinks.length === 0) {
+      return failure(`TXT records for ${lookup} found but none of them contained valid dnslink - ${hint}`);
+    }
+
+    if (dnslinks.length > 1) {
+      return failure(`Multiple TXT records with valid dnslink found for ${lookup}, only one allowed`);
```
dghelm, line 54: Is this limited to 1 _dnslink TXT record or one that starts with `/skynet-ns/`? If a site wants to support multiple protocols, that seems fine and we'd still be able to determine what to do serve.

kwypchlo: fine point but unfortunately the most referenced ipfs resolver does that https://github.com/JChanceHud/dnslink/blob/master/cmds/resolve.js#L41 but yeah I think I can make sure that we don't have 2 dnslink entries with /skynet-ns/ prefix instead

docker/nginx/conf.d/client.conf
```diff
@@ -444,6 +450,31 @@ server {
...
+		rewrite_by_lua_block {
+			local http = require("socket.http")
+			local ok, statusCode, headers, statusText = http.request {
+				url = "http://dnslink-api:3100/dnslink/" .. ngx.var.host
+			}
+
+			if statusCode == ngx.HTTP_OK then
+				ngx.var.dnslink = headers["skynet-skylink"]
+			else
+				ngx.status = statusCode
```
peterjan, line 467: `nit` this is redundant I believe because of `ngx.exit`

packages/dnslink-api/src/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+const host = process.env.HOSTNAME || "0.0.0.0";
+const port = Number(process.env.PORT) || 3100;
+
+const server = express();
+const cache = new NodeCache({ stdTTL: 300 }); // cache for 5 minutes
+
+const dnslinkRegExp = /^dnslink=.+$/;
+const skylinkDnslinkRegExp = /^dnslink=\/skynet-ns\/([a-zA-Z0-9_-]{46}|[a-z0-9]{55})/;
+const hint = "valid example: dnslink=/skynet-ns/3ACpC9Umme41zlWUgMQh1fw0sNwgWwyfDDhRQ9Sppz9hjQ";
```
peterjan, line 14: having the example link to Bitcoin whitepaper, gj 👍

packages/dnslink-api/src/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+    if (!matchSkylink) {
+      return failure(`TXT record with dnslink for ${lookup} contains invalid skylink - ${hint}`);
+    }
+
+    const skylink = matchSkylink[1];
+
+    cache.set(req.params.name, skylink);
+
+    console.log(`${req.params.name} => ${skylink}`);
```
peterjan, line 68: is this a debug statement we want to keep? Would be fine, just checking.

packages/dnslink-api/src/index.js
```diff
@@ -0,0 +1,78 @@
+const dns = require("dns");
+const express = require("express");
+const NodeCache = require("node-cache");
+const isValidDomain = require("is-valid-domain");
+
+const host = process.env.HOSTNAME || "0.0.0.0";
+const port = Number(process.env.PORT) || 3100;
+
+const server = express();
+const cache = new NodeCache({ stdTTL: 300 }); // cache for 5 minutes
```
peterjan, line 10: might be worthwhile making this an environment variable

peterjan (approved):
LGTM 👍  nice

packages/dnslink-api/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM node:16.3.0-alpine
+
+WORKDIR /usr/app
+
+COPY package.json .
+RUN yarn --no-lockfile
```
ro-tex, line 6: This is interesting. Why are we running with `--no-lockfile`? I usually hear arguments for always using the lockfile without updating it (in prod, you obviously update it during dev), as that guarantees that you'll be running the same module versions in production that you tested with during your CI run. Running without a lockfile will always get you the latest compatible versions but has a much weaker guarantee that those version will work sufficiently well together.

packages/dnslink-api/src/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+const dnslinkRegExp = /^dnslink=.+$/;
+const skylinkDnslinkRegExp = /^dnslink=\/skynet-ns\/([a-zA-Z0-9_-]{46}|[a-z0-9]{55})/;
```
ro-tex, line 13: We should define a skylink-matching regex somewhere and reuse it. We have a bunch of those defined across the project, most only matching 46-char skylinks.

ro-tex (dismissed):
LGTM.

docker/nginx/conf.d/client.conf
kwypchlo: no, ngx.exit will only exit from the rewrite block but the code you set there is only to indicate whether you want to break the flow (>= 400) or continue it (< 400) - it would return an error message with code 200 if not for this line - I know, really stupid :D

packages/dnslink-api/src/index.js
kwypchlo: Yeah I wanted that in docker logs, we're rotating them anyway so no harm done but I really want to see the requests being resolved (similar logging as in hns api)

packages/dnslink-api/src/index.js
kwypchlo: yeah but concatenating regexes is not possible, we could however have it as a string and then initialise with `new RegExp(regexString)` - I guess we could put that in skynet-js or sth

packages/dnslink-api/Dockerfile
kwypchlo: yeah that's due to how our package system works, we have one lock file for whole skynet-webportal repo becuase we're using workspaces but I guess we should really just treat it as a monorepo with multiple packages and have separate lock for each package - we can put that on the roadmap for sometime in the future (making new issue for that now)

packages/dnslink-api/src/index.js
kwypchlo: added support for multiple dnslink entries, we just need one that has skynet-ns

docker/nginx/conf.d/client.conf
peterjan: lol

peterjan (approved):

ro-tex (approved):

PR #946, opened 2021-07-07
Fix skylink v2 redirect

kwypchlo:
### problem

When accessing a skylink v2 skapp without the trailing slash, we would get redirected to a link with trailing slash BUT to underlying skylink v1 and not the skylink v2 that we requested.

### solution

replace skylink v1 with skylink v2 in skyd location header response if we were fetching skylink v1 for v2 request

### verification

verify with `AQBUUNFGvF261JuvCZjEBQILdfB1UqVmaWuLS8MKPv82Yw` skylink for example (currently not deployed anywhere - ping me for tests)

### DIFF
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -440,7 +440,7 @@ server {
 
 		proxy_read_timeout 600;
 		proxy_set_header User-Agent: Sia-Agent;
-		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is correct)
+		proxy_redirect $skylink_v1 $skylink_v2
 		proxy_pass http://siad/skynet/skylink/$skylink_v1$path$is_args$args;
 	}
 
--- a/docker/nginx/conf.d/client.conf
+++ b/docker/nginx/conf.d/client.conf
@@ -409,12 +409,16 @@ server {
 		set $skylink $2;
 		set $path $3;
 
-		# v2 support
+		# $skylink_v1 and $skylink_v2 variables default to the same value but in case the requested skylink was:
+		# a) skylink v1 - it wouldn't matter, no additional logic is executed
+		# b) skylink v2 - in a lua block below we will resolve the skylink v2 into skylink v1 and update 
+		#      $skylink_v1 variable so then the proxy request to skyd can be cached in nginx (proxy_cache_key 
+		#      in proxy-cache-downloads includes $skylink_v1 as a part of the cache key)
 		set $skylink_v1 $skylink;
 		set $skylink_v2 $skylink;
 
 		access_by_lua_block {
-			-- disable cache if this is skylink v2
+			-- detect whether requested skylink is v2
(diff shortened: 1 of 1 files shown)


ro-tex (dismissed):

docker/nginx/conf.d/client.conf
```diff
@@ -440,7 +440,7 @@ server {
 
 		proxy_read_timeout 600;
 		proxy_set_header User-Agent: Sia-Agent;
-		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is correct)
```
m-cat, line 443: Q: Why remove this comment?
Q: Should we add a new comment so that future readers can understand the purpose for this line? (You can always blame but the commit message is not that helpful).

docker/nginx/conf.d/client.conf
```diff
@@ -440,7 +440,7 @@ server {
...
+		proxy_redirect $skylink_v1 $skylink_v2;
 		proxy_pass http://siad/skynet/skylink/$skylink_v1$path$is_args$args;
```
m-cat, line 444:
```suggestion
		proxy_pass http://siad/skynet/skylink/$skylink_v2$path$is_args$args;
```

Q: Is this not sufficient for the fix?

kwypchlo: it would but this would trigger registry call on skyd because skyd would have to resolve the skylink v2 and we already have it resolved

docker/nginx/conf.d/client.conf
kwypchlo: this comment was kind of a leftover from time where this function used ip and we're using alias now which makes the target self explanatory

docker/nginx/conf.d/client.conf
```diff
@@ -440,7 +440,7 @@ server {
 
 		proxy_read_timeout 600;
 		proxy_set_header User-Agent: Sia-Agent;
-		# proxy this call to siad /skynet/skylink/ endpoint (make sure the ip is correct)
+		proxy_redirect $skylink_v1 $skylink_v2;
```
peterjan, line 443: I think this warrants a comment. But more importantly, why are we fixing this in nginx as opposed to fixing it in `skyd`? I think if `skyd` is triggering the redirect, it should just be smart enough to redirect it to the requested skylink... Unless ofcourse nginx is doing the v2->v1 resolving.

ro-tex: 1. A comment here is a really good idea - I had to google what this does when reviewing. I should have asked for a comment.
2. It's Nginx going the resolving. Here's the discussion: https://discord.com/channels/542938080349519882/834800136092909578/862318336531759164

kwypchlo: Yeah it's nginx that does the resolving because we want to potentially return the resource from nginx cache. I will add a proper comment to this section.

ro-tex (approved):

peterjan (approved):

PR #970, opened 2021-07-09
fix issue with "unexpected end of file" in health checks on subdomain test

kwypchlo:
- increase proxy timeouts to a higher than expected value on nginx endpoints that proxy to a different nginx location and let the location that proxies to an external resource manage the actual timeouts
- remove `/` in front of `$request_uri` because it already contains it and results in double slash otherwise


peterjan (approved):
Are we including this `proxy-pass-internal` in routes where we don't overwrite the timeouts? If so this PR is drastically changing the behaviour for those routes and we should verify whether that behaviour is wanted in all cases where that is true.

kwypchlo:
> timeouts

default proxy timeouts apply either way so we're not changing anything on those finishing locations
http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout

kwypchlo:
> Are we including this `proxy-pass-internal` in routes where we don't overwrite the timeouts? If so this PR is drastically changing the behaviour for those routes and we should verify whether that behaviour is wanted in all cases where that is true.

default proxy timeouts apply either way so we're not changing anything on those finishing locations
http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_read_timeout

docker/nginx/conf.d/client.conf
```diff
@@ -480,19 +484,21 @@ server {
 		}
 
 		proxy_set_header Dnslink-Lookup "";
-		proxy_pass http://127.0.0.1/$dnslink/$request_uri;
+		proxy_pass http://127.0.0.1/$dnslink$request_uri;
```
ro-tex, line 487: Do we ever verify that the `dnslink` ends with a slash? I see we're grabbing it from the headers in 
```
ngx.var.dnslink = headers["skynet-skylink"]
```
but I'm not sure we can rely on that being always valid.

Can we verify that it's either empty or ends with a slash?

ro-tex (approved):
LGTM, as long as the concern about the potential missing trailing slash is unfounded.

docker/nginx/conf.d/client.conf
kwypchlo: it's not that $dnslink ends with slash, it's $request_uri that always starts with one :)

PR #982, opened 2021-07-18
airtable retries logic

kwypchlo:
https://support.airtable.com/hc/en-us/articles/203313985-Public-REST-API

> **Is there a rate limit?** 
Yes, the public API has a rate limit of 5 requests per second, per base. This limit is the same across all pricing tiers and increased rate limits are not currently available. For more information, see the "Rate Limits" section of the API documentation.  

With 20+ servers running these scripts at the same time and each making 20+ paginated requests, we're 100% running into rate limiting, retries logic should help here (that's what airtable docs suggested too).

Also applied the change to nginx cache busting script from https://github.com/SkynetLabs/skynet-webportal/pull/978/commits/339310e90c99a845a0f42c5a13d85fd738c19202


setup-scripts/support/logrotate
```diff
@@ -0,0 +1,13 @@
...
+    rotate 3650
+    missingok
+    notifempty
+    nocompress
+    create 640 root root
+    sharedscripts
+    postrotate
+        docker exec nginx sh -c '! test -f /usr/local/openresty/nginx/logs/nginx.pid' || docker exec nginx sh -c 'kill -USR1 $(cat /usr/local/openresty/nginx/logs/nginx.pid)'
+    endscript
+}
```
peterjan, line 13: add newline at end of file

peterjan (dismissed):

setup-scripts/support/logrotate
kwypchlo: oh shit that file was not supposed to make it there, removing

peterjan (approved):

ChrisSchinnerl (approved):

PR #1018, opened 2021-07-26
maintenance updates to repo

kwypchlo:
- ~update default branch to "main" (already updated in the repo)~ - we decided not to introduce it right now
- update node
- remove package.json from main repo - we don't do workspaces any more
- added yarn.lock files to every npm project, added to Dockerfile for more reliable builds and replaced every npm call with yarn


ro-tex (approved):

packages/website/gatsby-config.js
```diff
@@ -1,7 +1,10 @@
 const { defaultIcons } = require("gatsby-plugin-manifest/common");
 
 module.exports = {
-  flags: { PRESERVE_WEBPACK_CACHE: true },
+  flags: {
+    // PARALLEL_QUERY_RUNNING: true,
+    // PARALLEL_SOURCING: true,
```
ChrisSchinnerl, line 6: Can we remove the commented out stuff?

ChrisSchinnerl:
I don't think we should rename master just like that. This is a big change and might need more communication to let people know that it's happening and that things might break.

So I'd merge these changes into master and start a discussion about how we can painlessly move over to a main branch.

ChrisSchinnerl (changes requested):

ChrisSchinnerl (approved):

ro-tex (approved):

PR #1022, opened 2021-07-26
Fix server list and use common NGINX command

MSevey:
- `eu-ger-8.siasky.net` had a typo
- I split out the siasky.dev servers 
- I copied over the NGINX command from the airtable command

### DIFF
--- a/scripts/blocklist-skylink.sh
+++ b/scripts/blocklist-skylink.sh
@@ -50,7 +50,7 @@ do
     for skylink in "${skylinks[@]}";
     do
         echo ".. ⌁ Blocking skylink ${skylink} on ${server}"
-        cached_files_command="find /data/nginx/cache/ -type f | xargs --no-run-if-empty grep -Elsq '^Skynet-Skylink: ${skylink}'"
+        cached_files_command="find /data/nginx/cache/ -type f | xargs -r grep -Elsq '^Skynet-Skylink: ${skylink}'"
         ssh -q -t user@${server} "docker exec -it nginx bash -c ${cached_files_command} | xargs -r rm"
         echo ".. ⌁ Skylink ${skylink} Blocked on ${server}"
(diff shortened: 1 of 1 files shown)


scripts/blocklist-skylink.sh
```diff
@@ -32,21 +37,23 @@ fi
...
-                      "siasky.xyz" "siasky.dev")
+                      "siasky.xyz" "dev1.siasky.dev" "dev2.siasky.dev" "dev3.siasky.dev")
 for server in "${servers[@]}";
 do
     for skylink in "${skylinks[@]}";
     do
         echo ".. ⌁ Blocking skylink ${skylink} on ${server}"
-
-        ssh -q -t user@${server} "docker exec sia siac skynet blocklist add $skylink && docker exec nginx curl -s -i -X PURGE http://localhost/$skylink | egrep \"^(OK|HTTP|X-)\""
+        cached_files_command="find /data/nginx/cache/ -type f | xargs --no-run-if-empty grep -Elsq '^Skynet-Skylink: ${skylink}'"
```
kwypchlo, line 53: is `q` searching just for first file and it stops or it just searches files until it finds a match but still searches all files?

kwypchlo:
```suggestion
        cached_files_command="find /data/nginx/cache/ -type f | xargs -r grep -Elsq '^Skynet-Skylink: ${skylink}'"
```

shorter syntax

MSevey: > is `q` searching just for first file and it stops or it just searches files until it finds a match but still searches all files?

q was to suppress the output. Running it manually I didn't want to see all the files that were being grepped.

MSevey: > shorter syntax

added

kwypchlo (approved):

PR #1023, opened 2021-07-27
use skydb to store ssl certs and share them with other portals in a cluster

kwypchlo:
WIP


kwypchlo:
not relevant any more due to https://github.com/SkynetLabs/skynet-webportal/pull/1880

PR #1029, opened 2021-08-03
Sequential uploads on siasky.net website

kwypchlo:
As a precaution measure to extensive memory allocation on a single server in a case of too many parallel requests, we're making uploads from our website sequential with possibility to parallelise them in future depending on the development progress.


ChrisSchinnerl (dismissed):

peterjan:
Is this precaution really necessary? I think it makes the demo less cool, which is important imo. Can we do batches instead?

kwypchlo:
> Is this precaution really necessary? I think it makes the demo less cool, which is important imo. Can we do batches instead?

Yeah we run out of client memory on skyd side pretty easily. With this pull request, we will be able to increase the number of parallel uploads from 1 to any number with just one variable if we feel it's safe enough.

packages/website/src/components/Uploader/UploaderElement.js
```diff
@@ -0,0 +1,180 @@
...
+import bytes from "bytes";
+import { getReasonPhrase, StatusCodes } from "http-status-codes";
+import copy from "copy-text-to-clipboard";
+import path from "path-browserify";
+import { SkynetClient } from "skynet-js";
+import { useTimeoutFn } from "react-use";
+import ms from "ms";
+import Link from "../Link";
+
+const TOO_MANY_REQUESTS_RETRY = ms("5s"); // retry delay after "492: TOO_MANY_REQUESTS"
```
peterjan, line 13: do we need an exponential backoff? thinking about a situation where the retry effectively keeps pushing us over the rate limit...

peterjan: `nit` 429

kwypchlo:
```suggestion
const TOO_MANY_REQUESTS_RETRY = ms("5s"); // retry delay after "429: TOO_MANY_REQUESTS"
```

peterjan (dismissed):

packages/website/src/components/Uploader/UploaderElement.js
kwypchlo: > do we need an exponential backoff? thinking about a situation where the retry effectively keeps pushing us over the rate limit...

added delay multiplication in https://github.com/SkynetLabs/skynet-webportal/pull/1029/commits/6bb7945fbfd81ba23affff680c09f7f65183491b

ChrisSchinnerl (approved):

ro-tex (approved):

PR #1058, opened 2021-08-09
bump worker_connections number and auto detect worker_processes number

kwypchlo:
We're getting errors:
> 1024 worker_connections are not enough

This means we need to increase number of connections available on the servers since it's either ddos or heavy load - currently skyd is under higher load and response times are higher so there is more open connections than usual.

### DIFF
--- a/docker/nginx/nginx.conf
+++ b/docker/nginx/nginx.conf
@@ -16,8 +16,8 @@
 # See https://github.com/openresty/docker-openresty/blob/master/README.md#nginx-config-files
 #
 
-user  root;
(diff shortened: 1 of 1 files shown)


peterjan (approved):

peterjan:
I approved but I think we should also very the amount of open files on our servers by running `ulimit -a` (I think?).
The number configured here should not be greater than that. I think it's set to 25k on our servers, in which case the 8192 is pretty conservative but probably fine for now.

MSevey (approved):

PR #1060, opened 2021-08-11
increase min uses and cache size

kwypchlo:
- increase cache size to 50g
- increase valid and inactive time of cached assets to 48 hours
- also cache 307 responses


peterjan (approved):
Super excited about the 307 cache! That's awesome!

ro-tex (approved):

PR #1062, opened 2021-08-11
Update handshake chain migration in docker compose

MSevey:
When setting up a new portal I found that this command needs to be commented out in order for a new handshake node to start. 

It should stay in as a comment for existing portals

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -103,9 +103,8 @@ services:
     build:
       context: ./docker/handshake
       dockerfile: Dockerfile
-  # Uncomment this line if you need to migrate the handshake db for the chain
-  # migration.  For new portal set ups this line does not appear to be needed.
-  # command: --chain-migrate=1 --wallet-migrate=1
+    # Migration.  
+    command: --chain-migrate=1 --wallet-migrate=1
     container_name: handshake
     restart: unless-stopped
     logging: *default-logging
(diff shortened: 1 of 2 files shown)


kwypchlo (commented):
AFAIK handshake devs said that it can stay there since it will only run the migration if it's not be ran before. I don't think there's reason to comment it out.

MSevey:
> AFAIK handshake devs said that it can stay there since it will only run the migration if it's not be ran before. I don't think there's reason to comment it out.

As I said in my initial comment, the reason to comment it out is it doesn't work. I'm setting up portals right now and handshake won't start until this is commented out.

kwypchlo (dismissed):

ro-tex (approved):

MSevey:
wait on this. testing a potential fix from the handshake devs

MSevey:
Handshake added a fix, once https://github.com/handshake-org/hsd/pull/628 is merged I will update the dockerfile to point to the new commit that contains the fix.

kwypchlo:
@MSevey https://github.com/handshake-org/hsd/pull/628 already merged :)

docker-compose.yml
```diff
@@ -103,6 +103,7 @@ services:
     build:
       context: ./docker/handshake
       dockerfile: Dockerfile
+    # Migration.
```
kwypchlo, line 106: this seems redundant

```suggestion
```

kwypchlo (approved):

ro-tex (dismissed):

ro-tex (approved):

kwypchlo (approved):

PR #1063, opened 2021-08-12
increase disk space size that warrants a warning

interstella-5555:
with cache size increased to 50 GB (https://github.com/SkynetLabs/skynet-webportal/pull/1060) we need to make sure we always have a good buffer of disk space and we don't run up due to any unforeseen events like logs overflow from different services


peterjan (approved):

peterjan:
@kwypchlo probably 60GiB is  bit on the low side even for critical

ro-tex (approved):

Issue #1056, opened 2021-08-09
`skynet-portal-api` and `skynet-server-api` header is missing in regular GET calls.


kwypchlo:
https://siasky.net/XABvi7JtJbQSMAcDwnUnmp2FKDPjg8_tTTFP4BwMSxVdEg

PR #1072, opened 2021-08-16
add missing skylink headers

kwypchlo:
`header_filter_by_lua_block` can only be called once in the request lifecycle so if location uses it, the one from server block defined in nginx.conf is not executed - this is why we just need to copy&paste the headers to every header_filter_by_lua_block block we have

closes https://github.com/SkynetLabs/skynet-webportal/issues/1056


MSevey (approved):

m-cat (approved):
Looks good.

PR #1074, opened 2021-08-16
include proof on resolver skylink requests

kwypchlo:
In case of a resolver skylink request, we make a subrequest to resolve it into regular skylink and request the regular immutable skylink from skyd (because we need it to be cacheable request in nginx). Resolver skylink responses should include proof header so we need to add it after we get a regular skylink response.

☝️  this was already included in https://github.com/SkynetLabs/skynet-webportal/pull/1075 but this PR aims to rename the variables and simplify the description a bit

### DIFF
--- a/docker/nginx/conf.d/include/cors
+++ b/docker/nginx/conf.d/include/cors
@@ -13,4 +13,4 @@ more_set_headers 'Access-Control-Allow-Origin: $http_origin';
 more_set_headers 'Access-Control-Allow-Credentials: true';
 more_set_headers 'Access-Control-Allow-Methods: GET, POST, HEAD, OPTIONS, PUT, PATCH, DELETE';
 more_set_headers 'Access-Control-Allow-Headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,Cache-Control,Content-Type,Range,X-HTTP-Method-Override,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,location';
-more_set_headers 'Access-Control-Expose-Headers: Content-Length,Content-Range,Skynet-File-Metadata,Skynet-Skylink,Skynet-Proof,Skynet-Portal-Api,Skynet-Server-Api,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,location';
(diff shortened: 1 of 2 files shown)


m-cat:
@kwypchlo Under what conditions is the proof equal to `""`? I feel like the header should always be present, even if there is no proof in which case it should be set to `[]`.

MSevey (approved):

kwypchlo:
> @kwypchlo Under what conditions is the proof equal to `""`? I feel like the header should always be present, even if there is no proof in which case it should be set to `[]`.

It's not the header that equals to `""` but the nginx variable that I set with `set $skynet_proof '';` - nginx doesn't have null value and I had to initialise this variable with something. This variable only contains the proof if it gets it from skylink v2 /resolve endpoint. 

```lua
if isBase32v2 or isBase64v2 then
    local res = ngx.location.capture("/skynet/resolve/" .. ngx.var.skylink_v2)
    if res.status == ngx.HTTP_OK then
	    local json = require('cjson')
	    local resolve = json.decode(res.body)
	    ngx.var.skylink_v1 = resolve.skylink
	    ngx.var.skynet_proof = res.header["Skynet-Proof"]
    end
end
```

m-cat (commented):
> This variable only contains the proof if it gets it from skylink v2 /resolve endpoint.

Just to confirm, this means the variable will always be set to `[]` for v1 links and not `""`?

I feel like the if-conditional is slightly redundant which is why I was confused. It can be `ngx.header["Skynet-Proof"] = ngx.var.skynet_proof ?? "[]"` or the equivalent in Lua (or whatever language that is), not a big deal though.

m-cat:
I can see the changes on `siasky.xyz` currently. In general it would be nice if PRs linked out to a deployed version so reviewers can verify the changes. Also, if they aren't already, PRs should be running a full test suite on the deployed version (e.g. the existing SDK integration tests) to catch regressions.

m-cat (changes requested):
I still feel like the v1 case needs to be documented. It should at least say that the proof variable contains `[]` and not `""` (assuming that is the case).

kwypchlo:
I made some changes in the pull request and also included new skynet-requested-skylink response header in.

MSevey (approved):

ro-tex (approved):

kwypchlo:
not needed any more for now, already implemented in #1075

PR #1075, opened 2021-08-17
remove caddy proxy

kwypchlo:
_work in progress_


ro-tex:
We have two copies of `mo` which seems kind of odd. Do we need both?

docker/nginx/nginx.conf
```diff
@@ -64,18 +63,70 @@ http {
...
+    proxy_set_header X-Forwarded-Proto $scheme;
+
+    # since we are proxying request to nginx from caddy, access logs will contain caddy's ip address
+    # as the request address so we need to use real_ip_header module to use ip address from
+    # X-Forwarded-For header as a real ip address of the request
+    set_real_ip_from  10.0.0.0/8;
+    set_real_ip_from  127.0.0.1/32;
+    set_real_ip_from  172.16.0.0/12;
+    set_real_ip_from  192.168.0.0/16;
+    real_ip_header    X-Forwarded-For;
```
ro-tex, line 121: This part is now obsolete, right?

docker/nginx/conf.d/include/track-upload
```diff
@@ -5,14 +5,13 @@ log_by_lua_block {
...
-        local ok, statusCode, headers, statusText = http.request {
-            url = "http://accounts:3000/track/upload/" .. skylink,
+        local httpc = require("resty.http").new()
+        local res, err = httpc:request_uri("http://10.10.10.70:3000/track/upload/" .. skylink, {
             method = "POST",
-            headers = ngx.req.get_headers()
-        }
-        if statusCode ~= ngx.HTTP_NO_CONTENT and statusCode ~= ngx.HTTP_UNAUTHORIZED then
-            ngx.log(ngx.ERR, "accounts endpoint /track/upload/" .. skylink .. " failed with error " .. statusCode)
+            headers = { ["Cookie"] = ngx.header["Cookie"] },
```
ro-tex, line 11: This syntax seems so off... I'm not saying it's wrong, I'm just saying it looks very weird to me, so please take a look again. :)

docker/nginx/conf.d/include/track-upload
```diff
@@ -5,14 +5,13 @@ log_by_lua_block {
 
     local skylink = ngx.header["Skynet-Skylink"]
     if skylink and ngx.status >= ngx.HTTP_OK and ngx.status < ngx.HTTP_SPECIAL_RESPONSE then
-        local http = require("socket.http")
-        local ok, statusCode, headers, statusText = http.request {
-            url = "http://accounts:3000/track/upload/" .. skylink,
+        local httpc = require("resty.http").new()
+        local res, err = httpc:request_uri("http://10.10.10.70:3000/track/upload/" .. skylink, {
```
ro-tex, line 9: Is there a reason to use IPs instead of names? I see you switched from `accounts` to `10.10.10.70` - is that for easier find-and-replace type of task or is there a different reason? I find it a little harder to read (constantly checking docker-compose) but if that works better for you, that's OK.

docker/nginx/conf.d/include/track-registry
```diff
@@ -4,15 +4,14 @@ log_by_lua_block {
     if os.getenv("ACCOUNTS_ENABLED") ~= "true" then return end
 
     if ngx.status == ngx.HTTP_OK or ngx.status == ngx.HTTP_NOT_FOUND then
-        local http = require("socket.http")
+        local httpc = require("resty.http").new()
```
ro-tex, line 7: I would call this `client`. `httpc` immediately sets off my alarm bells.

If I forget to switch back from cyrillic to latin layout and I try to change `http` to `https` it will end up `httpс`. :)

docker/nginx/conf.d/include/ssl-settings
```diff
@@ -0,0 +1,13 @@
+# https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&hsts=false&ocsp=false&guideline=5.6
+
+ssl_session_timeout 1d;
+ssl_session_cache shared:MozSSL:10m;  # about 40000 sessions
+ssl_session_tickets off;
+
+# curl https://ssl-config.mozilla.org/ffdhe2048.txt > /path/to/dhparam
```
ro-tex, line 7: Is this comment still valid? Aren't we generating a new one when we build the container?

kwypchlo:
> We have two copies of `mo` which seems kind of odd. Do we need both?

unfortunately yes because they are included in 2 separate docker contexts and they don't share any directory where we could move mo to

I considered doing a curl and fetching the latest mo in Dockerfile but I thought we'll better off if we have static copies, we're less open to someone maliciously hijacking the library and leaking environment variables

packages/dnslink-api/src/index.js
```diff
@@ -16,8 +16,8 @@ const dnslinkSkylinkRegExp = new RegExp(`^dnslink=/${dnslinkNamespace}/([a-zA-Z0
 const hint = `valid example: dnslink=/${dnslinkNamespace}/3ACpC9Umme41zlWUgMQh1fw0sNwgWwyfDDhRQ9Sppz9hjQ`;
 
 server.get("/dnslink/:name", async (req, res) => {
-  const success = (skylink) => res.set("Skynet-Skylink", skylink).send(skylink);
-  const failure = (message) => res.status(400).set("Dnslink-Error", message).send(message);
+  const success = (skylink) => res.send(skylink);
+  const failure = (message) => res.status(400).send(message);
```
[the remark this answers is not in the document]
kwypchlo: The only reason I had them in header is because I previously used `http.request` from "socket.http" library and it would return buffered body so it wasn't easy to read from it and I had to inject that data to headers in dnslink-api. We've changed the http library now and we have full body response as expected so we don't need the same data in headers any more.

docker/nginx/conf.d/include/track-upload
kwypchlo: yeah I know :) but this syntax is valid lua

Screenshot 2021-08-24 at 13 25 59

docker/nginx/conf.d/include/track-upload
kwypchlo: unfortunately `httpc:request_uri` doesn't have access to resolved aliases, I suspect that it operates in some kind of forked process or sth so passing aliases like `sia` or `account` doesn't work there

docker/nginx/conf.d/include/track-registry
kwypchlo: I used naming convention from the official documentation https://github.com/ledgetech/lua-resty-http#single-shot-request - it was easier to manage copy-pasting code :)

docker/nginx/conf.d/include/ssl-settings
kwypchlo: this whole file is a copy & paste from the generator - link is on top of the file https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&hsts=false&ocsp=false&guideline=5.6

I think it's beneficial to keep it unchanged as much as possible since we might want to periodically update this file based on the generator once we update nginx to support new tls protocols or ssl ciphers etc

We're not generating a new file on each server/build - we copied https://ssl-config.mozilla.org/ffdhe2048.txt once to our codebase and we're including it in Dockerfile

docker/nginx/nginx.conf
kwypchlo: yeah I think so, good catch

docker/caddy/Dockerfile
```diff
@@ -1,8 +1,18 @@
...
 
 # available dns resolvers: https://github.com/caddy-dns
 RUN xcaddy build --with github.com/caddy-dns/route53
 
-FROM caddy:2.4.1
+FROM caddy:2.4.3
 
 COPY --from=caddy-builder /usr/bin/caddy /usr/bin/caddy
+
+# bash required for mo executable to work
```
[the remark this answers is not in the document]
kwypchlo: added a comment here https://github.com/SkynetLabs/skynet-webportal/pull/1075/commits/53395dcea94ee5c139b30caa6dcdd25329d2f383

docker/nginx/conf.d/include/track-upload
ro-tex: Oh boy... :D

docker/nginx/conf.d/include/track-upload
ro-tex: Oh, good to know. It might be nice to leave a small comment explaining which service we're targeting? Definitely not a necessity, though.

docker/nginx/conf.d/include/ssl-settings
ro-tex: I see. Well, if that's their recommended style then that's fine. Let's use their recommendation.

docker/nginx/conf.d/include/track-registry
ro-tex: I see. Well, if that's their recommended style then that's fine. Let's use their recommendation.

docker/nginx/conf.d/include/ssl-settings
ro-tex: 1. That makes sense, let's keep it as-is.

2. Oh, right, this is the `dhparam` which is in its own file. What I mixed it up with was  this part of `docker/nginx/Dockerfile`:
```
    openssl req -new -newkey rsa:2048 -days 3650 -nodes -x509 \
    -subj '/CN=local-certificate' \
    -keyout /etc/ssl/local-certificate.key \
    -out /etc/ssl/local-certificate.crt
```

ro-tex (approved):

docker/nginx/conf.d/include/track-upload
kwypchlo: > Oh, good to know. It might be nice to leave a small comment explaining which service we're targeting? Definitely not a necessity, though.

added in https://github.com/SkynetLabs/skynet-webportal/pull/1075/commits/b339aeccc96478119470b6dd3f6d9093d40a253e

docker/nginx/conf.d/include/location-skylink
```diff
@@ -0,0 +1,98 @@
...
+
+    -- 10.10.10.10 points to accounts service (alias not available when using resty-http)
+    local res, err = httpc:request_uri("http://10.10.10.70:3000/user/limits", {
+        headers = { ["User-Agent"] = "Sia-Agent", ["Cookie"] = ngx.header["Cookie"] }
+    })
+    
+    -- fail gracefully in case /user/limits failed
+    if err or res and res.status ~= ngx.HTTP_OK then
+        ngx.log(ngx.ERR, err or res.body)
+        ngx.var.limit_rate = 2621440
```
peterjan, line 67: why is this exactly? how much is this, might warrant a comment?

docker/nginx/conf.d/include/location-skynet-registry
```diff
@@ -7,14 +7,23 @@ limit_req zone=registry_access_by_ip_throttled burst=200 nodelay;
...
 
 access_by_lua_block {
     -- this block runs only when accounts are enabled
     if os.getenv("ACCOUNTS_ENABLED") ~= "true" then return end
 
-    local res = ngx.location.capture("/accounts/user/limits", { copy_all_vars = true })
-    if res.status == ngx.HTTP_OK then
+    local httpc = require("resty.http").new()
+
+    -- 10.10.10.10 points to accounts service (alias not available when using resty-http)
```
peterjan, line 18: is this correct? should be `10.10.10.70`?

docker/nginx/conf.d/include/location-skynet-registry
```diff
@@ -7,14 +7,23 @@ limit_req zone=registry_access_by_ip_throttled burst=200 nodelay;
...
+    local res, err = httpc:request_uri("http://10.10.10.70:3000/user/limits", {
+        headers = { ["User-Agent"] = "Sia-Agent", ["Cookie"] = ngx.header["Cookie"] }
+    })
+    
+    -- fail gracefully in case /user/limits failed
+    if err or res and res.status ~= ngx.HTTP_OK then
```
peterjan, line 24: I assume Nginx handles this correctly? Combining `or` and `and` without parentheses?

docker/nginx/conf.d/include/location-skynet-registry
```diff
@@ -7,14 +7,23 @@ limit_req zone=registry_access_by_ip_throttled burst=200 nodelay;
...
+        ngx.log(ngx.ERR, err or res.body)
+    elseif res and res.status == ngx.HTTP_OK then
```
peterjan, line 26: same here

docker/nginx/conf.d/server/server.api
```diff
@@ -0,0 +1,283 @@
...
+
+    # increase request timeouts
+    proxy_read_timeout 600;
+    proxy_send_timeout 600;
+
+    proxy_request_buffering off; # stream uploaded files through the proxy as it comes in
+    proxy_set_header Expect $http_expect;
+    proxy_set_header User-Agent: Sia-Agent;
+
+    # access_by_lua_block {
```
peterjan, line 116: commented out block?

docker/nginx/conf.d/include/location-skynet-registry
kwypchlo:
```suggestion
    -- 10.10.10.70 points to accounts service (alias not available when using resty-http)
```

yeah copy & paste mistake

docker/nginx/conf.d/dhparam.pem
```diff
@@ -0,0 +1,8 @@
+-----BEGIN DH PARAMETERS-----
```
peterjan, line 1: should this be public?

docker/nginx/conf.d/include/location-skylink
```diff
@@ -0,0 +1,98 @@
...
+include /etc/nginx/conf.d/include/proxy-cache-downloads;
+include /etc/nginx/conf.d/include/track-download;
+
+# redirect purge calls to separate location
+error_page 462 = @purge;
+if ($request_method = PURGE) {
+    return 462;
+}
+
+limit_conn downloads_by_ip 100; # ddos protection: max 100 downloads at a time
```
peterjan, line 12: this is a limit applied cross all account types?

docker/nginx/conf.d/include/location-skylink
```diff
@@ -0,0 +1,98 @@
...
+        local res, err = httpc:request_uri("http://10.10.10.10:9980/skynet/resolve/" .. ngx.var.skylink_v2, {
+            headers = { ["User-Agent"] = "Sia-Agent" }
+        })
+
+        -- print error and exit with 500 or exit with response if status is not 200
+        if err or res and res.status ~= ngx.HTTP_OK then
+            ngx.status = err and ngx.HTTP_INTERNAL_SERVER_ERROR or res.status
+            ngx.header["content-type"] = "text/plain"
+            ngx.say(err or res.body)
+            return ngx.exit(err and ngx.HTTP_INTERNAL_SERVER_ERROR or res.status)
```
peterjan, line 47: not sure why we don't return res.status here if it's set before defaulting to 500 if an err was defined?

docker/nginx/conf.d/include/location-skylink
```diff
@@ -0,0 +1,98 @@
...
+        local json = require('cjson')
+        local resolve = json.decode(res.body)
+        ngx.var.skylink_v1 = resolve.skylink
+        ngx.var.skynet_proof = res.headers["Skynet-Proof"]
+    end
+
+    -- this block runs only when accounts are enabled
+    if os.getenv("ACCOUNTS_ENABLED") ~= "true" then return end
+
+    -- 10.10.10.10 points to accounts service (alias not available when using resty-http)
```
peterjan, line 59: typo

docker/nginx/conf.d/include/ssl-settings
```diff
@@ -0,0 +1,13 @@
+# https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&hsts=false&ocsp=false&guideline=5.6
+
+ssl_session_timeout 1d;
```
peterjan, line 3: seems high

kwypchlo: that's what mozilla suggests in https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&hsts=false&ocsp=false&guideline=5.6 so I'm just hoping they know a bit more about certificates than I do :)

docker/nginx/conf.d/include/location-skylink
kwypchlo:
```suggestion
    -- 10.10.10.70 points to accounts service (alias not available when using resty-http)
```

docker/nginx/conf.d/include/location-skynet-registry
kwypchlo: yeah, I'll add parentheses just for clarity though here and in similar places

docker/nginx/conf.d/include/location-skynet-registry
kwypchlo: yeah this is regular condition with one `and` - works fine

docker/nginx/conf.d/server/server.api
kwypchlo: this is a groundwork for the upload limits, they don't work out of the box but I wanted to keep the part that actually fetches the limit values from accounts not to have it handy once we go back to it

docker/nginx/conf.d/dhparam.pem
kwypchlo: yeah this is ok to be public https://ssl-config.mozilla.org/ffdhe2048.txt

docker/nginx/conf.d/include/location-skylink
kwypchlo: yes, it's not a new one though, been there for a while

docker/nginx/conf.d/include/location-skylink
kwypchlo: yeah that totally makes sense

peterjan (approved):
LGTM 👍 will require thorough testing though, kinda hard to review

PR #1103, opened 2021-08-23
include skapps checks

kwypchlo:
- include checks for some common skapps
- only print failed checks in discord failures report

### DIFF
--- a/packages/health-check/src/checks/extended.js
+++ b/packages/health-check/src/checks/extended.js
@@ -1036,6 +1036,90 @@ function hnsEndpointDirectoryRedirect(done) {
   skylinkVerification(done, expected, { followRedirect: false });
 }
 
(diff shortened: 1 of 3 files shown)


packages/health-check/src/checks/extended.js
```diff
@@ -1061,7 +1145,7 @@ async function skylinkVerification(done, expected, { followRedirect = true, meth
     }
 
     // Check if the response body is valid by checking against the known hash
-    if (expected.bodyHash) {
+    if ("bodyHash" in expected) {
```
kwypchlo, line 1148: this change is so that body hash `""` would still trigger this condition

ro-tex (approved):

ChrisSchinnerl (approved):

PR #1106, opened 2021-08-27
fix account endpoint nginx subrequests

kwypchlo:
- account endpoints do not need custom User-Agent
- /user endpoint should not log error on 401 - 401 means not authenticated and we should just return false in response


peterjan (approved):
LGTM 👍

ro-tex (approved):

PR #1124, opened 2021-09-02
Add changelog generator and initial changelog

MSevey:
Add changelog generator that has been used in skyd.

Items are added as md files in the corresponding changelog directory.
Changelog is then updated with the script during releases.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,19 @@
+Version Scheme
+--------------
+Skynet Webportal uses the following versioning scheme, vX.X.X
+ - First Digit signifies a major (compatibility breaking) release
+ - Second Digit signifies a major (non compatibility breaking) release
+ - Third Digit signifies a minor or patch release
+
+Version History
+---------------
+
+Latest:
+
+## August 9th, 2021:
+### v0.1.1 
+Monthly release
+
+## March 24th, 2021:
+### v0.1.0 
+Initial versioned release
--- a/changelog/.changelogignore
+++ b/changelog/.changelogignore
@@ -1,0 +1,2 @@
+.init
+.DS_Store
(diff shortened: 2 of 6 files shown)


kwypchlo:
@MSevey can we use automated tool that builds changelog based on git history? 

I really like https://www.npmjs.com/package/auto-changelog, example: 
```
npx auto-changelog -u -t keepachangelog --release-summary --ignore-commit-pattern="dependabot"
```

https://gist.github.com/kwypchlo/b3da5672cbcfa1f7830a36a012b0754e

It will also help me keeping meaningful PR titles and descriptions :)

MSevey:
> @MSevey can we use automated tool that builds changelog based on git history?
> 
> I really like https://www.npmjs.com/package/auto-changelog, example:
> 
> ```
> npx auto-changelog -u -t keepachangelog --release-summary --ignore-commit-pattern="dependabot"
> ```
> 
> https://gist.github.com/kwypchlo/b3da5672cbcfa1f7830a36a012b0754e
> 
> It will also help me keeping meaningful PR titles and descriptions :)

So my main critique of this is that it then includes everything to the point that the changelog would become over crowded a loose its effectiveness. 

It is easy to avoid the dependabot PRs but things like `Update Authorized Keys` is not a value add this to the Changelog imo.

This script that Fil has the added benefit of us being able to nicely categorize our PRs into `Key Updates`, `Bug Fixes`, and `Other`.  And since it isn't the git history, it is easy to go back and edit or add an changelog entry after the fact. So there is more flexibility there. 

So imo Fil's script is nicer and it has been working well for Skyd for over a year now.

kwypchlo (approved):

ChrisSchinnerl (approved):

PR #1223, opened 2021-09-20
expose skylink api over external domains

kwypchlo:
In addition to dnslink, we want to allow people to alias siasky.net with their domain and expose any skylink.

let's say https://example.com sets CNAME to siasky.net - it can either
- use dnslink to point to a specific skylink
- do not set dnslink, it will be able to request any any skylink so https://example.com/3ACpC9Umme41zlWUgMQh1fw0sNwgWwyfDDhRQ9Sppz9hjQ would work as well as any other skylink

### DIFF
--- a/docker/nginx/conf.d/include/location-hns
+++ b/docker/nginx/conf.d/include/location-hns
@@ -75,7 +75,7 @@ access_by_lua_block {
     end
 }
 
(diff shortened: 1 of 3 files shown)


ro-tex (approved):
LGTM.

MSevey (changes requested):
Change log file please

PR #1261, opened 2021-10-06
Use Testing Corpus Skylink

MSevey:
# PULL REQUEST

## Overview
We created a [testing-corpus](https://github.com/SkynetLabs/testing-corpus) repo as a source of all skapps and skylinks we want to test. That repo is set up to publish the list to skynet [here](https://040e2npojpl9tiahghgls0d13bvacn5qo9jodruda8lcp3l4q3h1ikg.siasky.net/).

This PR aims to use this list in the health checks instead of duplicating the information in the webportal repo.

Verified via deployment to dev1.

A workflow was added that downloads the latest testdata from the the testing-corpus skylink and stores in in `testdata.json`. This allows for the health checks to load the test data from disk and avoid a download. If there are changes in the testdata, the workflow will update the testdata.json file and commit the changes. If there are no changes then the workflow doesn't commit any changes. The workflow is set to run hourly.

## Checklist
 - [x] Changelog file created


packages/health-check/src/checks/extended.js
```diff
@@ -5,906 +5,21 @@ const { isEqual } = require("lodash");
...
-      "skynet-skylink": "3ACpC9Umme41zlWUgMQh1fw0sNwgWwyfDDhRQ9Sppz9hjQ",
-      "content-disposition": 'inline; filename="bitcoin.pdf"',
-      "content-type": "application/pdf",
-    },
+// corpusCheck runs through all the tests hosted at the resolver skylink created
+// by the testing-corpus repo https://github.com/SkynetLabs/testing-corpus
+async function corpusCheck(done) {
+  // Query the resolver skylink that the testing corpus is stored at
+  const query = `https://040e2npojpl9tiahghgls0d13bvacn5qo9jodruda8lcp3l4q3h1ikg.siasky.net/`;
+  const json = await got(query).json();
```
kwypchlo, line 13: we should not rely on a skylink to provide test data because if this fetch fails we will not run the full test suite - we should store that data as json inside of the repo

kwypchlo (changes requested):

MSevey:
Confirmed latest commit works on dev1

ro-tex (approved):

packages/health-check/src/updatetestdata.js
```diff
@@ -0,0 +1,17 @@
+// Imports
+const got = require("got");
+
+// getTestData will download the test data uploaded by the testing-corpus repo
+// for the webportal health-checks to use.
+async function getTestData() {
+  try {
+    const query = `https://040e2npojpl9tiahghgls0d13bvacn5qo9jodruda8lcp3l4q3h1ikg.siasky.net/`;
```
kwypchlo, line 8: why would we want to keep the test data in the skylink? just add the test corpus repo as a dev dependency to health-check package and import the json file

kwypchlo (changes requested):

packages/health-check/src/updatetestdata.js
MSevey: because it is cooler to use skynet 😛

PR #1325, opened 2021-10-30
add customer_email attribute to stripe session request

leotuna:
add customer_email attribute to stripe session request so people do not have to write it twice

# PULL REQUEST

## Checklist
 - [ ] Changelog file created


kwypchlo:
Thank you for your contribution! Unfortunately we cannot pass customer_email there because we're getting error
> You may only specify one of these parameters: customer, customer_email.

We create customer before processing the checkout so we can assign the customer id. We could assign the email to a customer but this would mean that the customer wouldn't be able to change that email in checkout process and I believe it is common use case to have separate emails for billing (see disabled email field).

Screenshot 2021-11-13 at 01 25 57

Unfortunately the best option is to leave this as is for the time being until Stripe allows more flexibility on their checkout api.

PR #1349, opened 2021-11-05
implement base32 decoder

kwypchlo:
# PULL REQUEST

## Overview

We want to ensure that we're always pass one version of skylinks (base64 / base32) because we need one version for cache key and access log.

- parse skylink on every request to ensure that we're always passing base64 skylink
- add requested skylink to access log

### DIFF
--- a/docker/nginx/libs/skynet/skylink.lua
+++ b/docker/nginx/libs/skynet/skylink.lua
@@ -1,58 +1,13 @@
 local _M = {}
 
-local function str_split(str, size)
-    local result = {}
-    for i=1, #str, size do
-        table.insert(result, str:sub(i, i + size - 1))
-    end
-    return result
-end
-
-local function dec2bin(num)
-    local result = ""
-    repeat
-        local halved = num / 2
-        local int, frac = math.modf(halved)
-        num = int
-        result = math.ceil(frac) .. result
-    until num == 0
-    return result
-end
-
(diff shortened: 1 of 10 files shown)


docker/nginx/libs/skynet/skylink.lua
```diff
@@ -0,0 +1,61 @@
...
+    return string.sub(concatenated, 1, 34)
+end
+
+-- parse any skylink and return base64 version
+function _M.parse(skylink)
+    if string.len(skylink) == 55 then
+        local base64 = require("ngx.base64")
+        local decoded = decodeBase32Skylink(skylink)
+
+        return base64.encode_base64url(decoded)
```
ro-tex, line 55: Base64 skylinks do not have padding, so we need to strip the `=`s from the end.
Also, base64 uses `/` and `+` which we cannot use in skylinks, so we replace them with `_`.

kwypchlo: encode_base64url is the version of encoder without padding and url safe character set (hence encode_base64url name :D )

ro-tex: Awesome :)

ro-tex (approved):

peterjan (approved):

ro-tex (approved):

PR #1375, opened 2021-11-11
introduce hard upload limits

kwypchlo:


ro-tex (approved):
LGTM.

Next time we should branch this off `portal-latest` and make a hotfix deployment. This time we have almost nothing on `main` between `portal-latest` and now but should get in the mindset of using the safer approach.

ChrisSchinnerl (approved):

PR #1377, opened 2021-11-13
upgrade accounts to new api

kwypchlo:


docker/accounts/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM golang:1.16.7
-LABEL maintainer="NebulousLabs <devs@nebulous.tech>"
+LABEL maintainer="NebulousLabs <devs@siasky.net>"
```
MSevey, line 2:
```suggestion
LABEL maintainer="SkynetLabs <devs@siasky.net>"
```

packages/dashboard/src/services/accountsApi.js
```diff
@@ -0,0 +1,5 @@
+import ky from "ky";
+
+const prefix = process.env.NEXT_PUBLIC_SKYNET_DASHBOARD_URL ?? "";
```
MSevey, line 3: Is this a new `.env` field that we need to add?

MSevey (approved):

packages/dashboard/src/services/accountsApi.js
kwypchlo: not on production, only used locally to replace localhost with actual account.siasky.net or so

packages/dashboard/src/services/useAccountsApi.js
```diff
@@ -1,15 +1,21 @@
 import useSWR from "swr";
+import { useRouter } from "next/router";
 import { StatusCodes } from "http-status-codes";
 
-const fetcher = (url) =>
-  fetch(url).then((res) => {
+const prefix = process.env.NEXT_PUBLIC_SKYNET_DASHBOARD_URL ?? "";
```
ro-tex, line 5: This env var used to be defined in `packages/dashboard/.env` which is being deleted. Is it defined anywhere else?

packages/dashboard/src/pages/user/settings.js
```diff
@@ -0,0 +1,85 @@
...
+  },
+];
+
+const emailValidationSchema = Yup.object().shape({
+  email: Yup.string().required("Email is required").email("This email is invalid"),
+  confirmEmail: Yup.string().oneOf([Yup.ref("email"), null], "Emails must match"),
+});
+
+const passwordValidationSchema = Yup.object().shape({
+  password: Yup.string().required("Password is required").min(6, "Password has to be at least 6 characters long"),
```
ro-tex, line 44: nit: I thought we decided to not impose any limits on the password aside from it not being empty?

packages/dashboard/src/pages/user/settings.js
```diff
@@ -0,0 +1,85 @@
...
+  confirmPassword: Yup.string().oneOf([Yup.ref("password"), null], "Passwords must match"),
```
ro-tex, line 45: Am I reading this correctly, as the `confirmPassword` being allowed to be `null`?
I see other instances of this but I won't duplicate the same comment there.

packages/dashboard/src/pages/payments.js
```diff
@@ -19,22 +16,23 @@ const ActiveBadge = () => {
...
-  const [user, stats, stripe] = await Promise.all([
-    api.get("user").json(),
-    api.get("user/stats").json(),
-    api.get("stripe/prices").json(),
-  ]);
-  const plans = [config.tiers.starter, ...stripe].sort((a, b) => a.tier - b.tier);
+export default function Payments() {
+  const { data: user } = useAccountsApi("user");
+  const { data: stats } = useAccountsApi("user/stats");
+  const { data: prices } = useAccountsApi("stripe/prices");
```
ro-tex, line 22: Q: My React is very rusty at best and non-existent at worst, so I have a question about promises here:
Are these executed sequentially or in parallel? `GET /user/stats` is not exactly fast - there's a lot of work to be done for it. If we want to improve the UX we should be doing these in parallel.

packages/dashboard/src/pages/auth/registration.js
```diff
@@ -1,88 +1,66 @@
...
+  email: Yup.string().required("Email is required").email("This email is invalid"),
+  password: Yup.string()
+    .required("Password is required")
+    .min(6, "Password has to be at least 6 characters long")
+    .test("levenshtein", "This password is too similar to your email", function (value) {
+      const email = this.parent.email;
+
+      // levenshtein distance higher than 5 and longest common sequence shorter than half of the password
+      return value && email && levenshtein.get(value, email) > 5 && lcs(value, email).length / value.length <= 0.5;
+    }),
```
ro-tex, line 44: Those are some password requirements! :)

ro-tex (dismissed):
LGTM.

I have some notes and questions but they are fine as F/U.

Also as F/U, it would be nice to move all those pagination magic numbers to centralized constants.

packages/dashboard/src/services/useAccountsApi.js
ro-tex: Based on https://github.com/SkynetLabs/skynet-webportal/pull/1377#discussion_r755498965 this is something only used for local development. Not a show-stopper then.

packages/dashboard/src/components/Form/SelfServiceForm.js
```diff
@@ -1,87 +1,68 @@
...
-    .map((field) => ({ ...field, ...fieldsConfig[field.name] }))
-    .sort((a, b) => (a.position < b.position ? -1 : 1));
+export default function SelfServiceForm({ fieldsConfig, onSubmit, title, validationSchema = null, button = "Submit" }) {
+  const [messages, setMessages] = React.useState([]);
+  const fields = fieldsConfig.sort((a, b) => (a.position < b.position ? -1 : 1));
   const formik = useFormik({
     initialValues: fields.reduce((acc, field) => setIn(acc, field.name, field.value ?? ""), {}),
+    validationSchema,
+    onSubmit: async (values) => {
+      if (!formik.isValid) return;
```
peterjan, line 13: `nit` I dislike these inline returns that do not adhere to the overall code style

packages/dashboard/src/pages/user/settings.js
```diff
@@ -0,0 +1,85 @@
...
+    position: 1,
+  },
+];
+
+const passwordConfig = [
+  {
+    name: "password",
+    type: "password",
+    label: "New password",
+    position: 1,
```
peterjan, line 28: are the positions correct here?

peterjan (approved):
LGTM 👍

ro-tex (approved):
LGTM

ro-tex:
All ongoing conversations are documented in https://github.com/SkynetLabs/skynet-webportal/issues/1439.  
I'll resolve them and I'll merge this.

PR #1391, opened 2021-11-24
Disable load check until we have a process of actively addressing it

MSevey:
# PULL REQUEST

## Overview
Disabling the load check notification until we have a system in place for acting on it.


ro-tex:
The load check is a nice indicator of something odd happening with the system. It's true that we don't know how to handle it in the general case but it's a red flag for us which prompts us to log into the system and look around.

I would disable any mentions that check performs but having the check in place is valuable.

DavidVorick (approved):

PR #1408, opened 2021-12-01
Abuse Reports

peterjan:
# PULL REQUEST

## Overview

This PR adds two routes to support the abuse skapp:
- `/report-abuse`
- `/abuse`

The first route redirect to the abuse skapp. 
This should be a proxy pass but that errors out, I have not been able to figure out why that is, 
I will F/U with that as a redirect works fine.

The second route exposes the `blocker` API and allows `POST`ing abuse reports.
This route is authorized using the accounts jwt.

## Example for Visual Changes

N/A

## Checklist
 - [x] Changelog file created


docker-compose.accounts.yml
```diff
@@ -41,6 +41,8 @@ services:
       - SKYNET_ACCOUNTS_LOG_LEVEL=${SKYNET_ACCOUNTS_LOG_LEVEL}
       - KRATOS_ADDR=${KRATOS_ADDR}
       - OATHKEEPER_ADDR=${OATHKEEPER_ADDR}
+    volumes:
+      - ./docker/accounts/conf:/accounts/conf
```
ChrisSchinnerl, line 45: Should be `./docker/accounts/conf:/conf` and then we set the environment var 
`- ACCOUNTS_JWKS_FILE=/conf/jwks.json`

ro-tex: What is the motivation behind this? `ACCOUNTS_JWKS_FILE` currently defaults to `/accounts/conf/jwks.json`. We can play around with that any way we like, of course, but I would prefer to use the default value, so if we happen to not set it on some server it would still work as expected.

ro-tex (approved):
LGTM.

This works, as tested on dev1. I'm happy to merge it.

ChrisSchinnerl (approved):

PR #1410, opened 2021-12-03
Add serverload endpoint

MSevey:
# PULL REQUEST

## Overview
This PR adds a serverload endpoint that servers a `serverload.json` file containing the CPU usage and free disk space with a timestamp.

I created a `systemd` service file that we can use to run this script since it needs to be outside of docker. If we agree with that approach I will follow up with the ansible scripts to verify the service is running. 

## Example for Visual Changes

Screen Shot 2021-12-02 at 10 09 28 PM

## Checklist
 - [x] Changelog file created


setup-scripts/serverload.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Create logfile if it doesn't exist
+if [[ ! -e $logfile ]]; then
+    echo "init" > $logfile 
+fi
+
+# Write the output in an infinite loop.
+while true; do
+    # CPU usage
+    cpu=$(grep 'cpu ' /proc/stat | awk '{usage=($2+$4)*100/($2+$4+$5)} END {print usage "%"}')
```
peterjan, line 26: I'm not sure whether this command is what we want. I think you got this from [stacko](https://stackoverflow.com/questions/9229333/how-to-get-overall-cpu-usage-e-g-57-on-linux) right?

There is a note on that thread that says:
> NOTE: this is not the current CPU usage, but the overall CPU usage in all the cores since the system bootup. This could be very different from the current CPU usage. To get the current value top (or similar tool) must be used.

It goes on to suggest:
> awk '{u=$2+$4; t=$2+$4+$5; if (NR==1){u1=u; t1=t;} else print ($2+$4-u1) * 100 / (t-t1) "%"; }' \
<(grep 'cpu ' /proc/stat) <(sleep 1;grep 'cpu ' /proc/stat)

The current command I believe will give the average CPU load since the server was started, 
so it would be a (hugely) lagging indicator. Perhaps we should use something like `top` for this instead.

MSevey: I just got the command from David. Easy enough to update to whatever we want.

peterjan (approved):

firyx (dismissed):

ro-tex (approved):

setup-scripts/serverload.sh
peterjan:
```suggestion
    cpu=$(echo $[100-$(vmstat 1 2|tail -1|awk '{print $15}')]) 
```

peterjan: This is an updated command from David. We can indeed change it whenever so I decided to apply that one and resolve your comment so we can merge this today.

ro-tex (approved):

PR #1430, opened 2021-12-08
Fil/disk usage root dirs

firyx:
# PULL REQUEST

## Overview

Add disku usage dumps for `root` and `skynet-webportal` directory on health checker disabling portal for low disk space. 

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/setup-scripts/disk-usage-dump.sh
+++ b/setup-scripts/disk-usage-dump.sh
@@ -26,7 +26,7 @@ dump () {
     # Root dirs
     echo
     echo "Root dirs:"
-    docker run -v /:/host-root alpine:3.15.0 sh -c 'du -hs /host-root/*'
+    docker run -v /:/host-root alpine:3.15.0 sh -c 'du -hs /host-root/*' | sed 's#/host-root##'
 
     # Home dirs
     echo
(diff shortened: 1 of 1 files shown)


setup-scripts/disk-usage-dump.sh
```diff
@@ -23,11 +23,21 @@ dump () {
     echo
     df -h /home/user
 
+    # Root dirs
+    echo
+    echo "Root dirs:"
+    docker run -v /:/host-root alpine:3.15.0 sh -c 'du -hs /host-root/*' | sed 's#/host-root##'
```
ro-tex, line 29: I am not sure we need this part. Historically, this has never been a problem and it would add more pressure on the disk.

ro-tex (approved):
LGTM.

I'm not convinced that we need to check `/` but I'm not strictly opposed to it.

setup-scripts/disk-usage-dump.sh
firyx: On `us-ny1` we had full disk while `/user/home` took only 70GB this night.

peterjan (approved):

PR #1443, opened 2021-12-13
Hot reload nginx after pruning cache

firyx:
# PULL REQUEST

## Overview

Some portals (`us-ny-1`, `us-ny-2`) are repeatedly disabled by health checker due to low disk space.
Disk usage dump shows difference (> 300GB) between `df`(includes deleted files) and `du` (doesn't include deleted files), which could be caused by deleted but still open files by nginx (and elastic).

This PR will try to compete deletion of files by nginx hot reload and releasing the files.

## Checklist
 - [x] Changelog file created

### DIFF
--- a/changelog/items/other/nginx-prune-hot-reload.md
+++ b/changelog/items/other/nginx-prune-hot-reload.md
@@ -1,0 +1,1 @@
+- Hot reload Nginx after pruning cache files.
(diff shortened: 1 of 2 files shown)


peterjan (approved):
LGTM, would be interesting to see how much of the allocated disk space is from ES and how much from Nginx

MSevey (approved):

PR #1452, opened 2021-12-13
Abuse Mail Scanner

peterjan:
# PULL REQUEST

## Overview

This PR introduces a new module called `abuse`. The module scans the email IMAP server for incoming emails, parses them and automatically blocks abusive content and generates a report for it automatically.

## Example for Visual Changes
N/A

## Checklist
 - [ ] Changelog file created

## Issues Closed
N/A


MSevey:
Should this be run on every portal? will that create race conditions?
What is the production plan?

dc
```diff
@@ -13,6 +13,11 @@ for i in $(seq 1 ${#PORTAL_MODULES}); do
     COMPOSE_FILES+=" -f docker-compose.mongodb.yml -f docker-compose.accounts.yml"
   fi
 
+  # abuse module - alias "c"
```
ro-tex, line 16: nit: We're running out of convenient letters, it seems. How about `u` for `abuse`?

docker-compose.abuse.yml
```diff
@@ -0,0 +1,34 @@
...
+    container_name: abuse
+    restart: unless-stopped
+    logging: *default-logging
+    env_file:
+      - .env
+    environment:
+      - ABUSE_LOG_LEVEL=${ABUSE_LOG_LEVEL}
+      - ABUSE_MAILBOX=${ABUSE_MAILBOX}
+      - ABUSE_SPONSOR=${ABUSE_SPONSOR}
+      - BLOCKER_API_URL=${BLOCKER_API_URL}
```
ro-tex, line 23: `malware-scanner` already introduces env vars for this:
```
      - BLOCKER_IP=${BLOCKER_IP:-10.10.10.110}
      - BLOCKER_PORT=${BLOCKER_PORT:-4000}
```
Let's reuse them here.

docker-compose.abuse.yml
```diff
@@ -0,0 +1,34 @@
...
+      - BLOCKER_AUTH_HEADER=${BLOCKER_AUTH_HEADER}
+      - EMAIL_SERVER=${EMAIL_SERVER}
+      - EMAIL_USERNAME=${EMAIL_USERNAME}
+      - EMAIL_PASSWORD=${EMAIL_PASSWORD}
```
ro-tex, line 27: We already have `ACCOUNTS_EMAIL_URI` which does the same thing plus some extras. The downside being that you need stuff like 
```go
matchPattern = regexp.MustCompile("smtps://(?P<user>.*):(?P<password>.*)@(?P<server>.*):(?P<port>\\d*)(/\\??skip_ssl_verify=(?P<skip_ssl_verify>\\w*))?")
```
to parse it.

docker-compose.abuse.yml
```diff
@@ -0,0 +1,34 @@
...
+      - MONGO_CONNECTIONSTRING=${MONGO_CONNECTIONSTRING}
```
ro-tex, line 28: For this we also have env vars already - all of the `SKYNET_DB_*` ones.

While the `ACCOUNTS_EMAIL_URI` is only being used by `accounts` and we can change it if you really want, the DB ones are being used by a bunch of services by now (accounts, skyd, blocker, malware-scanner) and I don't think we'll change it.

ro-tex, line 28: For this we also have env vars already - all of the `SKYNET_DB_*` ones.

While the `ACCOUNTS_EMAIL_URI` is only being used by `accounts` and we can change it if you really want, the DB ones are being used by a bunch of services by now (accounts, skyd, blocker, malware-scanner) and I don't think we'll change it.

docker-compose.abuse.yml
```diff
@@ -0,0 +1,34 @@
...
+    networks:
+      shared:
+        ipv4_address: 10.10.10.110
```
ro-tex, line 31: 10.10.10.110 is used by `blocker`
```suggestion
        ipv4_address: 10.10.10.120
```

ro-tex (changes requested):

docker-compose.abuse.yml
ro-tex: While reading the other PR about this service, I realised that this EMAIL_SERVER thing can be different from the email we want to use for sending emails. Let's rename these to something more descriptive, such as `ABUSE_EMAIL_SERVER` and so on, so we know right away (when reading the .env file) that this is the mailbox we want to watch for incoming abuse notifications and not the email account used for sending emails. Those can easily be different.

docker-compose.abuse.yml
peterjan: updated

docker-compose.abuse.yml
peterjan: updated, but we should in my opinion migrate to use `MONGO_CONNECTIONSTRING` for two reasons:
1. it's more descriptive than `SKYNET_DB_`
2. with mongo it's just easier to use the connection string because you can plug that into a mongo client directly

docker-compose.abuse.yml
peterjan: updated

docker-compose.abuse.yml
peterjan: I want to follow up with this one.

docker-compose.abuse.yml
peterjan: https://github.com/SkynetLabs/skynet-webportal/pull/1452#discussion_r772316565

docker-compose.abuse.yml
```diff
@@ -0,0 +1,38 @@
...
+    restart: unless-stopped
+    logging: *default-logging
+    env_file:
+      - .env
+    environment:
+      - ABUSE_LOG_LEVEL=${ABUSE_LOG_LEVEL}
+      - ABUSE_MAILBOX=${ABUSE_MAILBOX}
+      - ABUSE_SPONSOR=${ABUSE_SPONSOR}
+      - BLOCKER_IP=${BLOCKER_IP}
+      - BLOCKER_PORT=${BLOCKER_PORT}
```
ro-tex, line 24: `abuse` is now using `BLOCKER_HOST` and not `BLOCKER_IP`. This is not a real problem, though, because you have passed the `.env` file, so all values from it will be available to the service, whether they are deifned here or not.

We can also add those values here already, like this:
```
      - SKYNET_DB_HOST=${SKYNET_DB_HOST:-mongo}
      - SKYNET_DB_PORT=${SKYNET_DB_PORT:-27017}
```

ro-tex (approved):
Left one comment that can be F/U. Otherwise LGTM.

ChrisSchinnerl (dismissed):

docker-compose.abuse.yml
ro-tex: From offline:
We have different services which use different `readPreference`s, so this might be problematic.
We will discuss and figure out how to move forward.

ChrisSchinnerl (approved):

ro-tex (approved):

PR #1458, opened 2021-12-15
add accounts maintenance mode redirect

kwypchlo:
add accounts maintenance mode that is enabled by setting `ACCOUNTS_MAINTENANCE` env variable to `true`

dashboard redirects with 307 to /maintenance on every page request when maintenance mode is on

### DIFF
--- a/packages/dashboard/next.config.js
+++ b/packages/dashboard/next.config.js
@@ -1,0 +1,15 @@
+module.exports = {
+  async redirects() {
+    const redirects = [];
+
+    if (process.env.ACCOUNTS_MAINTENANCE === "true") {
+      redirects.push({
+        source: "/((?!maintenance$).*)",
+        destination: "/maintenance",
+        permanent: false,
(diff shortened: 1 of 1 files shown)


ro-tex (approved):

peterjan (approved):
LGTM 👍

packages/dashboard/next.config.js
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  async redirects() {
+    const redirects = [];
+
+    if (process.env.ACCOUNTS_MAINTENANCE === "true") {
+      redirects.push({
+        source: "/((?!maintenance$).*)",
```
peterjan, line 7: `nit` could've dropped a line here explaining the neg. lookahead

kwypchlo: I'll remove this once we're done with migration, let's just leave it as is for now

PR #1478, opened 2021-12-16
no delay blocklisting

kwypchlo:
- initialise 25m nginx shared blocklist cache
- on first skylink request, blocklist cache will be filled with current blocklist
- on every skylink request, skylink will be compared against local blocklist cache and reject if it's matched
- added new local nginx server (port) on port 8000 that is limited to only local network traffic (deny from all but local network) that exposes POST /skynet/skylink endpoint that in addition to blocklisting the skylink in skyd, updates nginx local blocklist cache so it's always up to date (works with adding new skylinks and removing existing skylinks)

deployed on dev3.siasky.dev

**test case (_test with regular v1 skylink and resolver skylink_):**
- upload new file and copy a skylink
- ensure the skylink **IS** accessible by opening it in browser
- block the skylink on that particular server
```bash
curl --data "{\"add\":[\"YOUR_SKYLINK_HERE\"]}" "http://10.10.10.30:8000/skynet/blocklist"
```
- ensure the skylink **IS NOT** accessible by opening it in browser - you should see "Unavailable For Legal Reasons"
- restart the server with `./dc down && ./dc up -d` and make sure the skylink is still blocked
- unblock the skylink
```bash
curl --data "{\"remove\":[\"YOUR_SKYLINK_HERE\"]}" "http://10.10.10.30:8000/skynet/blocklist"
```
- ensure the skylink **IS** accessible by opening it in browser
- restart the server with `./dc down && ./dc up -d` and make sure the skylink is still accessible

### DIFF
--- a/docker/nginx/conf.d/server.local.conf
+++ b/docker/nginx/conf.d/server.local.conf
@@ -4,11 +4,7 @@ server {
     listen [::]:8000;
 
     # secure traffic by limiting to only local networks
-    allow 10.0.0.0/8;
-    allow 127.0.0.1/32;
-    allow 172.16.0.0/12;
-    allow 192.168.0.0/16;
-    deny all;
-    
+    include /etc/nginx/conf.d/include/local-network-only;
+
     include /etc/nginx/conf.d/server/server.local;
 }
--- a/docker/nginx/conf.d/server.local.conf
+++ b/docker/nginx/conf.d/server.local.conf
@@ -1,0 +1,14 @@
+server {
+    # local server - no not expose this port externally
+    listen 8000;
+    listen [::]:8000;
+
+    # secure traffic by limiting to only local networks
+    allow 10.0.0.0/8;
+    allow 127.0.0.1/32;
+    allow 172.16.0.0/12;
+    allow 192.168.0.0/16;
+    deny all;
+    
+    include /etc/nginx/conf.d/server/server.local;
+}
--- a/docker/nginx/conf.d/server.local.conf
+++ b/docker/nginx/conf.d/server.local.conf
@@ -1,5 +1,5 @@
 server {
-    # local server - no not expose this port externally
+    # local server - do not expose this port externally
     listen 8000;
     listen [::]:8000;
 
--- a/docker/nginx/conf.d/server/server.local
+++ b/docker/nginx/conf.d/server/server.local
@@ -1,0 +1,34 @@
+include /etc/nginx/conf.d/include/init-optional-variables;
+
+location /skynet/blocklist {
+    content_by_lua_block {        
+        local httpc = require("resty.http").new()
+
+        ngx.req.read_body() -- ensure the post body data is read before using get_body_data
+
+        -- proxy blocklist update request
+        -- 10.10.10.10 points to sia service (alias not available when using resty-http)
+        local res, err = httpc:request_uri("http://10.10.10.10:9980/skynet/blocklist", {
+            method = "POST",
+            body = ngx.req.get_body_data(),
+            headers = {
+                ["Content-Type"] = "application/x-www-form-urlencoded",
+                ["Authorization"] = require("skynet.utils").authorization_header(),
+                ["User-Agent"] = "Sia-Agent",
+            }
+        })
+
+        -- print error and exit with 500 or exit with response if status is not 204
+        if err or (res and res.status ~= ngx.HTTP_NO_CONTENT) then
+            ngx.status = (err and ngx.HTTP_INTERNAL_SERVER_ERROR) or res.status
+            ngx.header["content-type"] = "text/plain"
+            ngx.say(err or res.body)
+            return ngx.exit(ngx.status)
+        end
+
+        require("skynet.blocklist").reload()
+
+        ngx.status = ngx.HTTP_NO_CONTENT
+        return ngx.exit(ngx.status)
+    }
+}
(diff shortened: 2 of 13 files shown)


docker/nginx/libs/skynet/skylink.lua
```diff
@@ -13,4 +14,27 @@ function _M.parse(skylink)
...
+    local base64Skylink = _M.parse(skylink)
+
+    -- decode skylink from base64 encoding
+    local rawSkylink = basexx.from_url64(base64Skylink)
+
+    -- drop first two bytes and leave just merkle root
+    local rawMerkleRoot = string.sub(rawSkylink, 3)
+
+    -- parse with blake2b with key length of 32
+    local blake2bHashed = hasher.blake2b(rawMerkleRoot, 32)
```
ro-tex, line 29: The stuff we do in lua now...

docker/nginx/conf.d/server.local.conf
```diff
@@ -0,0 +1,14 @@
+server {
+    # local server - no not expose this port externally
```
ro-tex, line 2: nit: typo. If there is nothing blocking feel free to ignore this, it's not important.
```suggestion
    # local server - do not expose this port externally
```

ChrisSchinnerl, line 2: *do not?

docker/nginx/conf.d/server.local.conf
```diff
@@ -0,0 +1,14 @@
...
+    listen 8000;
+    listen [::]:8000;
+
+    # secure traffic by limiting to only local networks
+    allow 10.0.0.0/8;
+    allow 127.0.0.1/32;
```
ChrisSchinnerl, line 8: Shouldn't this only be 127.0.0.1/32 and 10.0.0.0/8? 

We might want to keep this to only the essential ones to prevent other servers in the datacenter from stumbling across this.

docker/nginx/conf.d/server/server.local
```diff
@@ -0,0 +1,34 @@
...
+            body = ngx.req.get_body_data(),
+            headers = {
+                ["Content-Type"] = "application/x-www-form-urlencoded",
+                ["Authorization"] = require("skynet.utils").authorization_header(),
+                ["User-Agent"] = "Sia-Agent",
+            }
+        })
+
+        -- print error and exit with 500 or exit with response if status is not 204
+        if err or (res and res.status ~= ngx.HTTP_NO_CONTENT) then
```
ChrisSchinnerl, line 22: Isn't this overkill? Can't we just catch status codes which are not 2xx ? 
Because theoretically all 2xx codes indicate success anyway.

docker/nginx/conf.d/server/server.local
```diff
@@ -0,0 +1,34 @@
...
+            ngx.status = (err and ngx.HTTP_INTERNAL_SERVER_ERROR) or res.status
+            ngx.header["content-type"] = "text/plain"
+            ngx.say(err or res.body)
+            return ngx.exit(ngx.status)
+        end
+
+        require("skynet.blocklist").reload()
```
ro-tex, line 29: Are we reloading the entire blocklist each time we block a new skylink? Can we instead have an `add` method that does `x.shared.blocklist:set(hash, true)` and `ngx.shared.blocklist:flush_expired()`?

peterjan (dismissed):

docker/nginx/conf.d/server/server.local
kwypchlo: Reloading whole blocklist is really fast and we do that very rarely - tops couple of times a day. Otherwise I would need to handle resolver skylinks in the script since blocklist automatically resolves them and blocks underlying skylink v1. The improvement I would be looking for is a native skyd endpoint that would accept a skylink and return whether it's blocked or not - discussed on discord.

docker/nginx/libs/skynet/skylink.lua
kwypchlo: makes my shed a tear every now and then

docker/nginx/conf.d/server.local.conf
kwypchlo: I think even this would do

```
allow 127.0.0.1/32;  # localhost
allow 10.10.10.0/24; # docker network
deny all;
```

docker/nginx/nginx.conf
```diff
@@ -72,11 +72,16 @@ http {
     # proxy cache definition
     proxy_cache_path /data/nginx/cache levels=1:2 keys_zone=skynet:10m max_size=50g min_free=100g inactive=48h use_temp_path=off;
 
+    # create a shared blocklist dictionary
+    lua_shared_dict blocklist 25m;
```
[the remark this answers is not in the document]
kwypchlo: sure 👍

docker/nginx/libs/skynet/blocklist.lua
```diff
@@ -0,0 +1,54 @@
...
+end
+
+function _M.is_blocked(skylink)
+    -- make sure that blocklist has been preloaded
+    if not ngx.shared.blocklist:get("__init") then _M.reload() end
+
+    -- hash skylink before comparing it with blocklist
+    local hash = require("skynet.skylink").hash(skylink)
+
+    -- we need to use get stale because we're using expiring when updating blocklist
```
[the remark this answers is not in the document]
kwypchlo: improved docs

ChrisSchinnerl (approved):

docker/nginx/conf.d/server/server.local
ro-tex: I thought we already had the V1 skylink.

Anyway, if this only happens rarely then we can leave it.

ro-tex (approved):

peterjan (approved):

PR #1492, opened 2021-12-20
Update skynet-js to v4.0.21-beta

m-cat:
# PULL REQUEST

## Overview

Updates to the latest `skynet-js` version:

1. Improved performance of large file uploads
2. Fixed data race in get/setJSON methods. (**note:** now, `getJSON` must always be called first for existing data.)

### DIFF
--- a/packages/health-check/package.json
+++ b/packages/health-check/package.json
@@ -13,7 +13,7 @@
     "http-status-codes": "^2.1.2",
     "lodash": "^4.17.21",
     "lowdb": "^1.0.0",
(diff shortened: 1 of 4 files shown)


dghelm:
2 concerns when using the auto-deploy above:

1. have we re-introduced the "pause at 40MB" UI thing, but now just pausing for 2 chunks at the same time, so "pause at 80MB"?
2. Have we accounted for other SDK changes?:
image

kwypchlo:
v4.0.21 has been rolled back

PR #1497, opened 2021-12-23
do not prune nginx cache on airtable blocklist script

kwypchlo:
with https://github.com/SkynetLabs/skynet-webportal/pull/1478 merged we can now drop nginx cache pruning from airtable plugin

### DIFF
--- a/scripts/blocklist-skylink.sh
+++ b/scripts/blocklist-skylink.sh
@@ -1,18 +1,15 @@
 #! /usr/bin/env bash
 
-# This script adds a skylink to the sia blocklist and removes the skylink from
-# nginx cache. The script should be run locally on each skynet webportal
-# server. The automatic script that is used to continuously sync an Airtable
-# sheet list with the blocklist on the web portals is
-# /setup-scripts/blocklist-airtable.py 
+# This script is for manual skylink blocking. It accepts either a single 
+# skylink or a file containing list of skylinks. The script is intented
+# for manual use and it should be run locally on each skynet webportal server. 
+# The automatic script that is used to continuously sync an Airtable sheet 
+# list with the blocklist on the web portals is /setup-scripts/blocklist-airtable.py 
 
 set -e # exit on first error
 
-# Number of skylinks to block within one batch
-BATCH_SIZE=1000
-
 if [ -z "$1" ]; then
-    echo "Please provide either a skylink or file with skylinks separated by new lines" && exit 1
+    echo "Please provide either a skylink or a file with skylinks separated by new lines" && exit 1
 fi
 
 #########################################################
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (approved):

setup-scripts/blocklist-airtable.py
```diff
@@ -122,70 +122,36 @@ async def block_skylinks_from_airtable():
...
-            "Siad blocklist endpoint responded with code "
+            "Airtable blocklist request responded with code "
             + status_code
             + ": "
             + response_text
         )
         return await send_msg(message, force_notify=False)
 
-    # Remove from NGINX cache
-    # NOTE:
```
MSevey, line 152: Should do the same for the `blocklist-skylink.sh` script

MSevey (changes requested):

setup-scripts/blocklist-airtable.py
kwypchlo: > Should do the same for the `blocklist-skylink.sh` script

Changes in https://github.com/SkynetLabs/skynet-webportal/pull/1497/commits/dac4d57b1236a6250ce21ae782a3cd63a26dd643

scripts/blocklist-skylink.sh
```diff
@@ -37,45 +34,18 @@ else
...
-    # If (batch len == batch size) or (we have last batch):
-    if (( ${#batch_skylinks[@]} == $BATCH_SIZE || $i == $skylinks_len - 1 )); then
-        echo "--------------------------------------------"
-
-        # Add to Sia blocklist
-        echo "Blocking batch skylinks in skyd..."
-        skylinks_space_separated="$(IFS=' '; echo "${batch_skylinks[*]}")"
-        docker exec sia siac skynet blocklist add $skylinks_space_separated
+# get local nginx ip adress
+nginx_ip=$(docker inspect -f '{{range.NetworkSettings.Networks}}{{.IPAddress}}{{end}}' nginx)
```
ro-tex, line 38: nit: A quick check that we managed to get an IP (i.e. nginx is running) might be nice. Not blocking, though.

ro-tex (approved):

scripts/blocklist-skylink.sh
```diff
@@ -37,45 +34,18 @@ else
...
+for skylink in "${skylinks[@]}"; do
+    printf "Blocking ${skylink} ... "
+    status_code=$(curl --write-out '%{http_code}' --silent --output /dev/null --data "{\"add\":[\"$skylink\"]}" "http://${nginx_ip}:8000/skynet/blocklist")
 
-        # Clear batch
-        batch_skylinks=()
-
-        echo "--------------------------------------------"
+    # print blocklist response status code
+    if [ $status_code = "204" ]; then
```
peterjan, line 46: `nit` use `==`? it's a synonym right, but still 🤷‍♂️

peterjan (approved):
LGTM

scripts/blocklist-skylink.sh
kwypchlo: If nginx wouldn't be running we would get `Error: No such object: nginx` from the docker command

scripts/blocklist-skylink.sh
kwypchlo: afaik `==` is `bash` specific alias and doesn't port to `ash` or some other shells so I think it's safer to use single quote just in case we want to run it on a different shell (?)

MSevey (approved):

PR #1498, opened 2021-12-23
introduce portal mode for authenticated users only

kwypchlo:
Introduced new env variable `ACCOUNTS_LIMIT_ACCESS` that defaults to `authenticated` and sets portal mode to accept only authenticated users traffic - all downloads, uploads and registry reads and writes. All unauthenticated requests to those methods will return 401 Unauthorized error code. Frontend has been prepared to detect whether portal is in auth only mode and adjusts uploader component accordingly (fades it and shows log in / sign up links instead).

**Migration note:** since `ACCOUNTS_LIMIT_ACCESS` defaults to `authenticated`, any existing public portal with accounts enabled will need to set it explicitly to `false` or empty string in `.env` file // cc @firyx


docker/nginx/conf.d/include/init-optional-variables
```diff
@@ -3,10 +3,16 @@
...
 # set only if server has been access through SERVER_DOMAIN
-set $server_alias '';
+set $server_alias "";
 
 # expose skylink variable so we can use it in access log
-set $skylink '';
+set $skylink "";
+
+# cached account limits (json string) - applies only if accounts are enabled
+set $account_limits "";
```
peterjan, line 15: default to anon?

peterjan (approved):

docker/nginx/conf.d/include/init-optional-variables
kwypchlo: I need it to default to empty string so we know we have to pull the limits from accounts for the first time.

peterjan (dismissed):

ro-tex (approved):

ChrisSchinnerl (approved):

PR #1518, opened 2022-01-03
prevent files larger than 1GB from being stored in cache

kwypchlo:
Since we removed slicing module, we need to make sure too large files don't end up in cache and overflow our free disk space causing services to halt. I've set the threshold to 1GB to start off.


ChrisSchinnerl (approved):

ChrisSchinnerl:
Manually confirmed using curl that >1GB files never get cached and <1GB files eventually do.

ro-tex (approved):

PR #1519, opened 2022-01-04
use portal modules env to discover services in nginx

kwypchlo:
- use PORTAL_MODULES env variable to discover services in nginx
- replace ACCOUNTS_ENABLED usage in nginx with PORTAL_MODULES
- run scanner /scan only if scanner module is enabled (check with PORTAL_MODULES)

### DIFF
--- a/docker/nginx/conf.d/include/track-download
+++ b/docker/nginx/conf.d/include/track-download
@@ -1,48 +1,51 @@
 # register the download in accounts service (cookies should contain jwt)
 log_by_lua_block {
     -- this block runs only when accounts are enabled
-    if os.getenv("ACCOUNTS_ENABLED") ~= "true" then return end
+    if os.getenv("PORTAL_MODULES"):match("a") then
+        local function track(premature, skylink, status, body_bytes_sent, jwt)
+            if premature then return end
 
-    local function track(premature, skylink, status, body_bytes_sent, jwt)
-        if premature then return end
+            local httpc = require("resty.http").new()
+            local query = table.concat({ "status=" .. status, "bytes=" .. body_bytes_sent }, "&")
 
-        local httpc = require("resty.http").new()
-        local query = table.concat({ "status=" .. status, "bytes=" .. body_bytes_sent }, "&")
+            -- 10.10.10.70 points to accounts service (alias not available when using resty-http)
+            local res, err = httpc:request_uri("http://10.10.10.70:3000/track/download/" .. skylink .. "?" .. query, {
+                method = "POST",
+                headers = { ["Cookie"] = "skynet-jwt=" .. jwt },
+            })
 
-        -- 10.10.10.70 points to accounts service (alias not available when using resty-http)
-        local res, err = httpc:request_uri("http://10.10.10.70:3000/track/download/" .. skylink .. "?" .. query, {
-            method = "POST",
-            headers = { ["Cookie"] = "skynet-jwt=" .. jwt },
-        })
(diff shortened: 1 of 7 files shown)


docker/nginx/conf.d/include/track-registry
```diff
@@ -1,27 +1,27 @@
...
     -- this block runs only when accounts are enabled
-    if os.getenv("ACCOUNTS_ENABLED") ~= "true" then return end
+    if os.getenv("PORTAL_MODULES"):match("a") then
+        local function track(premature, request_method, jwt)
+            if premature then return end
 
-    local function track(premature, request_method, jwt)
-        if premature then return end
+            local httpc = require("resty.http").new()
+            local method = request_method == "GET" and "read" or "write"
```
ro-tex, line 9: This is sooo weird. It merits a comment, at the very least. I've re-read it 5-6 times and I'm still not 100% how it works.

I guess it returns `read` when the HTTP method is `GET` but I don't know why. Maybe it's just how lua works. A short comment would be nice.

If we're storing the `read`/`write` operation in the variable then it should be called something else - `operation`/`action`/etc. If we call it `method` then I'll immediately expect it to be `GET`/`POST`/etc down in the code.

ro-tex (commented):
Looks good overall. It would be nice to address the comments before we merge.

docker/nginx/conf.d/include/track-download
```diff
@@ -1,48 +1,51 @@
...
 
-    local function scan(premature, skylink, jwt)
-        if premature then return end
+    -- this block runs only when scanner module is enabled
+    if os.getenv("PORTAL_MODULES"):match("s") then
+        local function scan(premature, skylink)
+            if premature then return end
 
-        local httpc = require("resty.http").new()
+            local httpc = require("resty.http").new()
```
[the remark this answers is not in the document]
kwypchlo: we shouldn't reuse - it's good for one request, otherwise it gets a bit complicated and I don't think it's worth it

docker/nginx/conf.d/include/track-registry
kwypchlo: so accounts api exposes tracking api for registry
- /track/registry/read
- /track/registry/write

Depending whether the actual client request was GET (registry read) or POST (registry write) we need to report to different api endpoint (above). So I'm creating `method` variable that contains either "read" or "write" string and concatenate it with the "/track/registry/<method_here>" to report to correct endpoint.

kwypchlo: And if you are asking about the syntax, it's a way to write a ternary in lua since lua has no native ternary operator

`local foo = true and "bar" or "baz"` - foo contains "bar"
`local foo = false and "bar" or "baz"` - foo contains "baz"

docker/nginx/conf.d/include/track-download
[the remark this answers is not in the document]
ro-tex: OK, I'll trust you on that. :)

docker/nginx/conf.d/include/track-registry
ro-tex: I got the part with the different endpoint we want to hit. I guess my two points are:
* lua's ternary operator is a little unusual and we could use a comment explaining what's going on
* the name of the variable is misleading and we should rename it

ro-tex (approved):

ChrisSchinnerl (approved):

Issue #1521, opened 2022-01-07
Add support for Matomo Tag Manager


dghelm:
In addition to our Matomo Analytics, we should add support for the [Matomo Tag Manager](https://matomo.org/docs/tag-manager/) to better identify events and behaviors on the homepage.

This is a piece of the SEO & marketing initiative we're working on.

Issue #1495, opened 2021-12-21
Add site-map generation to the Gatsby site


dghelm:
For SEO purposes (and helping distinguish Skynet homepage from files hosted on Skynet) we should generate a sitemap with our Gatsby build.

https://www.gatsbyjs.com/plugins/gatsby-plugin-sitemap/

The main complication will be making sure we include accurate "lastmod" values or Google will penalize us.

kwypchlo:
merged in https://github.com/SkynetLabs/skynet-webportal/pull/1544

PR #1534, opened 2022-01-11
Add support for Matomo Tag Manager

dghelm:
Changes:
- Matomo Gatsby Plugin now loads piwik.js from static files in build instead of externally
- Matomo Tag Manager code added in each page's `<head>`. This will load an external .js file for further SEO/analytics using our self-hosted Matomo instance.


dghelm:
This addresses #1521

dghelm:
This commit addresses #1495 and adds sitemap generation. (Will edit to link to XML file.)

kwypchlo (approved):
we can merge it but we need to figure out how to correctly make it configurable for community portals

peterjan (approved):

PR #1546, opened 2022-01-13
support auth only portal mode in health-check module

kwypchlo:
When portal is set to allow only authenticated users traffic with `ACCOUNTS_LIMIT_ACCESS` set to `authenticated`, health checks will fail since there was no concept of authentication there before, we were health-checking only with anonymous user on public endpoints. We need to send authenticated requests via health-check so we either need a registered user. 

Easiest thing to do at the moment is just programmatically create a user, sign in and use skynet-jwt cookie in health-check requests. We don't want to create a new user every time we run health-checks so we need to provide username and password with `ACCOUNTS_TEST_USER_EMAIL` and `ACCOUNTS_TEST_USER_PASSWORD` (can be either shared on all servers or each server can have it's own). We don't need to manually create the user, for the first time the credentials are used (user doesn't exist), health-check will create the user automatically.

Note: when portal is set to only accept authenticated traffic, health-check module will fail if both `ACCOUNTS_TEST_USER_EMAIL` and `ACCOUNTS_TEST_USER_PASSWORD` are provided, just to make sure that portal would not come up fully until these requirements are met.

Example credentials (email prefixed with healthcheck followed by random number and some random but secure password):
```bash
ACCOUNTS_TEST_USER_EMAIL=healthcheck+812739871246@siasky.net
ACCOUNTS_TEST_USER_PASSWORD=xBE4DVeuv83NtS9LeY3cU9oFtF19a0GTyZEvFUpcjsn3SiH
```

### DIFF
--- a/packages/health-check/src/utils.js
+++ b/packages/health-check/src/utils.js
@@ -1,3 +1,5 @@
+const got = require("got");
+
 /**
  * Get the time between start and now in milliseconds
  */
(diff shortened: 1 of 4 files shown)


packages/health-check/src/utils.js
```diff
@@ -39,9 +41,60 @@ function ensureValidJSON(object) {
...
+
+  const email = process.env.ACCOUNTS_TEST_USER_EMAIL;
+  const password = process.env.ACCOUNTS_TEST_USER_PASSWORD;
+
+  if (!email) throw new Error("ACCOUNTS_TEST_USER_EMAIL cannot be empty");
+  if (!password) throw new Error("ACCOUNTS_TEST_USER_PASSWORD cannot be empty");
+
+  async function authenticate() {
+    try {
+      // authenticate with given test credentials
```
peterjan, line 64: test credentials?

packages/health-check/src/utils.js
```diff
@@ -39,9 +41,60 @@ function ensureValidJSON(object) {
...
+      const response = await got.post(`${process.env.SKYNET_DASHBOARD_URL}/api/login`, {
+        json: { email, password },
+      });
+
+      // extract set-cookie from successful authentication request
+      const cookies = response.headers["set-cookie"];
+
+      // find the skynet-jwt cookie
+      const jwtcookie = cookies.find((cookie) => cookie.startsWith("skynet-jwt"));
```
peterjan, line 73: isn't the result of `.find` `null` or `undefined`, so won't you run into trouble calling `.match` on it later?

kwypchlo: well if we run into that then we're gonna throw an error and have to debug why we didn't get a cookie on successful authentication request :)

packages/health-check/src/utils.js
kwypchlo:
```suggestion
      // authenticate with given test user credentials
```

packages/health-check/src/utils.js
peterjan: lol, ok fine I'lll approve it but then you should
```
if (!jwtcookie) {
    throw new Error("'skynet-jwt' cookie not found');
}
```
}

peterjan (approved):

peterjan (approved):

peterjan (approved):

ChrisSchinnerl (approved):

PR #1569, opened 2022-01-19
enable cors on /logout endpoint

kwypchlo:
enable cors on /logout accounts endpoint to match /login and /register behaviour


m-cat (approved):

ro-tex (changes requested):
Based on the discussion on Discord this is considered to be a security issue and we therefore don't want to merge it.

I'm "requesting changes" in order to block a potential merge until we figure out what we want to do.

PR #1573, opened 2022-01-20
health checks depends on blocker when it is available

kwypchlo:
- env variables `BLOCKER_HOST` and `BLOCKER_PORT` should not be read from .env file because they are not configurable, making them configurable at this point just brings confusion because changing the ip in .env file will not reflect on the service itself and will just cause services depending on it to fail to start
- we should not use `blocker` alias in health-check since we would have to add `depends_on` to health-check but we want health-check to start regardless blocker service works on not


MSevey (approved):

peterjan (approved):
might make sense to document in an issue what would be the ideal approach

PR #1574, opened 2022-01-21
unique SERVER_UID

peterjan:
# PULL REQUEST

## Overview

This PR extends the `dc` command with some scripting that ensures the `.env` file has a valid `SERVER_UID`.

## Example for Visual Changes
N/A

## Checklist

## Issues Closed
N/A

## Testing

I would appreciate if someone would test this. I tested it myself with:
- no .env file
- a .env file with an invalid `SERVER_UID`
- a .env file with an empty `SERVER_UID`
- a .env file with a valid `SERVER_UID`

### DIFF
--- a/dc
+++ b/dc
@@ -10,6 +10,24 @@ if [ -f .env ]; then
   IFS=$'\n'
   for x in $(grep -v '^#.*' .env); do export $x; done
   IFS=$OLD_IFS
+else
+  # ensure a .env is present
+  touch .env
+fi
+
+if [ -z $(cat .env | grep -i '^SERVER_UID=') ];
+then
+  # add a SERVER_UID to .env if none was found
+  SERVER_UID=$(openssl rand -hex 8)
+  printf "# randomly generated 16 character string that must be unique across servers\nSERVER_UID=$SERVER_UID" >> .env
+  printf "\nIMPORTANT NOTE: added SERVER_UID '$SERVER_UID' to .env file\n\n"
+elif [ -z $(cat .env | grep -E -i '^SERVER_UID=[0-9A-F]{16}') ];
+then
+  # ensure the SERVER_UID is a valid 16 char hex string if the current value was
+  # either empty or malformed
+  SERVER_UID=$(openssl rand -hex 8)
+  sed -i '' "/^SERVER_UID=/s/=.*/=$SERVER_UID/" .env
+  printf "\nIMPORTANT NOTE: added SERVER_UID '$SERVER_UID' to .env file\n\n"
 fi
 
 # include base docker compose file
--- a/dc
+++ b/dc
@@ -12,13 +12,13 @@ if [ -f .env ]; then
   IFS=$OLD_IFS
 fi
 
-if [ -z $(cat .env | grep -i '^SERVER_UID=') ];
+if [ -z $(grep -i '^SERVER_UID=' .env) ];
 then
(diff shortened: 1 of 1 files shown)


dc
```diff
@@ -10,6 +10,24 @@ if [ -f .env ]; then
   IFS=$'\n'
   for x in $(grep -v '^#.*' .env); do export $x; done
   IFS=$OLD_IFS
+else
+  # ensure a .env is present
+  touch .env
+fi
+
+if [ -z $(cat .env | grep -i '^SERVER_UID=') ];
```
peterjan, line 18: this takes care of a `.env` file which does not mention a `SERVER_UID`

dc
```diff
@@ -10,6 +10,24 @@ if [ -f .env ]; then
...
+then
+  # add a SERVER_UID to .env if none was found
+  SERVER_UID=$(openssl rand -hex 8)
+  printf "# randomly generated 16 character string that must be unique across servers\nSERVER_UID=$SERVER_UID" >> .env
+  printf "\nIMPORTANT NOTE: added SERVER_UID '$SERVER_UID' to .env file\n\n"
+elif [ -z $(cat .env | grep -E -i '^SERVER_UID=[0-9A-F]{16}') ];
```
peterjan, line 24: this takes care of `.env` files where the `SERVER_UID` is malformed, or not set

peterjan: I think it's important that we fill the value if it's `SERVER_UID=` for instance,
still on the fence whether we should assert it's a 16 char lowercase hex string
might make sense to replace it with `grep -E -i '^SERVER_UID=.+'`

peterjan: I personally think this is fine though

dc
ChrisSchinnerl: This could be `$(grep -i '^SERVER_UID=' .env)` I think. `grep` already supports specifying a file.

dc
ChrisSchinnerl: I don't think we need to assert anything about it. A user should be able to set it to whatever they want. e.g. `eu-ger-1`. It's their responsibility to make it unique if they do it but it should be possible.
We also need this upgrade to not touch any env vars that we set to non-compliant strings.

dc
```diff
@@ -10,6 +10,24 @@ if [ -f .env ]; then
...
+else
+  # ensure a .env is present
+  touch .env
+fi
+
+if [ -z $(cat .env | grep -i '^SERVER_UID=') ];
+then
+  # add a SERVER_UID to .env if none was found
+  SERVER_UID=$(openssl rand -hex 8)
+  printf "# randomly generated 16 character string that must be unique across servers\nSERVER_UID=$SERVER_UID" >> .env
```
kwypchlo, line 22:
```suggestion
  printf "\n# randomly generated 16 character string that must be unique across servers\nSERVER_UID=$SERVER_UID" >> .env
```

otherwise it appends without new line

dc
```diff
@@ -10,6 +10,24 @@ if [ -f .env ]; then
   IFS=$'\n'
   for x in $(grep -v '^#.*' .env); do export $x; done
   IFS=$OLD_IFS
+else
+  # ensure a .env is present
+  touch .env
+fi
```
kwypchlo, line 16:
```suggestion
fi
```

You don't need that, if .env doesn't exist the code that does `>> .env` will create the file

kwypchlo (changes requested):

PR #1575, opened 2022-01-21
add option to limit access to subscription users only

kwypchlo:
support for `ACCOUNTS_LIMIT_ACCESS` set to "subscription" to limit access to users with active subscription only


kwypchlo (commented):
PJ here, I approve this PR

ChrisSchinnerl (approved):

PR #1627, opened 2022-01-28
fix website build on m1 mac

kwypchlo:
build on mac m1 fails with gatsby 4.6.x - related issue https://github.com/gatsbyjs/gatsby/issues/34605

- downgraded gatsby to 4.5.4
- removed unused gatsby-image plugin
- added avatar as optional file referenced by relative path


dghelm:
For anyone else wondering: `gatsby-image` is the plugin deprecated and replaced with `gatsby-plugin-image`

dghelm (approved):

peterjan (approved):

PR #1643, opened 2022-01-31
Simple fix for Login URL formatting of double slash

dghelm:
# PULL REQUEST

## Overview
Currently, URLs for login and signup are formatted to look like: `https://account.skynetpro.net//auth/login`, using a double slash.

image

I'm unsure if there is code elsewhere to ensure that accountsUrl ends with a `/`, but I doubt the react app is the place for that logic anyways.

### DIFF
--- a/packages/website/src/components/Uploader/Uploader.js
+++ b/packages/website/src/components/Uploader/Uploader.js
@@ -26,7 +26,7 @@ const RegistrationLink = () => {
 
   return (
     <Link
-      href={`${accountsUrl}/auth/registration`}
+      href={`${accountsUrl}auth/registration`}
       className="uppercase underline-primary hover:text-primary transition-colors duration-200"
     >
       Sign up
@@ -39,7 +39,7 @@ const LogInLink = () => {
 
   return (
     <Link
-      href={`${accountsUrl}/auth/login`}
+      href={`${accountsUrl}auth/login`}
       className="uppercase underline-primary hover:text-primary transition-colors duration-200"
     >
       Log in
--- a/packages/website/src/components/Uploader/Uploader.js
+++ b/packages/website/src/components/Uploader/Uploader.js
@@ -22,11 +22,11 @@ const getRootDirectory = (file) => {
 };
 
 const RegistrationLink = () => {
-  const accountsUrl = useAccountsUrl();
+  const createAccountsUrl = useAccountsUrl();
 
   return (
     <Link
-      href={`${accountsUrl}auth/registration`}
+      href={createAccountsUrl("auth/registration")}
       className="uppercase underline-primary hover:text-primary transition-colors duration-200"
     >
       Sign up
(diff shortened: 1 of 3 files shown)


m-cat (approved):
> I'm unsure if there is code elsewhere to ensure that accountsUrl ends with a /, but I doubt the react app is the place for that logic anyways.

Sounds like something we should put in the (non-existent) `skynet-utils` repo I've been dreaming about.

dghelm:
Ah, nice Karol. I will copy that trick in the future.

packages/website/src/components/Uploader/Uploader.js
```diff
@@ -22,11 +22,11 @@ const getRootDirectory = (file) => {
 };
 
 const RegistrationLink = () => {
-  const accountsUrl = useAccountsUrl();
+  const createAccountsUrl = useAccountsUrl();
```
m-cat, line 25: Why is this function being assigned to a const? And why are we overriding the existing `createAccountsUrl` function? I don't know how React works but this looks really weird.

packages/website/src/services/useAccountsUrl.js
```diff
@@ -3,6 +3,7 @@ import skynetClient from "./skynetClient";
 
 export default function useAccountsUrl() {
   const [url, setUrl] = React.useState("");
+  const createAccountsUrl = React.useCallback((path = "") => url && new URL(path, url).toString(), [url]);
```
m-cat, line 6: Can we unit test this?

packages/website/src/components/Uploader/Uploader.js
m-cat: I see that `createAccountsUrl` is actually returned from the `useAccountsUrl` function. It still looks weird. I also don't understand why we are assigning this to a variable.

m-cat (changes requested):
- I don't really understand the new changes introduced by @kwypchlo .
- Would be awesome if the changes were unit-tested. If there is no unit-testing suite we need to set it up.

- f/u: If we don't have it yet we should enact a baseline test coverage. We don't need full coverage but new changes should not decrease coverage.

packages/website/src/services/useAccountsUrl.js
m-cat: I see that there were a couple additional commits to fix bugs which is why we need automated tests for functions like this. It's easy to get logic wrong or miss edge cases.

packages/website/src/components/Uploader/Uploader.js
kwypchlo: `useAccountsUrl` is a custom react hook https://reactjs.org/docs/hooks-intro.html and that's why it has to be assigned inside of the component

packages/website/src/services/useAccountsUrl.js
kwypchlo: I skipped on unit testing components due to time constraints, maybe now that we have one more frontend dev on board we will be able to do better in this regard.

peterjan (approved):
LGTM 👍

m-cat (approved):

PR #1660, opened 2022-02-04
reenable end to end uploader test

kwypchlo:
bring back commented out upload file test for company website

### DIFF
--- a/packages/website/cypress/integration/intergration.spec.js
+++ b/packages/website/cypress/integration/intergration.spec.js
@@ -9,27 +9,26 @@ context("Skynet website", () => {
     cy.contains("Decentralized Internet");
   });
 
-  // it("should be able to upload a file", () => {
-  //   cy.intercept("POST", "/skynet/skyfile").as("upload");
+  it("should be able to upload a file", () => {
+    cy.intercept("POST", "/skynet/skyfile").as("upload");
 
-  //   const fileName = "check.json";
+    const fileName = "example.json";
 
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (dismissed):

kwypchlo:
At last it passed 👍 @meeh0w @ChrisSchinnerl can I get another look now?

.github/workflows/deploy-website.yml
```diff
@@ -36,8 +36,10 @@ jobs:
           working-directory: packages/website
           install: false
           record: true
-          start: yarn serve
-          wait-on: "http://127.0.0.1:9000"
+          start: yarn develop
```
meeh0w, line 39: Why was the switch needed?

meeh0w (approved):

.github/workflows/deploy-website.yml
kwypchlo: because when we build the production bundle without GATSBY_API_URL we rely on the current url to be a portal - when we serve from localhost we error out the upload because localhost is not a portal 

`yarn develop` sets GATSBY_API_URL to siasky.net so running from localhost works, other solution would have to be a have a separate job that for building prod artifact, separate job to build test artifact, and separate job for uploading the prod artifact when tests succeed - seemed a bit much tbh :)

meeh0w: Makes sense, thanks for the explanation 🍻

peterjan (approved):

PR #1668, opened 2022-02-07
make ip check in health checks more reliable

kwypchlo:
- whatsmyip script will now exit with code 1 on error preventing docker container from starting and will keep restarting until we get the ip correctly - this will ensure that once the container is started, we already have the ip
- ensure that whatever response we got from the external ip checking tool, we get an actual ip (compare with ip regex)
- when serverip env var is available (defined in CMD of Dockerfile) - instead of querying service every time for current ip, use the ip provided in the env variable - this will ensure that any outage in akamai service will not affect our health checks
- when no external server ip is available (so serverip not there, maybe running the script manually outside of docker and also akamai is not available - super edge case) - do not fail health checks, just skip the ip check on that execution
- use `/proc/1/fd/1` - `/dev/stdout` never really worked heh

### DIFF
--- a/packages/health-check/Dockerfile
+++ b/packages/health-check/Dockerfile
@@ -23,14 +23,15 @@ COPY bin bin
 EXPOSE 3100
 ENV NODE_ENV production
 
-# 1. start dnsmasq in the background with:
+# 1. get public server ip and save it in /etc/environment (passed to cron tasks as env variable)
+# 2. start dnsmasq in the background with:
 #    - alias PORTAL_DOMAIN with current server ip so it overrides potential load balancer request
 #    - default docker nameserver 127.0.0.11 for any other request
-# 2. replace docker nameserver with dnsmasq nameserver in /etc/resolv.conf
-# 3. start crond in the background to schedule periodic health checks
-# 4. start the health-check api service
+# 3. replace docker nameserver with dnsmasq nameserver in /etc/resolv.conf
+# 4. start crond in the background to schedule periodic health checks
+# 5. start the health-check api service
 CMD [ "sh", "-c", \
-      "serverip=$(node src/whatismyip.js) && \
+      "export serverip=$(node src/whatismyip.js) && \
        echo \"export serverip=${serverip}\" >> /etc/environment && \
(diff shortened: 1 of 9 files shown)


MSevey (approved):

packages/health-check/Dockerfile
```diff
@@ -30,9 +30,10 @@ ENV NODE_ENV production
 # 3. start crond in the background to schedule periodic health checks
 # 4. start the health-check api service
 CMD [ "sh", "-c", \
-      "serverip=$(node src/whatismyip.js) ; \
-       dnsmasq --no-resolv --log-facility=/var/log/dnsmasq.log --address=/$PORTAL_DOMAIN/$serverip --server=127.0.0.11 ; \
-       echo \"$(sed 's/127.0.0.11/127.0.0.1/' /etc/resolv.conf)\" > /etc/resolv.conf ; \
-       crond ; \
+      "serverip=$(node src/whatismyip.js) && \
+       echo export serverip=${serverip} >> /etc/environment && \
```
peterjan, line 34: `nit` single quotes couldn't hurt here

peterjan (approved):

ChrisSchinnerl (dismissed):

packages/health-check/Dockerfile
kwypchlo: it's double quotes that we want here, otherwise `${serverip}` would not get substituted for the actual ip

```suggestion
       echo "export serverip=${serverip}" >> /etc/environment && \
```

packages/health-check/src/utils.js
```diff
@@ -1,4 +1,7 @@
-const got = require("got");
+const ipCheckService = "whatismyip.akamai.com";
+const ipRegex = new RegExp(
```
meeh0w, line 2: Should we care about IPv6 at this point?

meeh0w (approved):
Left a small question, lgtm otherwise 👍

packages/health-check/src/utils.js
kwypchlo: I don't think so, if we'll have to support ipv6 then it will mean some major refactor anyway I believe

MSevey (approved):

PR #1670, opened 2022-02-08
require server uid env var for blocker service

kwypchlo:
When `SERVER_UID` is not set or empty, docker will throw an error when trying to start the stack.

```bash
➜  skynet-webportal git:(master) ✗ ./dc up
invalid interpolation format for services.sia.environment.[]: "required variable SERVER_UID is missing a value: cannot be empty". You may need to escape any $ with another $.
```

### DIFF
--- a/docker-compose.blocker.yml
+++ b/docker-compose.blocker.yml
@@ -21,6 +21,8 @@ services:
     logging: *default-logging
     env_file:
       - .env
+    environment:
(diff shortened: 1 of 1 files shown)


MSevey (approved):

kwypchlo:
Closing this in favour of services dealing with their env variable requirements inside of their logic because:
- docker does not allow conditionally required env variables
- adding/removing env variable requirement in service will result in a need to add/remove it in docker compose file

PR #1671, opened 2022-02-08
expose internal authenticate function on health-check cli

kwypchlo:
Adds function that takes env variables and authenticates user returning auth cookie.

**note:** this function will be removed when api keys are implemented

Usage example:

`docker exec health-check cli __authenticate`

On success: prints cookie in a format `skynet-jwt=cookie-value-here;`
On error: exits with code 1

Required env variables:
- `PORTAL_MODULES` - must include "a"
- `ACCOUNTS_TEST_USER_EMAIL` - test user email
- `ACCOUNTS_TEST_USER_PASSWORD` - test user password
- `PORTAL_DOMAIN` - current portal domain

### DIFF
--- a/packages/health-check/src/utils.js
+++ b/packages/health-check/src/utils.js
@@ -45,24 +45,35 @@ function ensureValidJSON(object) {
  * Authenticate with given credentials and return auth cookie
  * Creates new account if username does not exist
  * Only authenticates when portal is set to authenticated users only mode
+ * @param {boolean} forceAuth forcibly ensure authentication with test credentials
  */
-function getAuthCookie() {
+function getAuthCookie(forceAuth = false) {
   // cache auth promise so only one actual request will be made
   if (getAuthCookie.cache) return getAuthCookie.cache;
 
-  // do not authenticate if it is not necessary
-  if (!["authenticated", "subscription"].includes(process.env.ACCOUNTS_LIMIT_ACCESS)) return {};
+  // accounts disabled, do not try to authenticate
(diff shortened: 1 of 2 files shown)


packages/health-check/src/utils.js
```diff
@@ -45,24 +45,35 @@ function ensureValidJSON(object) {
...
+  if (!forceAuth && !["authenticated", "subscription"].includes(process.env.ACCOUNTS_LIMIT_ACCESS)) return "";
+
+  const portalDomain = process.env.PORTAL_DOMAIN;
   const email = process.env.ACCOUNTS_TEST_USER_EMAIL;
   const password = process.env.ACCOUNTS_TEST_USER_PASSWORD;
 
-  if (!email) throw new Error("ACCOUNTS_TEST_USER_EMAIL cannot be empty");
-  if (!password) throw new Error("ACCOUNTS_TEST_USER_PASSWORD cannot be empty");
+  try {
+    if (!portalDomain) throw new Error("PORTAL_DOMAIN cannot be empty");
```
meeh0w, line 65: No need to wrap it in `try...catch` if we raise errors only to log their messages, I guess?

How about something like this:

```js
const fetchRequiredEnvVar = (name) => {
  const value = process.env[name];

  if (!value) {
    console.log(`${name} cannot be empty`);
    process.exit(1);
  }
  
  return value;
}
```

and then use it like:
```js
const portalDomain = fetchRequiredEnvVar('PORTAL_DOMAIN');
```

Issue #718, opened 2021-05-05
Clean up environment variables and what goes into env file


kwypchlo:

kwypchlo:
I think we're in a good position right now, might close it in favour of more specific issue in the future

PR #1688, opened 2022-02-09
remove dashboard url env var usage

kwypchlo:
- replaces `SKYNET_DASHBOARD_URL` usage with `https://account.${PORTAL_DOMAIN}`
- completely removes all traces of `SKYNET_DASHBOARD_URL`

part of https://github.com/SkynetLabs/skynet-webportal/issues/718 effort

cc @firyx - we will have to remove it from ansible scripts too


meeh0w (dismissed):

packages/dashboard/src/services/accountsApi.js
```diff
@@ -1,5 +1,5 @@
 import ky from "ky";
 
-const prefix = process.env.NEXT_PUBLIC_SKYNET_DASHBOARD_URL ?? "";
+const prefix = process.env.NEXT_PUBLIC_PORTAL_DOMAIN ? `https://account.${process.env.NEXT_PUBLIC_PORTAL_DOMAIN}` : "";
```
ro-tex, line 3: I find it odd that we're fine with returning an invalid URI here but since it's not a new thing I won't dig into it.

ro-tex (approved):

meeh0w (approved):

packages/dashboard/src/services/accountsApi.js
kwypchlo: it's not going to be invalid, if it's empty, it's going to use the base uri from browser

Issue #1384, opened 2021-11-16
Subfiles with filenames containing spaces do not work


m-cat:
## Describe the bug

Subfiles with filenames containing spaces do not work for base32 skylinks.

## To Reproduce

1. Go to `https://siasky.net/skynet/metadata/AAC-knq-m00M56G39LYzke4ytgroIlBBv4Ff3er8_p1VCQ`, confirm there is a file called `file 1.txt`.
1. Go to `https://siasky.net/AAC-knq-m00M56G39LYzke4ytgroIlBBv4Ff3er8_p1VCQ/file%201.txt`, confirm the file is accessible for base64 skylinks.
1. Go to `https://000bt4jqnqdkq377k6rv9dhji7n35dgat0h50gdvg5ftrqnsvqela28.siasky.net/file%201.txt`, confirm the file is not accessible for a base32 skylink.
1. Go to `https://000bt4jqnqdkq377k6rv9dhji7n35dgat0h50gdvg5ftrqnsvqela28.siasky.net/file0.txt`, confirm that subfiles without spaces do work for base32 skylinks.

## Expected behavior

`https://000bt4jqnqdkq377k6rv9dhji7n35dgat0h50gdvg5ftrqnsvqela28.siasky.net/file%201.txt` should return the requested file.

## Screenshots

n/a

## Version Information (please complete the following information):
 
n/a

## Additional context

n/a

PR #1697, opened 2022-02-09
escape uri on subdomain skylink requests

kwypchlo:
Fixes a bug when accessing file from skylink with a filename that had escaped characters.
The issue was that we assigned `$uri` as a file path but `$uri` in nginx contains already unescaped uri string so `/file%201.txt` would become `/file 1.txt` which when passed to proxy_pass resulted in an issue from skyd.
The solution was to use `$request_uri` that in nginx is the full,  original, escaped request uri with query params and then strip it from query params since those should not be a part of our `$path` variable.
I also tried easier approach by just escaping `$uri` but it also escaped `/` so it resulted in malformed uri hence `$request_uri` approach was the only valid one.

closes https://github.com/SkynetLabs/skynet-webportal/issues/1384

### DIFF
--- a/docker/nginx/conf.d/server/server.skylink
+++ b/docker/nginx/conf.d/server/server.skylink
@@ -9,7 +9,7 @@ location / {
     set_by_lua_block $path {
         -- strip ngx.var.request_uri from query params - this is basically the same as ngx.var.uri but
         -- do not use ngx.var.uri because it will already be unescaped and we need to use escaped path
-        -- examples: escaped uri "/b%20r56+7" and unescaped uri: "/b r56 7"
+        -- examples: escaped uri "/b%20r56+7" and unescaped uri "/b r56 7"
         return string.gsub(ngx.var.request_uri, "?.*", "")
     }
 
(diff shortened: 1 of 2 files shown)


docker/nginx/conf.d/server/server.skylink
```diff
@@ -6,7 +6,12 @@ include /etc/nginx/conf.d/include/init-optional-variables;
 
 location / {
     set_by_lua_block $skylink { return string.match(ngx.var.host, "%w+") }
-    set $path $uri;
+    set_by_lua_block $path {
+        -- strip ngx.var.request_uri from query params - this is basically the same as ngx.var.uri but
+        -- do not use ngx.var.uri because it will already be unescaped and we need to use escaped path
+        -- examples: escaped uri "/b%20r56+7" and unescaped uri "/b r56 7"
+        return string.gsub(ngx.var.request_uri, "?.*", "")
```
peterjan, line 13: I trust we don't have to escape the '?' here right.

peterjan (approved):
LGTM 👍

MSevey (approved):

docker/nginx/conf.d/server/server.skylink
kwypchlo: `string.gsub(ngx.var.request_uri, "?.*", "")` will strip the uri from `?` and anything after that so:

`/hello?world=foo&bar=baz` will become `/hello`

PR #1701, opened 2022-02-11
incremental logs and daily skylinks backup

kwypchlo:
- connect to aws with aws cli
- create bucket skynet-backup-[portaldomain]
- servers will store their backups in [serveruid]-[serverdomain] directory
- back up all nginx logs (incremental, not versioned)
- back up dump of skylinks from local skyd (versioned daily)

### DIFF
--- a/scripts/backup-aws-s3.sh
+++ b/scripts/backup-aws-s3.sh
@@ -1,0 +1,43 @@
+#! /usr/bin/env bash
+
+set -e # exit on first error
+
+echo "`date +"%Y-%m-%d %H:%M"` Starting backup process"
+
+# ensure aws cli is upgraded
+pip3 install --quiet --upgrade awscli
+
+# import required environment variables from .env file
+ENV_VARIABLES=("AWS_ACCESS_KEY_ID" "AWS_SECRET_ACCESS_KEY" "PORTAL_DOMAIN" "SERVER_DOMAIN" "SERVER_UID")
+
+for ENV_VARIABLE in "${ENV_VARIABLES[@]}"; do
+    ENV_VARIABLE_VALUE=$(grep -v '^#' /home/user/skynet-webportal/.env | grep ${ENV_VARIABLE} || true)
+    if test -z "${ENV_VARIABLE_VALUE}"; then
+        if $ENV_VARIABLE ~= "SERVER_DOMAIN"; then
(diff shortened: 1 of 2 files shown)


scripts/backup-aws-s3.sh
```diff
@@ -0,0 +1,43 @@
...
+BUCKET_NAME=$(echo skynet-backup-${PORTAL_DOMAIN} | tr -cd '[[:alnum:]]')
+
+# create server prefix
+if test -z "${SERVER_DOMAIN}"; then
+    SERVER_PREFIX=$(echo ${SERVER_UID} | tr -cd '[[:alnum:]]-') # if domain name is empty use just uid
+else
+    SERVER_PREFIX=$(echo ${SERVER_UID}-${SERVER_DOMAIN} | tr -cd '[[:alnum:]]-') # use both uid and server domain if available
+fi
+
+aws s3api create-bucket --acl private --bucket ${BUCKET_NAME}
```
ro-tex, line 34: Automating bucket creation is risky. The bucket name might already be taken. I would prefer to do that manually and specify the bucket name in a config var.

The other thing here is that we might want to disable file versioning on the bucket. This will make newly uploaded files overwrite their previous versions and we will only be charged for the current usage and not for multiple copies of the same data. We might also not want this, if we want to see those versions for some reason.

MSevey: I'm not sure I'm following the difference between defining the bucket name in the env file vs here. 

The key point is not overwriting buckets right? So we just need to confirm that `create-bucket` doesn't overwrite an existing bucket?

MSevey (approved):

scripts/backup-aws-s3.sh
meeh0w: There might also be conflicts.

For example, we'd be writing data of `skynetpro.net` and `skynetpron.et` (Ethiopia TLD) to the same bucket (`skynet-backup-skynetpronet`).

Unlikely, but possible.

kwypchlo: even if there will be a conflict (unlikely), we use `SERVER_PREFIX` that contains unique `SERVER_UID` as a part of first level directory name so there will never be a case where 2 servers overwrite their backups

I am open to introducing an env variable for `BUCKET_NAME` if we get a request to do so, otherwise I'd keep it simple

PR #1703, opened 2022-02-14
adjust airtable script to new response code

kwypchlo:
- move skylinks validation from script to skyd (push all skylinks from airtable to skyd and expect "invalids" in response)
- expect response code 200 with either `{"invalids": null}` or ie `{"invalids": [{"input": "foo", "error": "[failed to parse skylink - data is malformed; [unable to decode skylink; skylink has incorrect size]]"}]}`


ro-tex (approved):

peterjan (approved):
LGTM

PR #1723, opened 2022-02-15
Allow the Skynet-API-Key header.

ro-tex:
Add a health-check command for fetching the test user's API key.

# PULL REQUEST

## Overview

This PR add the `Skynet-API-Key` header to the list of allowed headers.
It also adds a health-check command for fetching the API key of the test user, similar to the way we fetch the cookie of the test user in order to run integration tests.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/docker/nginx/conf.d/include/cors-headers
+++ b/docker/nginx/conf.d/include/cors-headers
@@ -2,4 +2,4 @@ more_set_headers 'Access-Control-Allow-Origin: $http_origin';
 more_set_headers 'Access-Control-Allow-Credentials: true';
 more_set_headers 'Access-Control-Allow-Methods: GET, POST, HEAD, OPTIONS, PUT, PATCH, DELETE';
 more_set_headers 'Access-Control-Allow-Headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,If-None-Match,Cache-Control,Content-Type,Range,X-HTTP-Method-Override,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,upload-concat,location,Skynet-API-Key';
-more_set_headers 'Access-Control-Expose-Headers: Content-Length,Content-Range,ETag,Skynet-File-Metadata,Skynet-Skylink,Skynet-Proof,Skynet-Portal-Api,Skynet-Server-Api,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,upload-concat,location,Skynet-API-Key';
+more_set_headers 'Access-Control-Expose-Headers: Content-Length,Content-Range,ETag,Skynet-File-Metadata,Skynet-Skylink,Skynet-Proof,Skynet-Portal-Api,Skynet-Server-Api,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,upload-concat,location';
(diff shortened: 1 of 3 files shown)


docker/nginx/conf.d/include/cors-headers
```diff
@@ -1,5 +1,5 @@
...
-more_set_headers 'Access-Control-Allow-Headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,If-None-Match,Cache-Control,Content-Type,Range,X-HTTP-Method-Override,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,upload-concat,location';
-more_set_headers 'Access-Control-Expose-Headers: Content-Length,Content-Range,ETag,Skynet-File-Metadata,Skynet-Skylink,Skynet-Proof,Skynet-Portal-Api,Skynet-Server-Api,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,upload-concat,location';
+more_set_headers 'Access-Control-Allow-Headers: DNT,User-Agent,X-Requested-With,If-Modified-Since,If-None-Match,Cache-Control,Content-Type,Range,X-HTTP-Method-Override,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,upload-concat,location,Skynet-API-Key';
+more_set_headers 'Access-Control-Expose-Headers: Content-Length,Content-Range,ETag,Skynet-File-Metadata,Skynet-Skylink,Skynet-Proof,Skynet-Portal-Api,Skynet-Server-Api,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,upload-concat,location,Skynet-API-Key';
```
kwypchlo, line 5: we don't want to expose Skynet-API-Key header to browser, server shouldn't ever be setting it right?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers

```suggestion
more_set_headers 'Access-Control-Expose-Headers: Content-Length,Content-Range,ETag,Skynet-File-Metadata,Skynet-Skylink,Skynet-Proof,Skynet-Portal-Api,Skynet-Server-Api,upload-offset,upload-metadata,upload-length,tus-version,tus-resumable,tus-extension,tus-max-size,upload-concat,location';
```

packages/health-check/bin/cli
```diff
@@ -16,6 +16,16 @@ require("yargs/yargs")(process.argv.slice(2))
...
+  .command(
+    "apikey",
+    "Get test user's API key",
+    () => {},
+    async () => {
+      const { getAPIKey } = require("../src/utils");
+
+      console.log(await getAPIKey());
+    }
+  )
```
kwypchlo, line 28: I don't think we need this method if we've decided to set test api key in env variables, other services should use the env variable directory instead.

packages/health-check/src/utils.js
```diff
@@ -61,6 +61,19 @@ function getRequiredEnvironmentVariable(name) {
...
+ * @returns {string} API key
+ */
+function getAPIKey() {
+  const apiKey = getRequiredEnvironmentVariable("ACCOUNTS_TEST_USER_API_KEY");
+  if (!apiKey) {
+    throw new Error("Missing or empty environment variable ACCOUNTS_TEST_USER_API_KEY.");
+  }
+  return apiKey;
+}
+
```
kwypchlo, line 76: I don't think we need this method if we've decided to set test api key in env variables, other services should use the env variable directory instead.

kwypchlo (changes requested):

docker/nginx/conf.d/include/cors-headers
m-cat: Okay, my supposition was correct. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers answered my question.

m-cat (approved):

PR #1725, opened 2022-02-16
ops: initial Dashboard v2 directory setup

meeh0w:
# PULL REQUEST

## Overview
* Just a project setup with dependencies and some initial utilities. Actual visual stuff will come in the next PR.


packages/dashboard-v2/README.md
```diff
@@ -0,0 +1,10 @@
+# Skynet Account Dashboard
+
+Code behind [account.skynetpro.net](https://account.skynetpro.net/)
+
+
+## Development
+This is a Gatsby application. To run it locally, all you need is:
+
+* `npm install`
+* `npm run start`
```
kwypchlo, line 10: we're using yarn everywhere else, I would appreciate consistency unless there is a case why we should use npm instead

packages/dashboard-v2/gatsby-ssr.js
```diff
@@ -0,0 +1,13 @@
...
+import "@fontsource/sora/500.css"; // medium
+import "@fontsource/sora/600.css"; // semibold
+import "@fontsource/source-sans-pro/400.css"; // normal
+import "@fontsource/source-sans-pro/600.css"; // semibold
+import "./src/styles/global.css";
+
+export function wrapPageElement({ element, props }) {
+  const Layout = element.type.Layout ?? React.Fragment
+  return <Layout {...props}>{element}</Layout>
+}
```
kwypchlo, line 13: we're not doing ssr because our pages will be hosted statically, we can skip this file

kwypchlo (changes requested):

packages/dashboard-v2/gatsby-ssr.js
meeh0w: I initially added it because I had issues with first render on development server, but it looks like it's no longer needed (probably got resolved by clearing gatsby's `.cache` 🤔 )

Removed in 0c77705

packages/dashboard-v2/README.md
meeh0w: Cool 👍  00eb919

kwypchlo (approved):

PR #1730, opened 2022-02-17
Dashboard v2 code quality checks

meeh0w:
## Overview
* Added eslint
* Added Prettier
* Addressed some missing peer-dependencies
* Added a Github action to lint the code on each PR

### DIFF
--- a/packages/dashboard-v2/.prettierrc.json
+++ b/packages/dashboard-v2/.prettierrc.json
@@ -1,0 +1,5 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "printWidth": 120
+}
--- a/packages/dashboard-v2/.prettierrc.json
+++ b/packages/dashboard-v2/.prettierrc.json
@@ -1,5 +1,3 @@
 {
-  "semi": false,
-  "singleQuote": true,
   "printWidth": 120
 }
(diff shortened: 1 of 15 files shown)


packages/dashboard-v2/.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "semi": false,
```
meeh0w, line 2: Let me know what you think about it. I've been using this no-semis, single-quoted setup for the past 5 years and it's much cleaner & easier imho.

* Single quotes are easier to type (one keypress instead of two).
    * When you need escape the single quotes, you can still use double-quotes or backticks, but in my experience any content with apostrophes usually lives outside of JS (e.g. in a YML file with translations).
* Semicolons will be inserted by Prettier whenever they're necessary ([reference](https://prettier.io/docs/en/rationale.html#semicolons))

kwypchlo: I would like to leave it as is to keep consistency across other directories and projects and also minimise the amount of project/team opinion based configuration and defer to widely accepted defaults. I remember the time before-prettier where every project/team would come up with it's own code style guidelines and every time it was a mess :)

I'm actually on the fence when it comes to line length override that we do everywhere (we use 120 from default 80) but I tried default 80 chars and it just makes code less readable most of the times and I found myself using shorter, less verbose variable names just to keep something in one line which was really dumb.

meeh0w: Alrighty 👍

peterjan (approved):

PR #1736, opened 2022-02-18
add skynet-cache-ratio header to access log

kwypchlo:
source: https://gitlab.com/SkynetLabs/skyd/-/merge_requests/328

### DIFF
--- a/docker/nginx/conf.d/include/proxy-cache-downloads
+++ b/docker/nginx/conf.d/include/proxy-cache-downloads
@@ -4,18 +4,18 @@ proxy_cache_min_uses 3; # cache after 3 uses
 proxy_cache_valid 200 206 307 308 48h; # keep 200, 206, 307 and 308 responses valid for up to 2 days
 add_header X-Proxy-Cache $upstream_cache_status; # add response header to indicate cache hits and misses
 
-# bypass - this will bypass cache hit on request (status BYPASS)
-# but still stores file in cache if cache conditions are met
-proxy_cache_bypass $cookie_nocache $arg_nocache;
-
-# no cache - this will ignore cache on request (status MISS)
-# and does not store file in cache under no condition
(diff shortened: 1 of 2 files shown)


docker/nginx/conf.d/include/proxy-cache-downloads
```diff
@@ -4,18 +4,18 @@ proxy_cache_min_uses 3; # cache after 3 uses
...
 
+# map skyd env variable value to "1" for true and "0" for false (expected by proxy_no_cache)
+set_by_lua_block $skyd_disk_cache_enabled {
+    return os.getenv("SKYD_DISK_CACHE_ENABLED") == "true" and "1" or "0"
+}
+
 # bypass - this will bypass cache hit on request (status BYPASS)
 # but still stores file in cache if cache conditions are met
-proxy_cache_bypass $cookie_nocache $arg_nocache;
+proxy_cache_bypass $cookie_nocache $arg_nocache $skyd_disk_cache_enabled;
```
kwypchlo, line 14: adding `$skyd_disk_cache_enabled` check to `proxy_cache_bypass` in addition to `proxy_no_cache` will cause nginx to report cache status `"BYPASS"` instead of `"MISS"` which makes more sense

peterjan (approved):
LGTM, only `nit` I had was the strict evaluation of "true" instead of it being truthy, like 1 or True or whatever but I think it's fine. Just commenting in case you suddenly feel it might be better to write a helper function that evaluates environment variables in a truthy fashion

kwypchlo:
> LGTM, only `nit` I had was the strict evaluation of "true" instead of it being truthy, like 1 or True or whatever but I think it's fine. Just commenting in case you suddenly feel it might be better to write a helper function that evaluates environment variables in a truthy fashion

thanks for the idea, documented in a follow up issue

PR #1759, opened 2022-02-22
ensure path exists and fix filename

kwypchlo:
- ensure sync paths exist (otherwise it will throw error)
- fix skylinks filename
- replace dots with dashes in domain names (instead of removing them completely) for better readability

### DIFF
--- a/scripts/backup-aws-s3.sh
+++ b/scripts/backup-aws-s3.sh
@@ -20,24 +20,29 @@ for ENV_VARIABLE in "${ENV_VARIABLES[@]}"; do
     export ${ENV_VARIABLE_VALUE}
 done
 
-# create bucket skynet-backup-[portaldomain] (portal domain is stripped of all non alnum characters)
-# ie. siasky.net backup results in skynet-backup-siaskynet basket name
-BUCKET_NAME=$(echo skynet-backup-${PORTAL_DOMAIN} | tr -cd '[[:alnum:]]-')
+# create bucket skynet-backup-[portaldomain] (replace dots with dashes and strip anything other than alnum)
+# ie. siasky.net backup results in skynet-backup-siasky-net basket name
+BUCKET_NAME=$(echo skynet-backup-${PORTAL_DOMAIN} | tr  '.'  '-' | tr -cd '[[:alnum:]]-')
 
 # create server prefix
 if test -z "${SERVER_DOMAIN}"; then
-    SERVER_PREFIX=$(echo ${SERVER_UID} | tr -cd '[[:alnum:]]-') # if domain name is empty use just uid
+    # if domain name is empty use just uid (replace dots with dashes and strip anything other than alnum)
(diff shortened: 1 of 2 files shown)


scripts/backup-aws-s3.sh
```diff
@@ -20,24 +20,29 @@ for ENV_VARIABLE in "${ENV_VARIABLES[@]}"; do
...
 
 # sync all nginx logs
+mkdir -p /home/user/skynet-webportal/docker/data/nginx/logs # ensure path exists
 aws s3 sync --no-progress /home/user/skynet-webportal/docker/data/nginx/logs s3://${BUCKET_NAME}/${SERVER_PREFIX}/docker/data/nginx/logs
 
-# generate and sync skylinks
-docker exec sia siac skynet ls --recursive --alert-suppress > /home/user/skynet-webportal/logs/skylinks-$(date +"%Y-%m-%d").log
-aws s3 cp --no-progress /home/user/skynet-webportal/logs/skylinks.log s3://${BUCKET_NAME}/${SERVER_PREFIX}/logs/skylinks.log
+# generate and sync skylinks dump
+SKYLINKS_PATH=logs/skylinks/$(date +"%Y-%m-%d").log
```
ro-tex, line 43: This relative path is not ideal, right? Can we make it cover the full absolute path?

scripts/backup-aws-s3.sh
```diff
@@ -20,24 +20,29 @@ for ENV_VARIABLE in "${ENV_VARIABLES[@]}"; do
...
+mkdir -p /home/user/skynet-webportal/logs/skylinks # ensure path exists
+docker exec sia siac skynet ls --recursive --alert-suppress > /home/user/skynet-webportal/${SKYLINKS_PATH}
```
ro-tex, line 45: It would be nice if this operation doesn't overwrite the log each time, e.g. `>` can become `>>`.

firyx (approved):

ro-tex (dismissed):
I left some comments with what I consider to be improvements but they are not blocking.

scripts/backup-aws-s3.sh
kwypchlo: I'm changing to just run it daily instead so since there's date suffix it will not be overwritten anyway. Also the format of the dump is not append-friendly unfortunately.

scripts/backup-aws-s3.sh
kwypchlo: it's relative but we use it as a part of absolute path, both for local filename and for s3 so it can't be exactly the same (local has /home/user/skynet-webportal base and s3 has bucket name base)

firyx (approved):

scripts/backup-aws-s3.sh
ro-tex: Oh. I didn't notice it in the bucket path.

ro-tex (approved):

PR #1760, opened 2022-02-23
Dashboard V2: main screen

meeh0w:
## Overview

* [Visual overview](https://skynetpro.net/AAC-4JgZpvPcELSM8aG7lOPoOYeauwUPFT7P3Hlcmegefg)

### 👣  Commit-by-commit review is recommended :)

* **Composes most of the main dashboard screen** (with mocked data)
* Makes top navigation bar responsive
* Adds a CSS helper to make media queries easier inside Styled Components. Also allows referencing computed Tailwind config from javascript, by using `babel-preval-plugin` (config is evaluated on build time, not in the browser).
* Some styling fixes / improvements to generic components (`Tabs` / `Panel` / `Table`).
* Adds `gatsby-plugin-styled-components`, as without it adding custom CSS only works in `gatsby develop` mode, but not in a production build (`gatsby build`).
* Adds back `gatsby-ssr.js` file -- it's needed for `gatsby build` followed by `gatsby serve` flow (seems like SSR is enabled with `gatsby serve`).


packages/dashboard-v2/src/components/CurrentUsage/GraphBar.js
```diff
@@ -0,0 +1,35 @@
...
+const BarLabel = styled.span.attrs({
+  className: "bg-white rounded border-2 border-palette-200 px-3 whitespace-nowrap absolute shadow",
+})`
+  right: max(0%, ${({ $percentage }) => 100 - $percentage}%);
+  top: -0.5rem;
+  transform: translateX(50%);
+`;
+
+export const GraphBar = ({ value, limit, label }) => {
+  const percentage = (value / limit) * 100;
```
peterjan, line 23: `nit` cover div by zero here?

peterjan (approved):
LGTM 👍

packages/dashboard-v2/src/components/Slider/Slider.js
```diff
@@ -0,0 +1,105 @@
...
+
+Slider.defaultProps = {
+  breakpoints: [
+    {
+      minWidth: parseInt(theme.screens.xl),
+      scrollable: true,
+      visibleSlides: 3,
+    },
+    {
+      minWidth: parseInt(theme.screens.md, 10),
```
kwypchlo, line 94: instead of minWidth we should probably use breakpoint name as a string and then resolve the values in useActiveBreakpoint

packages/dashboard-v2/src/components/Slider/useActiveBreakpoint.js
```diff
@@ -0,0 +1,23 @@
...
+  const findActiveBreakpoint = useCallback(
+    () => monitoredBreakpoints.find((breakpoint) => windowWidth >= breakpoint.minWidth),
+    [monitoredBreakpoints, windowWidth]
+  );
+
+  const [activeBreakpoint, setActiveBreakpoint] = useState(findActiveBreakpoint());
+
+  useEffect(() => {
+    setActiveBreakpoint(findActiveBreakpoint());
+  }, [findActiveBreakpoint]);
```
kwypchlo, line 20: in the follow up, separate pull request, let's include https://www.npmjs.com/package/eslint-plugin-react-hooks and scan the whole project with it

packages/dashboard-v2/src/components/LatestActivity/ActivityTable.js
```diff
@@ -0,0 +1,22 @@
...
+    <Table style={{ tableLayout: "fixed" }}>
+      <TableBody>
+        {data.map(({ name, type, size, uploaded, skylink }) => (
+          <TableRow key={skylink}>
+            <TableCell>{name}</TableCell>
+            <TableCell width="80px">{type}</TableCell>
+            <TableCell width="80px" align="right">
+              {size}
+            </TableCell>
+            <TableCell width="180px">{uploaded}</TableCell>
```
kwypchlo, line 15: let's use tailwind class names for widths

packages/dashboard-v2/package.json
```diff
@@ -21,6 +21,7 @@
   "dependencies": {
     "@fontsource/sora": "^4.5.3",
     "@fontsource/source-sans-pro": "^4.5.3",
+    "byte-size": "^8.1.0",
```
kwypchlo, line 24: we're using https://www.npmjs.com/package/pretty-bytes in other repositories, would be nice to keep it consistent

packages/dashboard-v2/src/components/CurrentUsage/CurrentUsage.js
```diff
@@ -0,0 +1,52 @@
...
+  return (
+    <>
+      <h4>{`${size(storageUsed)} of ${size(storageLimit)}`}</h4>
+      <p className="text-palette-400">
+        {filesUsed} of {filesLimit} files
+      </p>
+      <div className="relative mt-7 font-sans uppercase text-xs">
+        <div className="flex place-content-between">
+          <span>Storage</span>
+          <span>{`${size(storageLimit)}`}</span>
```
kwypchlo, line 30: you don't need to wrap the function call in ticks here

packages/dashboard-v2/src/components/CurrentUsage/CurrentUsage.js
```diff
@@ -0,0 +1,52 @@
...
+        </div>
+        <UsageGraph>
+          <GraphBar value={storageUsed} limit={storageLimit} label={size(storageUsed)} />
+          <GraphBar value={filesUsed} limit={filesLimit} label={{ value: filesUsed, unit: "files" }} />
```
kwypchlo, line 34: you're passing new label object reference each time this component re-renders, you should either useMemo on the label object or split the label into label-value and label-unit

packages/dashboard-v2/src/components/CurrentUsage/GraphBar.js
```diff
@@ -0,0 +1,35 @@
...
+
+  return (
+    <div className="relative flex items-center">
+      <Bar $percentage={percentage}>
+        <BarTip />
+      </Bar>
+      <BarLabel $percentage={percentage}>
```
kwypchlo, line 30: same concern about $ prefixed params

packages/dashboard-v2/gatsby-ssr.js
```diff
@@ -0,0 +1,13 @@
...
+import "@fontsource/sora/500.css"; // medium
+import "@fontsource/sora/600.css"; // semibold
+import "@fontsource/source-sans-pro/400.css"; // normal
+import "@fontsource/source-sans-pro/600.css"; // semibold
+import "./src/styles/global.css";
+
+export function wrapPageElement({ element, props }) {
+  const Layout = element.type.Layout ?? React.Fragment;
+  return <Layout {...props}>{element}</Layout>;
+}
```
kwypchlo, line 13: seems you've added this file again in this pr - what's up with that, is the build failing without this file?

packages/dashboard-v2/src/components/Tabs/ActiveTabIndicator.js
```diff
@@ -35,6 +35,9 @@ export const ActiveTabIndicator = ({ tabRef }) => {
...
 
-ActiveTabIndicator.propTypes = {
-  tabRef: PropTypes.oneOfType([PropTypes.func, PropTypes.shape({ current: PropTypes.instanceOf(Element) })]),
-};
+// Needed, because we're using an Element constant here which Gatsby doesn't recognize during build time.
+if (typeof window !== "undefined") {
+  ActiveTabIndicator.propTypes = {
+    tabRef: PropTypes.oneOfType([PropTypes.func, PropTypes.shape({ current: PropTypes.instanceOf(Element) })]),
+  };
+}
```
kwypchlo, line 43: can you use `PropTypes.elementType` instead of `Element` so we don't need to add this window condition?

packages/dashboard-v2/src/components/NavBar/NavBar.js
```diff
@@ -1,4 +1,4 @@
-import { Link } from "gatsby";
+import Link from "gatsby-link";
```
kwypchlo, line 1: https://www.npmjs.com/package/gatsby-link

> All components and utility functions from this package are now exported from [gatsby](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby) package. You should not import anything from this package directly.

kwypchlo (changes requested):

packages/dashboard-v2/src/components/Slider/Slider.js
```diff
@@ -0,0 +1,105 @@
...
+    [slides, visibleSlides, setActiveIndex]
+  );
+
+  return (
+    <Container>
+      <Scroller
+        $visibleSlides={visibleSlides}
+        $allSlides={slides.length}
+        $activeIndex={activeIndex}
+        $scrollable={scrollable}
```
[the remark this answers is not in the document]
meeh0w: It's a way to mark the prop as transient: https://styled-components.com/docs/api#transient-props
It prevents them from being passed down to HTML nodes.

Styled Components can filter out most non-standard props, but I figured it's better to prefix all props (that are only used for styles) with `$` as opposed to having it mixed.

packages/dashboard-v2/src/components/Slider/useActiveBreakpoint.js
meeh0w: It's already a part of the project, as `.eslintrc` extends `react-app` config. Adding `setActiveBreakpoint` is not necessary here, as React guarantees it to be stable ([reference](https://reactjs.org/docs/hooks-reference.html#usestate)):
image

packages/dashboard-v2/gatsby-ssr.js
meeh0w: Yeah, I mentioned it in the PR description:
> * Adds back gatsby-ssr.js file -- it's needed for `gatsby build` followed by `gatsby serve` flow (seems like SSR is enabled with `gatsby serve`).

packages/dashboard-v2/src/components/Slider/Slider.js
meeh0w: Good idea 👍

meeh0w: Improved in 278ab698

packages/dashboard-v2/package.json
meeh0w: Fair enough 👍  aaaa9c13

packages/dashboard-v2/src/components/NavBar/NavBar.js
meeh0w: Changed in 5123a23e -- making it work requires a Storybook addon.

packages/dashboard-v2/src/components/CurrentUsage/GraphBar.js
meeh0w: @peterjan  @kwypchlo How do we want to interpret `0` in terms of usage limit, though? Would it mean user cannot store any data at all, or that the storage is unlimited?

packages/dashboard-v2/src/components/Tabs/ActiveTabIndicator.js
meeh0w: Never mind, Gatsby caching mislead me again -- after a restart it started complaining. 🙄 
There is [a feature request](https://github.com/facebook/prop-types/issues/240) in `prop-types` repo, but it never got too much traction (I guess TypeScript is more common these days).

I reverted it back to what it was, this way we get the protection & build passes (`prop-types` are runtime anyways, so we don't need them on build time, and we don't do SSR).

7f26c95b

kwypchlo:
```
ActiveTabIndicator.propTypes = {
  tabRef: PropTypes.oneOfType([PropTypes.func, PropTypes.shape({ current: PropTypes.elementType })]),
};
```

This got me successful build. I did run gatsby clean before the build too so not sure why it failed for you.

packages/dashboard-v2/gatsby-ssr.js
kwypchlo: I just locally removed this file and ran a build and it passed. Can you double check please?

kwypchlo:
there seems to be some issue with navigation bar @meeh0w 

Screenshot 2022-02-28 at 11 38 32

packages/dashboard-v2/src/components/Slider/useActiveBreakpoint.js
kwypchlo: perfect

packages/dashboard-v2/src/components/Slider/Slider.js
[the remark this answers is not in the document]
kwypchlo: as much as I don't like the name prefix you are totally correct here 👍

packages/dashboard-v2/src/components/CurrentUsage/GraphBar.js
kwypchlo: let's say it means unlimited

packages/dashboard-v2/src/components/Tabs/ActiveTabIndicator.js
meeh0w: Build doesn't fail, but it complains when you visit the page.

kwypchlo (approved):

meeh0w:
Hey @kwypchlo @peterjan, I updated the PR with the division-by-zero code, can you take a final look please?

kwypchlo (approved):

peterjan (approved):

PR #1761, opened 2022-02-23
nginx stats endpoint

kwypchlo:
/nginx/stats endpoint returns whatever is available from nginx [stub status](http://nginx.org/en/docs/http/ngx_http_stub_status_module.html) module but in a json format

example output:
```json
{
  "connections_reading": 0,
  "connections_writing": 1,
  "connections_waiting": 1,
  "connections_active": 2
}
```

deployed currently to https://dev3.siasky.dev/nginx/stats

**caveat: not sure we need that, it would be a good addition for the siasky.tools dashboard though if we can keep refreshing it**


docker/nginx/conf.d/server/server.api
```diff
@@ -72,6 +72,29 @@ location /skynet/stats {
     proxy_pass http://sia:9980/skynet/stats;
 }
 
+location /nginx/stats {
```
peterjan, line 75: is this something we want to expose?

peterjan (approved):

docker/nginx/conf.d/server/server.api
kwypchlo: idk, it feels like we should have an aggregate from all servers exposed and not just a single server data since that's really tied to our current load balancing architecture, same goes for /skynet/stats

PR #1763, opened 2022-02-24
switch malware-scanner to image

kwypchlo:
- use [malware-scanner](https://hub.docker.com/repository/docker/skynetlabs/malware-scanner) dockerhub image
- hardcode CLAMAV_IP and CLAMAV_PORT env variables - those are not portal operator configurable and should not be overridden in .env file

depends on https://github.com/SkynetLabs/malware-scanner/pull/13


docker-compose.malware-scanner.yml
```diff
@@ -26,19 +26,15 @@ services:
         ipv4_address: 10.10.10.100
 
   malware-scanner:
-    build:
-      context: ./docker/malware-scanner
-      dockerfile: Dockerfile
-      args:
-        branch: main
+    image: skynetlabs/malware-scanner
```
peterjan, line 29: 🥳  yes! this is the way

ChrisSchinnerl (approved):

PR #1767, opened 2022-02-24
use handshake docker image

kwypchlo:
Switch from building handshake node locally to using docker image [skynetlabs/hsd](https://hub.docker.com/r/skynetlabs/hsd) based off our [docker-hsd](https://github.com/SkynetLabs/docker-hsd) repository.

PR #1770, opened 2022-02-25
use hsd image instead of Dockerfile

kwypchlo:
use https://hub.docker.com/repository/docker/skynetlabs/hsd image instead of building handshake deamon on every deploy


kwypchlo:
duplicate of https://github.com/SkynetLabs/skynet-webportal/pull/1767

PR #1771, opened 2022-02-25
set portal domain header to server domain when accessing through server domain

kwypchlo:
Portals always return 2 informative headers:
- `Skynet-Portal-Api` ie. https://siasky.net
- `Skynet-Server-Api` ie. https://server-one.siasky.net

This branch changes `Skynet-Portal-Api` when user accesses through server domain (opens up https://server-one.siasky.net) from https://siasky.net to https://server-one.siasky.net.

This change is going to affect only users accessing portal through a specific server by it's server domain and mostly relates to skynet-js usage where skynet-js sends a HEAD request to current address and uses Skynet-Portal-Api header from response as the default endpoint for the connection. This change will allow us to go to https://server-one.siasky.net and use skynet-js to upload a file and it will send it to https://server-one.siasky.net and not https://siasky.net like it did before. 

**Testing:** Currently deployed to https://dev3.siasky.dev - open the main website, open network tab in chrome, uplaod a file and observe requests go straight to https://dev3.siasky.dev and then the uploaded skylink url that you get after uploading below the uplaoder will also point directly to https://dev3.siasky.dev/[skylink]


ro-tex (dismissed):

docker/nginx/conf.d.templates/server.account.conf
```diff
@@ -9,6 +9,9 @@
 	server {
 		server_name account.{{PORTAL_DOMAIN}}; # example: account.siasky.net
 
+		set_by_lua_block $skynet_portal_domain { return "{{PORTAL_DOMAIN}}" }
+		set_by_lua_block $skynet_server_domain { return "{{SERVER_DOMAIN}}" == "" and "{{PORTAL_DOMAIN}}" or "{{SERVER_DOMAIN}}" }
```
peterjan, line 13: this is mixing `and` and `or` without brackets, I'm assuming it's correct so I'll approve but should we add them? could we add them even though they're implicit for whatever reason to improve readability

peterjan (dismissed):
this is great, in general, though I think we're missing way too much context with changes like this, I would even say that the entire PR description belongs somewhere in our Nginx docs, or at least a very thorough description of why those headers are set and why they're set to whatever value they're set to

docker/nginx/conf.d/server/server.api
```diff
@@ -247,8 +247,7 @@ location /skynet/tus {
     proxy_set_header X-Forwarded-Proto $scheme;
 
     # rewrite proxy request to use correct host uri from env variable (required to return correct location header)
-    set_by_lua_block $server_domain { return os.getenv("SERVER_DOMAIN") }
```
peterjan, line 250: so I guess we can get rid of this because it's now set in the `server` blocks right?

kwypchlo: yes 👍

docker/nginx/conf.d.templates/server.account.conf
kwypchlo: changed in https://github.com/SkynetLabs/skynet-webportal/pull/1771/commits/520f935cd2d3ba7eb2dd1586e2681907111a3cc3

peterjan (approved):

ChrisSchinnerl (approved):

PR #1801, opened 2022-03-01
feat(dashboard-v2): API access in dev environment

meeh0w:
## Overview

* Goal here is to have access to the APIs despite CORS (we're starting a proxy server along with `gatsby develop`)
* Proxy is targeting prod as agreed with @kwypchlo, but I'll change it as soon as possible (dev servers are not too stable at the moment)
* There is no way to authenticate here yet (I just manually created a cookie in the browser and it's sufficient for me to continue coding for now)


kwypchlo (approved):
good good

PR #1822, opened 2022-03-01
Dashboard v2 - Uploader

meeh0w:
## ⚠️ Based on #1801, don't merge prematurely :)
* _Reviewing commit-by-commit should be helpful_

## Overview
* Adds some new, shiny icons
* Fixes styles for `Button`
* Improves `Tabs`, so the content fills available space. Also stops unmounting the child component when tab becomes inactive (this will allow the user to change tabs without interrupting any background tasks started by the child component, like uploads).
* Adds `ProgressBar` component
* Copies `Uploader` from `packages/website` and simplifies it a little bit (it contained quite a lot of components we won't need in the dashboard view)
* Adds `Uploader` on the main screen

## 🎥  Visual overview
https://user-images.githubusercontent.com/3853033/156216391-96afd121-44ef-41ae-ac19-20137415991d.mov


packages/dashboard-v2/src/components/Tabs/TabPanel.js
```diff
@@ -3,17 +3,17 @@ import PropTypes from "prop-types";
...
+  <div
+    role="tabpanel"
+    id={`tabpanel-${tabId}`}
+    aria-labelledby={`tab-${tabId}`}
+    {...props}
+    style={{ display: active ? "block" : "none" }}
+  >
+    {children}
+  </div>
+);
```
kwypchlo, line 16: I think the component should return null until it is mounted for the first time (active is true for the first time). That way we don't load the invisible tabs and first paint is faster and we potentially save some unnecessary requests if other invisible tabs would request a list of data or sth like that.

packages/dashboard-v2/src/components/Uploader/UploaderItem.js
```diff
@@ -0,0 +1,134 @@
+import * as React from "react";
+import cn from "classnames";
+import bytes from "pretty-bytes";
+import { StatusCodes } from "http-status-codes";
+import copy from "copy-text-to-clipboard";
+import path from "path-browserify";
+import { useTimeoutFn } from "react-use";
+// import ms from "ms";
```
kwypchlo, line 8: not needed any more?

kwypchlo (changes requested):
Small comments, otherwise looks good 👍 

Also since you're at it, maybe you can take care of this warning that shows up in github:

Screenshot 2022-03-01 at 18 54 06

packages/dashboard-v2/src/components/Tabs/TabPanel.js
meeh0w: Good idea 👍  e3fe7c60

packages/dashboard-v2/src/components/Uploader/UploaderItem.js
meeh0w: Yup. Cleaned it up here 👉  18e198d3

kwypchlo (approved):

packages/dashboard-v2/src/components/Tabs/TabPanel.js
```diff
@@ -1,15 +1,33 @@
...
 
 /**
  * Besides documented props, it accepts all HMTL attributes a `<div>` element does.
  */
 export const TabPanel = ({ children, active, tabId, ...props }) => {
-  if (!active) {
+  const [wasActivated, setWasActivated] = useState(false);
+
+  useEffect(() => {
+    if (active) {
```
peterjan, line 11: `(uber) nit` this is redundant, you can just do `setWasActivated(active)`

packages/dashboard-v2/src/components/Uploader/buildUploadErrorMessage.js
```diff
@@ -0,0 +1,37 @@
...
+
+export default function buildUploadErrorMessage(error) {
+  // The request was made and the server responded with a status code that falls out of the range of 2xx
+  if (error.response) {
+    if (error.response.data.message) {
+      return `Upload failed with error: ${error.response.data.message}`;
+    }
+
+    const statusCode = error.response.status;
+    const statusText = getReasonPhrase(error.response.status);
```
peterjan, line 12: you can use `statusCode` here

peterjan (approved):

packages/dashboard-v2/src/components/Uploader/buildUploadErrorMessage.js
meeh0w: Thanks for this -- I'll address it in future PRs to not cause a chain of changes between this branch and other PRs that depend on it 👌

PR #1830, opened 2022-03-03
redirect sdk docs to sdk.skynetlabs.com

kwypchlo:
redirect /docs portal endpoint (ie. https://siasky.net/docs) to https://sdk.skynetlabs.com

- works with static subpages like https://siasky.net/docs/v4 => https://sdk.skynetlabs.com/v4
- works with trailing slash or without it ie https://siasky.net/docs or https://siasky.net/docs/

### DIFF
--- a/docker/nginx/conf.d/server/server.api
+++ b/docker/nginx/conf.d/server/server.api
@@ -21,7 +21,7 @@ client_max_body_size 128k;
 rewrite ^/portals /skynet/portals permanent;
 rewrite ^/stats /skynet/stats permanent;
 rewrite ^/skynet/blacklist /skynet/blocklist permanent;
-rewrite ^/docs https://sdk.skynetlabs.com permanent;
+rewrite ^/docs(?:/(.*))?$ https://sdk.skynetlabs.com/$1 permanent;
 
 location / {
     include /etc/nginx/conf.d/include/cors;
--- a/docker/nginx/conf.d/server/server.api
+++ b/docker/nginx/conf.d/server/server.api
@@ -21,6 +21,7 @@ client_max_body_size 128k;
 rewrite ^/portals /skynet/portals permanent;
 rewrite ^/stats /skynet/stats permanent;
 rewrite ^/skynet/blacklist /skynet/blocklist permanent;
+rewrite ^/docs https://sdk.skynetlabs.com permanent;
 
 location / {
     include /etc/nginx/conf.d/include/cors;
@@ -37,10 +38,6 @@ location / {
 
 location @fallback {
     proxy_pass http://website:9000;
-}
-
-location /docs {
-    proxy_pass https://skynetlabs.github.io/skynet-docs;
 }
 
 location /skynet/blocklist {


m-cat:
I'm missing the background here. I assume this has been discussed already but I would still like to know what the reasoning for this change is.

kwypchlo:
> I'm missing the background here. I assume this has been discussed already but I would still like to know what the reasoning for this change is.

we are moving the sdk docs off the default portal page (page that serves as a default for all community portals like https://fileportal.org) to our company page https://skynetlabs.com to have more control over it and improve discoverability

m-cat:
@kwypchlo Thanks, makes sense. Why is it moving from `/docs` path to the `sdk.` subdomain? I do like the change, but I remember people saying that "docs" made more sense than "sdk" back when we were discussing it the first time.

kwypchlo:
> @kwypchlo Thanks, makes sense. Why is it moving from `/docs` path to the `sdk.` subdomain? I do like the change, but I remember people saying that "docs" made more sense than "sdk" back when we were discussing it the first time.

Because we already have docs => https://docs.skynetlabs.com and it's super confusing. I just arbitrarily chose the "sdk" because nothing else made much sense. I still think our documentation should receive proper restructuring because we also have support => https://support.skynetlabs.com and our docs and support gitbooks look the same to me.

docker/nginx/conf.d/server/server.api
```diff
@@ -21,6 +21,7 @@ client_max_body_size 128k;
 rewrite ^/portals /skynet/portals permanent;
 rewrite ^/stats /skynet/stats permanent;
 rewrite ^/skynet/blacklist /skynet/blocklist permanent;
+rewrite ^/docs(?:/(.*))?$ https://sdk.skynetlabs.com/$1 permanent;
```
m-cat, line 24: I understand the match (`(.*)`) but I don't understand this part: `?:/`.

This document helped me a bit: 
https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#rewrite

m-cat, line 24: So this is redirecting `/docs` right? And we are keeping that endpoint just for backwards-compatibility? I feel like we don't need it if we are moving off of `siasky.net` as we don't have to worry about existing `skynetlabs.com/docs` links.

m-cat, line 24: I wonder if we can avoid hard-coding the name of the portal `skynetlabs.com` here.

m-cat (changes requested):
The idea behind this is good. I don't fully understand the code though. Would be nice if there were a deployed version of this somewhere so we could test it.

m-cat:
I see that this is deployed on `dev3.siasky.net` but it always redirects me to `skynetpro.com` instead of staying on `dev3.

peterjan:
I guess this move is temporary right? We should probably merge all of the content under `sdk` into the `docs`? That makes the most sense to me, leaving us with just `docs.skynetlabs.com` which has an SDK section.

docker/nginx/conf.d/server/server.api
kwypchlo: `?:` means that the outside group is non matching so we don't pollute the `$1` result with unnecessary match

kwypchlo: we don't need skynetlabs.com to be backwards compatible with /docs because no one ever used /docs on skynetlabs.com yet

we are keeping someportal.com/docs redirect to sdk.skynetlabs.com as a backwards compatibility for a while because some people might have bookmarked it and we want them redirected

kwypchlo: I don't think we'd want that, that's our direct link for sdk that we're going to share with people and that seo should track

kwypchlo:
> I see that this is deployed on `dev3.siasky.net` but it always redirects me to `skynetlabs.com` instead of staying on `dev3.

it's not deployed there, the way it works right now is github.io redirects any traffic that reaches it and is not skynetlabs.com to skynetlabs.com - this means all links, even hardlinks like siasky.net/docs/v4 will get redirected to the main page of sdk.skynetlabs.com because it's a dummy redirect - this is why we want to keep the legacy in-portal redirect for a while, just in case someone bookmarked a subpage of the docs (see pr description)

kwypchlo:
> I guess this move is temporary right? We should probably merge all of the content under `sdk` into the `docs`? That makes the most sense to me, leaving us with just `docs.skynetlabs.com` which has an SDK section.

I'm not aware of any plans like that

m-cat (approved):
Looks good. 👍

dghelm (approved):
Looks good to me. I think medium term we need to reconfigure docs including incorporating API endpoint docs for skyd and web portal, but for now this is a good move.

Issue #1372, opened 2021-11-10
NGINX fallback for Website Skylink error on 404


MSevey:
## Describe the bug

coroutine 404 is not being handled properly.

## To Reproduce

Delete the skylink here https://github.com/SkynetLabs/skynet-webportal/blob/master/docker/nginx/conf.d/server/server.api#L29, or change it, and rebuild/restart nginx. 

## Expected behavior

Website still loads fine.

## Screenshots

When I changed some characters in the skylink I actually got the skylink download error. 
<img width="947" alt="Screen Shot 2021-11-10 at 10 45 12 AM" src="https://user-images.githubusercontent.com/15232757/141146390-3443e87a-ae26-429b-9c0d-57ffbc9e6970.png">

## Additional context

Karol and I talked about it in DM.

PR #1835, opened 2022-03-06
Revert "use skynet to load our homepage"

kwypchlo:
This reverts commit ffb1e499b9c996dd1075c2b4778384f39f49101e.

- website fallback is broken if skyd is not responsive https://github.com/SkynetLabs/skynet-webportal/issues/1372
- different portal operators have different website needs, defaulting to a skylink that gets deployed when skynetlabs deploys new version is not wanted
- there was no easy way to override the skylink and go with custom website container without changing nginx configuration
- merging changes to master meant that all portals get new version of website rolled out immediately and we already had issues with website and api mismatch - website should roll out with api changes
- version of the website from skylink and version of the website could have been mismatched and fallback, even if working properly would cause different versions to be served
- removing skylink also makes it a little bit easier to produce different versions of website for our usage on new pro and free clusters

closes https://github.com/SkynetLabs/skynet-webportal/issues/1372

### DIFF
--- a/docker/nginx/conf.d/server/server.api
+++ b/docker/nginx/conf.d/server/server.api
@@ -26,17 +26,6 @@ rewrite ^/docs(?:/(.*))?$ https://sdk.skynetlabs.com/$1 permanent;
 location / {
     include /etc/nginx/conf.d/include/cors;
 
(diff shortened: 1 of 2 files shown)


meeh0w (approved):

peterjan (approved):
too bad we won't be serving from skynet any longer tho

PR #1832, opened 2022-03-04
permanent redirect some-portal.com/hns/[domain] to [domain].hns.some-portal.com

kwypchlo:
this is a follow up to https://github.com/SkynetLabs/skynet-webportal/pull/1831 - please read motivation

Redirects `some-portal.com/hns/[domain]` to `[domain].hns.some-portal.com` with permanent redirect.

- works with and without trailing slash (ie. `some-portal.com/hns/[domain]` and `some-portal.com/hns/[domain]/`)
- works with pathname (ie. `some-portal.com/hns/[domain]/rick/roll/rickroll.jpeg` => `[domain].hns.some-portal.com/rick/roll/rickroll.jpeg`)
- works with query args (ie. `...?foo=bar&baz` will persist after redirect)

Also contains fix for case with empty args introduced by https://github.com/SkynetLabs/skynet-webportal/pull/1831.

Test links:

Regular handshake link
A) https://dev3.siasky.dev/hns/note-to-self
B) https://note-to-self.hns.dev3.siasky.dev/

Regular handshake link with trailing slash
A) https://dev3.siasky.dev/hns/note-to-self/
B) https://note-to-self.hns.dev3.siasky.dev/

Accessing single file from hosted directory
A) https://dev3.siasky.dev/hns/note-to-self/manifest.json
B) https://note-to-self.hns.dev3.siasky.dev/manifest.json

Query args persisting
A) https://dev3.siasky.dev/hns/note-to-self/manifest.json?foo=bar&baz
B) https://note-to-self.hns.dev3.siasky.dev/manifest.json?foo=bar&baz

PR #1831, opened 2022-03-03
permanent redirect some-portal.com/[skylink] to [base32-skylink].some-portal.com

kwypchlo:
When someone is accessing skylink by using pathname `https://siasky.net/[some-skylink]` we will redirect (301 permanent) to base32 subdomain version of that skylink `https://[same-skylink-but-base32].siasky.net`

Example redirect from A to B:
A) https://dev3.siasky.dev/CABqaqxbdeRjC5hAj1v8I8ZKDPT1uNPmarTE017m40Av_g
B )https://1006kqlcbdqu8oobj108umvs4f34k37kumsd7pjamj2d6nn6sd02vvg.dev3.siasky.dev

Example redirect with subfolders from A to B:
A) https://dev3.siasky.dev/CACddJs4tSX75E768mTYOjsfI9ZcHSivqJ2EMwjgRlsVgg/rick/roll/rickroll.jpeg
B) https://1009qt4r72qibuv49rtf4p6o78thu8umbgeihbt8jm2362708pdhb0g.dev3.siasky.dev/rick/roll/rickroll.jpeg

This is in line of what ipfs has been doing with their subdomain gateways and it adds the necessary layer of security that no apps should be ran from shared context of root domain.

Read more:
- path gateways: https://docs.ipfs.io/how-to/address-ipfs-on-web/#path-gateway
- subdomain gateways: https://docs.ipfs.io/how-to/address-ipfs-on-web/#subdomain-gateway

PR #1851, opened 2022-03-08
fix health checks working with root domain skylink and hns redirects

kwypchlo:
The library that we use in health checks - [got](https://www.npmjs.com/package/got), on a redirect made to a different hostname [clears some headers](https://github.com/sindresorhus/got/blob/main/source/core/index.ts#L774-L795) - namely "host", "cookie", "authorization", "username" and "password". Since the change introduced in https://github.com/SkynetLabs/skynet-webportal/pull/1831 and https://github.com/SkynetLabs/skynet-webportal/pull/1832 redirects skylinks and handshake domains from root domain api to subdomain level api, hostname is changed and previously set headers are cleared. We need to add them back in `beforeRedirect` hook. 

This issue is specific to [got](https://www.npmjs.com/package/got) and does not affect normal browser behaviour that attaches cookie to every matching request. Also this would not be an issue with api keys either because we're using custom "Skynet-Api-Key` header that wouldn't be cleared by the aforementioned piece of code.

### DIFF
--- a/packages/health-check/src/checks/critical.js
+++ b/packages/health-check/src/checks/critical.js
@@ -201,7 +201,11 @@ async function genericAccessCheck(name, url) {
   const data = { up: false, url };
 
   try {
-    const response = await got(url, { headers: { cookie: `nocache=true;${authCookie}` } });
+    const cookie = `nocache=true;${authCookie}`;
+    const response = await got(url, {
+      headers: { cookie },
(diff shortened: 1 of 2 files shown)


peterjan (approved):
This means health checks broke due to a last-minute-ish change right? Just trying to see whether we are missing the appropriate unit tests.

ChrisSchinnerl (approved):

PR #1852, opened 2022-03-09
improve user communication on subscription required portal

kwypchlo:
- on subscription required portal, show link to available subscription plans when user is authenticated but without paid plan
- make sure to return `auth_required = true` when portal mode is set to "subscription" - requiring subscription also implicitly requires authentication


packages/website/src/components/Uploader/Uploader.js
```diff
@@ -47,15 +47,37 @@ const LogInLink = () => {
...
   const [uploads, setUploads] = React.useState([]);
 
   const { data: accounts } = useAccounts();
-  const showAccountFeatures =
-    accounts && accounts.enabled !== false && !accounts?.auth_required && !accounts?.authenticated;
-  const disabledComponent =
-    accounts && accounts.enabled !== false && accounts?.auth_required && !accounts?.authenticated;
+
+  // variables extracted from useAccounts response
+  const isAccountsEnabled = accounts?.enabled === true;
```
meeh0w, line 70: `nit` These return boolean values, right?

`=== true` feels over the top, but maybe I'm missing something.

meeh0w (approved):

docker/nginx/libs/skynet/account.lua
```diff
@@ -78,7 +78,8 @@ function _M.has_subscription()
 end
 
 function _M.is_auth_required()
-    return os.getenv("ACCOUNTS_LIMIT_ACCESS") == "authenticated"
+    -- authentication is required if mode is set to "authenticated" or "subscription"
+    return os.getenv("ACCOUNTS_LIMIT_ACCESS") == "authenticated" or _M.is_subscription_required()
```
ro-tex, line 82: nit: I assume `subscription` implies a paid tier? It would be nice to add a comment to that effect.

ro-tex (approved):

peterjan (dismissed):
loving the extra variables to improve readability

docker/nginx/libs/skynet/account.lua
kwypchlo: added in https://github.com/SkynetLabs/skynet-webportal/pull/1852/commits/d0de18fcd0270a5bda3835048b83aba6c21310c4

packages/website/src/components/Uploader/Uploader.js
kwypchlo: I usually like to enforce boolean when dealing with variables that I use in jsx comparisons to avoid any mistakes that could potentially print a variable value. Although in this case I think it's safe to drop those - https://github.com/SkynetLabs/skynet-webportal/pull/1852/commits/18822edae8bb867dafbee4cc97258ea7bde74ce8

kwypchlo:
Ready for another look @meeh0w @ro-tex

packages/website/src/components/Uploader/Uploader.js
```diff
@@ -47,15 +47,37 @@ const LogInLink = () => {
...
   const [uploads, setUploads] = React.useState([]);
 
   const { data: accounts } = useAccounts();
-  const showAccountFeatures =
-    accounts && accounts.enabled !== false && !accounts?.auth_required && !accounts?.authenticated;
-  const disabledComponent =
-    accounts && accounts.enabled !== false && accounts?.auth_required && !accounts?.authenticated;
+
+  // variables extracted from useAccounts response
+  const isAccountsEnabled = accounts?.enabled;
```
peterjan, line 70: do we need to add a check to see whether accounts is undefined and handle that?

peterjan (approved):

packages/website/src/components/Uploader/Uploader.js
kwypchlo: if it's undefined then it's falsy and should serve as false here correctly

PR #1856, opened 2022-03-11
Nginx to pass Skynet-Api-Key and Authorization headers

ro-tex:
# PULL REQUEST

## Overview

In order for API keys and authorization tokens to be valid authentication mechanisms, we need Nginx to pass them to `accounts` any time it would pass the `Cookie` header.

## Checklist
 - [ ] Changelog file created


kwypchlo (changes requested):
Those functions executed with `ngx.timer.at` where you added use of `ngx.header["..."]` are async functions executed in forked threads, I am about 90% sure you need to pass the value of the headers to the caller as an argument because most of the request context (prominently `ngx.header`) is not available inside of these functions. 

Check the information here https://github.com/openresty/lua-nginx-module#ngxtimerat

I already pass some headers to some of those functions so you can take a look how it's done and follow suit.

packages/dashboard/src/services/useAnonRoute.js
```diff
@@ -13,5 +11,5 @@ const fetcher = (url, router) => {
 export default function useAnonRoute() {
   const router = useRouter();
 
-  return useSWR(`${prefix}/api/user`, (url) => fetcher(url, router));
```
kwypchlo, line 16: removed because when running dashboard from server domain, requests were hitting portal domain and cookie did not authenticate them, we need them to hit the domain that we are currently on

packages/dashboard/src/services/useAccountsApi.js
```diff
@@ -17,5 +15,5 @@ const fetcher = (url, router) => {
 export default function useAccountsApi(key, config) {
   const router = useRouter();
 
-  return useSWR(`${prefix}/api/${key}`, (url) => fetcher(url, router), config);
```
kwypchlo, line 20: removed because when running dashboard from server domain, requests were hitting portal domain and cookie did not authenticate them, we need them to hit the domain that we are currently on

packages/dashboard/src/services/accountsApi.js
```diff
@@ -1,5 +1,3 @@
 import ky from "ky";
 
-const prefix = process.env.NEXT_PUBLIC_PORTAL_DOMAIN ? `https://account.${process.env.NEXT_PUBLIC_PORTAL_DOMAIN}` : "";
-
-export default ky.create({ prefixUrl: `${prefix}/api` });
```
kwypchlo, line 5: removed prefix because when running dashboard from server domain, requests were hitting portal domain and cookie did not authenticate them, we need them to hit the domain that we are currently on

docker/nginx/nginx.conf
```diff
@@ -117,13 +117,6 @@ http {
     proxy_set_header X-Forwarded-Host  $host;
     proxy_set_header X-Forwarded-Proto $scheme;
 
-    # skynet-jwt contains dash so we cannot use $cookie_skynet-jwt
-    # https://richardhart.me/2012/03/18/logging-nginx-cookies-with-dashes/
-    map $http_cookie $skynet_jwt {
-        default '';
-        ~skynet-jwt=(?<match>[^\;]+) $match;
-    }
```
kwypchlo, line 125: removed, we don't need it any more, we have lua function for that (with unit tests too :) )

peterjan (approved):
LGTM 👍

kwypchlo (approved):
LGTM 🙄

PR #1857, opened 2022-03-13
Dashboard v2 files page

meeh0w:
# `/files` page

> As always, reviewing commit-by-commit will be easier to follow through

## Big-picture change
* Adds `/files` page. Just visual stuff and basic data fetching.
* Filtering, sorting and all the context-menu stuff is not functional yet. You'll notice a bunch of `console.log()`s as callbacks -- they'll be replaced with proper stuff once I get to implementing the above.

### Breakdown of smaller changes
* adds new icons: `Share`, `Search` and `Copy`
* re-implements `TextInputIcon` component to allow it to be more flexible
* fixes behavior out `defaultTab` prop on `Tabs` component
* styling fixes for table headers
* migrates `Switch` component to `styled-components` approach
* adds a new component called `PopoverMenu` (to serve as a context menu)
* adds new component called `CopyButton`

## Visual overview

https://user-images.githubusercontent.com/3853033/158056550-872f0188-012f-4bb8-a435-44ccb738c4cf.mov


peterjan (approved):
perfect 👍

packages/dashboard-v2/src/components/FileList/useFormattedFilesData.js
```diff
@@ -0,0 +1,26 @@
+import { useMemo } from "react";
+import prettyBytes from "pretty-bytes";
```
kwypchlo, line 2: Just a note of something I noticed, I usually try to name my imports either as official docs say or just camel case the name of the library like you did here so it's consistent across the codebase. I noticed in some past pull requests you used different names for the import of pretty-bytes and while it's not a problem or an issue, I like to keep the names of the imports consistent across the codebase and the above mentioned guidelines help me with that.

Screenshot 2022-03-16 at 10 36 03

Feel free to resolve this comment once read, it's just a note.

packages/dashboard-v2/src/components/FileList/FileList.js
```diff
@@ -0,0 +1,74 @@
...
+import { Switch } from "../Switch";
+import { TextInputIcon } from "../TextInputIcon/TextInputIcon";
+import { SearchIcon } from "../Icons";
+
+import FileTable from "./FileTable";
+import useFormattedFilesData from "./useFormattedFilesData";
+
+const FileList = ({ type }) => {
+  const isMediumScreenOrLarger = useMedia(`(min-width: ${theme.screens.md})`);
+  const { data, error } = useSWR(`user/${type}?pageSize=10`);
```
kwypchlo, line 18: I think it should be prefixed with `/` to make sure we're not appending this to some subpage

`/user/${type}?pageSize=10`

also not sure if this worked but all api calls **I think** are prefixed with `/api` so `/api/user/${type}?pageSize=10` (not sure if this comment is not relevant in some other useSWR usages too)

Screen from account.siasky.net network tab:
Screenshot 2022-03-16 at 10 48 19

kwypchlo (changes requested):

packages/dashboard-v2/src/components/FileList/FileList.js
meeh0w: Yeah, the `/api` part is added in the SWR config as a `baseUrl` (line 5), same as prefixing the key with slash (line 8):

https://github.com/SkynetLabs/skynet-webportal/blob/3128a2e7805c1d847e2591b8178fbf2acc0844fc/packages/dashboard-v2/src/lib/swrConfig.js#L4-L8

packages/dashboard-v2/src/components/FileList/useFormattedFilesData.js
meeh0w: Good point, I'll make it consistent in the upcoming PR 👍

kwypchlo (approved):

PR #1868, opened 2022-03-14
Fix video delay

ro-tex:
# PULL REQUEST

## Overview

Nginx cache removed.
Nginx rate limiting removed.

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/docker/nginx/conf.d/include/location-skylink
+++ b/docker/nginx/conf.d/include/location-skylink
@@ -11,13 +11,13 @@ set_by_lua_block $skylink { return require("skynet.skylink").parse(ngx.var.skyli
 
 # $skylink_v1 and $skylink_v2 variables default to the same value but in case the requested skylink was:
 # a) skylink v1 - it would not matter, no additional logic is executed
-# b) skylink v2 - in a lua block below we will resolve the skylink v2 into skylink v1 and update 
-#      $skylink_v1 variable so then the proxy request to skyd can be cached in nginx (proxy_cache_key 
+# b) skylink v2 - in a lua block below we will resolve the skylink v2 into skylink v1 and update
+#      $skylink_v1 variable so then the proxy request to skyd can be cached in nginx (proxy_cache_key
 #      in proxy-cache-downloads includes $skylink_v1 as a part of the cache key)
 set $skylink_v1 $skylink;
 set $skylink_v2 $skylink;
 
-# variable for Skynet-Proof header that we need to inject 
+# variable for Skynet-Proof header that we need to inject
 # into a response if the request was for skylink v2
 set $skynet_proof '';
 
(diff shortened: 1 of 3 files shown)


docker/nginx/libs/skynet/account.lua
```diff
@@ -14,6 +14,9 @@ local anon_limits = {
     ["registry"] = 250
 }
 
+-- basic caching of user data
+local curr_limits
```
ro-tex, line 18: The idea behind this one is that we're hitting accounts twice on each request just to get the user's speed limits. Those calls are cached by accounts but it's still two network calls, which is super slow.

I'm writing this comment mostly to ask if any reviewer know whether these `local` values in Lua are somehow persisted between requests. If that is going we can't do this caching in such a simple case but something tells me that it's not the case. I'll try to research that as well but anything nginx tends to spiral down into shamanism pretty quickly.

kwypchlo: This is already cached in `ngx.var.account_limits` - we're not hitting the /limits twice unless the requests come at the exactly same time because we don't have queuing here.

docker/nginx/conf.d/include/location-skylink
```diff
@@ -11,99 +11,37 @@ set_by_lua_block $skylink { return require("skynet.skylink").parse(ngx.var.skyli
...
+# variable for Skynet-Proof header that we need to inject
 # into a response if the request was for skylink v2
 set $skynet_proof '';
 
 # default download rate to unlimited
 set $limit_rate 0;
 
 access_by_lua_block {
-    -- the block below only makes sense if we are using nginx cache
-    if not ngx.var.skyd_disk_cache_enabled then
```
kwypchlo, line 29: all this code here and in another block that are protected with `ngx.var.skyd_disk_cache_enabled` shouldn't really matter because nginx doesn't access it when we have skyd cache enabled unless there is something wrong with this flag? I'd like to bring back this code to be able to limit this pull request to changes that fix the issue.

I just checked and `skyd_disk_cache_enabled` is expected to have either `"1"` or `"0"` so the check for `not ...` is definitely invalid here. It should be `if ngx.var.skyd_disk_cache_enabled == "0" then`.

reference https://github.com/SkynetLabs/skynet-webportal/pull/1762
```lua
# map skyd env variable value to "1" for true and "0" for false (expected by proxy_no_cache)
set_by_lua_block $skyd_disk_cache_enabled {
    return os.getenv("SKYD_DISK_CACHE_ENABLED") == "true" and "1" or "0"
}
```

kwypchlo (changes requested):
Description mentions that rate limiting is removed but limit_rate and limit_rate_after directives are still there.

Generally I think I am for removing all nginx cache related code since I think skyd cache is doing ok but I'd like to make a specific pull request for it so it's easier to find in future and possibly roll back just the relevant code. This is why I'd like to limit the changes to only what actually fixes the issue **or** create a separate pull request that removes only nginx skylink cache related code.

docker/nginx/libs/skynet/account.lua
```diff
@@ -78,7 +87,7 @@ function _M.has_subscription()
 end
 
 function _M.is_auth_required()
-    return os.getenv("ACCOUNTS_LIMIT_ACCESS") == "authenticated"
+    return os.getenv("ACCOUNTS_LIMIT_ACCESS") == "authenticated" or os.getenv("ACCOUNTS_LIMIT_ACCESS") == "subscription"
```
peterjan, line 90: maybe store the result of `os.getenv` in a local variable here? not sure whether it caches this value although I guess it does

docker/nginx/libs/skynet/account.lua
```diff
@@ -58,7 +66,8 @@ function _M.get_account_limits()
         end
     end
 
-    return cjson.decode(ngx.var.account_limits)
+    curr_limits = cjson.decode(ngx.var.account_limits)
```
peterjan, line 69: it seems dangerous and counterintuitive to cache the value inside of this function,
I think it's more explicit if the result would be cached and we simply don't call this
function when we already have a cached result, seems like foot-gun territory to me

docker/nginx/conf.d/include/proxy-cache-downloads
```diff
@@ -1,21 +1,12 @@
...
 
-# map skyd env variable value to "1" for true and "0" for false (expected by proxy_no_cache)
-set_by_lua_block $skyd_disk_cache_enabled {
-    return os.getenv("SKYD_DISK_CACHE_ENABLED") == "true" and "1" or "0"
-}
-
 # bypass - this will bypass cache hit on request (status BYPASS)
 # but still stores file in cache if cache conditions are met
-proxy_cache_bypass $cookie_nocache $arg_nocache $skyd_disk_cache_enabled;
+proxy_cache_bypass $cookie_nocache $arg_nocache "1";
```
peterjan, line 5: I think the `1` warrants a comment here

docker/nginx/conf.d/include/proxy-cache-downloads
```diff
@@ -1,21 +1,12 @@
...
 
 # no cache - this will ignore cache on request (status MISS)
 # and does not store file in cache under no condition
-set_if_empty $nocache "0";
+set $nocache "1";
```
peterjan, line 9: comment should be updated after updating the behaviour here

docker/nginx/conf.d/include/proxy-cache-downloads
```diff
@@ -1,21 +1,12 @@
...
 
 # disable cache when nocache is set or skyd cache is enabled
-proxy_no_cache $nocache $skyd_disk_cache_enabled;
+proxy_no_cache $nocache;
```
peterjan, line 12: do you still need a variable here? isn't `$nocache` always "1" here, I think we removed all the variables here so we can probably get rid of the `$nocache` var and just set `proxy_no_cache` directly

docker/nginx/conf.d/include/proxy-cache-downloads
peterjan: actually the `1` makes `$cookie_nocache $arg_nocache` copmletely redundant right?
this whole thing should be

```
# disable nginx cache
proxy_no_cache 1;
proxy_cache_bypass 1; 
```

peterjan (changes requested):

docker/nginx/conf.d/include/proxy-cache-downloads
ro-tex: Probably not. I was afraid to touch more than I had to but I see I might need to.

docker/nginx/conf.d/include/proxy-cache-downloads
ro-tex: actually 
```
# disable nginx cache
proxy_no_cache 1;
```
pretty much covers it.

docker/nginx/libs/skynet/account.lua
ro-tex: `os.getenv` is a method that returns a single env variable, unless I'm reading these docs incorrectly: https://www.lua.org/manual/5.1/manual.html#pdf-os.getenv

docker/nginx/libs/skynet/account.lua
ro-tex: I disagree with this assessment. Caching the data close to its use is a valid choice, especially if you don't use the cached value outside of the function. This becomes a part of the internal implementation of the function.

docker/nginx/conf.d/include/proxy-cache-downloads
peterjan: you also want the `proxy_cache_bypass 1; ` just in case it is in the cache for whatever reason to be on the safe side

PR #1880, opened 2022-03-19
replace caddy with certbot

kwypchlo:
work in progress 

- replace [caddy](https://caddyserver.com/) with [certbot](https://certbot.eff.org/)
  - one benefit is that certbot is much more mature and widely used so potentially less buggy
  - other benefit is that it's easier for portal operators to change dns provider, with caddy it was not configurable because we had to include the dns plugin in caddy Dockerfile build and in configuration json while with certbot portal operators can replace the existing [certbot/dns-route53](https://hub.docker.com/r/certbot/dns-route53) image with one of the other [certbot](https://hub.docker.com/r/certbot/certbot) images and adjust `CERTBOT_ARGS` env variable accordingly
- creates bundled SAN certificate with all domains in one
  - benefit is that letsencrypt gives every unique bundle a separate rate limit so if we bundle portal domain and server domain on one certificate, we never run out of quota since every server requests only one certificate and every server has unique server domain
  - on a negative side is that this certificate is supposed to be a bit larger than one domain cert (TODO: check what's the difference)

Also this fixes nginx not reloading certificates by adding a configuration reload command that runs every 6 hours but that's really not in the scope of this feature.


docker/nginx/conf.d/include/ssl-settings
```diff
@@ -1,7 +1,10 @@
 # https://ssl-config.mozilla.org/#server=nginx&version=1.17.7&config=intermediate&openssl=1.1.1d&hsts=false&ocsp=false&guideline=5.6
 
+ssl_certificate /etc/letsencrypt/live/skynet-portal/fullchain.pem;
+ssl_certificate_key /etc/letsencrypt/live/skynet-portal/privkey.pem;
+
 ssl_session_timeout 1d;
-ssl_session_cache shared:MozSSL:10m;  # about 40000 sessions
+ssl_session_cache shared:MozSSL:10m; # about 40000 sessions
```
peterjan, line 7: is there a reason to keep it low-ish?

docker/nginx/Dockerfile
```diff
@@ -18,5 +18,6 @@ CMD [ "bash", "-c", \
        ./mo < /etc/nginx/conf.d.templates/server.api.conf > /etc/nginx/conf.d/server.api.conf; \
        ./mo < /etc/nginx/conf.d.templates/server.hns.conf > /etc/nginx/conf.d/server.hns.conf; \
        ./mo < /etc/nginx/conf.d.templates/server.skylink.conf > /etc/nginx/conf.d/server.skylink.conf ; \
+       while :; do sleep 6h & wait ${!}; /usr/local/openresty/bin/openresty -s reload; done & \
```
peterjan, line 21: I assume this is a standard (best) practice? does is unclean right? e.g. what happens if I'm watching a movie and openresty reloads

docker/nginx/conf.d/include/ssl-settings
kwypchlo: I certainly don't know, that's the default I got from https://ssl-config.mozilla.org/

docker/nginx/Dockerfile
kwypchlo: Yeah, it's safe to run `-s reload` running nginx 👍 

> Once the master process receives the signal to reload configuration, it checks the syntax validity of the new configuration file and tries to apply the configuration provided in it. If this is a success, the master process starts new worker processes and sends messages to old worker processes, requesting them to shut down. Otherwise, the master process rolls back the changes and continues to work with the old configuration. Old worker processes, receiving a command to shut down, stop accepting new connections and continue to service current requests until all such requests are serviced. After that, the old worker processes exit.

http://nginx.org/en/docs/beginners_guide.html

ro-tex (commented):
LGTM

docker/certbot/entrypoint.sh
```diff
@@ -0,0 +1,55 @@
...
+    # --email: required for generating certificates, used for communication with CA
+    # --domains: comma separated list of domains (will generate one bundled SAN cert)
+    # Use CERTBOT_ARGS env variable to pass any additional arguments, ie --dns-route53
+    certbot certonly \
+        --non-interactive --agree-tos --cert-name skynet-portal \
+        --email ${EMAIL_ADDRESS} --domains ${DOMAINS} ${CERTBOT_ARGS}
+
+    # Run a background sleep process that counts down given time
+    # Certbot docs advise running maintenance process every 12 hours
+    sleep 12h &
```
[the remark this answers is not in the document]
kwypchlo: is to execute every 12 hours

we could put it in crontab but crontab runs in a separate process and any critical errors will get silenced (instead of restarting container) or we could shut down container while certificates are renewing and certbot wouldn't have a chance to clean up - running process in the foreground allows us to better handle those cases

ChrisSchinnerl (approved):

PR #1911, opened 2022-03-22
Dashboard v2 upgrade page

meeh0w:
## Overview
* Some styling changes to generic components (icons, `Button`, `Slider`)
* ✨  New page -> `/upgrade`
* `Choose` / `Upgrade` buttons don't link anywhere yet.

## Visual overview
https://user-images.githubusercontent.com/3853033/159486266-1fbcdd1b-fa48-4129-8fdd-64794fd85a9f.mov

### DIFF
--- a/packages/dashboard-v2/src/components/Button/Button.js
+++ b/packages/dashboard-v2/src/components/Button/Button.js
@@ -1,3 +1,4 @@
+import cn from "classnames";
 import PropTypes from "prop-types";
 import styled from "styled-components";
 
@@ -6,17 +7,12 @@ import styled from "styled-components";
  */
 export const Button = styled.button.attrs(({ disabled, $primary }) => ({
   type: "button",
-  className: `px-6 py-2.5 rounded-full font-sans uppercase text-xs tracking-wide transition-[filter]
-            ${
-              $primary
-                ? `bg-primary ${disabled ? "" : "text-palette-600"}`
-                : `bg-white border-2 ${disabled ? "border-palette-300" : "border-black text-palette-600"}`
-            }
-            ${
-              disabled
-                ? `${$primary ? "saturate-50 brightness-125 text-palette-400" : "text-palette-300"} cursor-default`
-                : "hover:brightness-90"
-            }`,
+  className: cn("px-6 py-2.5 rounded-full font-sans uppercase text-xs tracking-wide transition-[opacity_filter]", {
+    "bg-primary text-palette-600": $primary,
+    "bg-white border-2 border-black text-palette-600": !$primary,
+    "cursor-not-allowed opacity-60": disabled,
+    "hover:brightness-90": !disabled,
+  }),
 }))``;
 Button.propTypes = {
   /**
--- a/packages/dashboard-v2/src/components/Button/Button.js
+++ b/packages/dashboard-v2/src/components/Button/Button.js
@@ -6,9 +6,17 @@ import styled from "styled-components";
  */
 export const Button = styled.button.attrs(({ disabled, $primary }) => ({
   type: "button",
-  className: `px-6 py-2.5 rounded-full font-sans uppercase text-xs tracking-wide text-palette-600 transition-[filter]
-            ${$primary ? "bg-primary" : "bg-white border-2 border-black"}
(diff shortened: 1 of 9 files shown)


packages/dashboard-v2/src/components/Button/Button.js
```diff
@@ -6,9 +6,17 @@ import styled from "styled-components";
...
+            ${
+              $primary
+                ? `bg-primary ${disabled ? "" : "text-palette-600"}`
+                : `bg-white border-2 ${disabled ? "border-palette-300" : "border-black text-palette-600"}`
+            }
+            ${
+              disabled
+                ? `${$primary ? "saturate-50 brightness-125 text-palette-400" : "text-palette-300"} cursor-default`
+                : "hover:brightness-90"
+            }`,
```
kwypchlo, line 19: this block looks quite unreadable at least from github diff perspective - do you think there is a good way to make it more readable?

packages/dashboard-v2/src/pages/upgrade.js
```diff
@@ -0,0 +1,132 @@
...
+        <p>Something went wrong, please try again later.</p>
+      </div>
+    );
+  }
+
+  return (
+    <div className="w-full mb-24">
+      {!loading && (
+        <Slider
+          slides={plans.map((p) => {
```
kwypchlo, line 85: It helps readability to use more verbose names for variables

```suggestion
          slides={plans.map((plan) => {
```

packages/dashboard-v2/src/pages/upgrade.js
```diff
@@ -0,0 +1,132 @@
...
+  <div className="my-8 text-center">
+    <h2>
+      <sup className="text-lg -top-4">$</sup>
+      {price}
+    </h2>
+    <p className="uppercase text-sm font-light -mt-2">per month</p>
+  </div>
+);
+
+const bandwidth = (value) => parseFloat(bytes(value, { bits: true, binary: true }));
```
kwypchlo, line 62: why do you use parseFloat on this ?

kwypchlo (dismissed):

packages/dashboard-v2/src/pages/upgrade.js
meeh0w: So, our limits are _not_ returned in accordance to [SI](https://en.wikipedia.org/wiki/International_System_of_Units#Prefixes) -- we multiply by 1024, not 1000.

Now, `pretty-bytes` allows us to pass a `binary` flag which deals with that, but then it also changes the unit name.
For example, for `downloadBandwidth` we may return `83886080`, which translates either into `83.9 Mbit` (with `binary: false`) or `80 Mibit` (with `binary: true`).

So to get a nice, round value here I'm using the `binary` flag, but also `parseFloat` to get rid of the `Mibit` suffix.

Even though technically `Mibit` is correct, it will confuse users who are used to seeing `Mbps`.

We could fix that by adding a tooltip to the unit. Something like this, but with a nicer tooltip maybe? WDYT?
image

packages/dashboard-v2/src/components/Button/Button.js
meeh0w: Managed to get a similar effect using `opacity` for both variants of the button:
* 18600533

packages/dashboard-v2/src/pages/upgrade.js
meeh0w: Fixed along with `Button`'s classes, in 18600533

packages/dashboard-v2/src/pages/upgrade.js
kwypchlo: I had a meeting with @ro-tex and @ChrisSchinnerl and we're gonna change the values in accounts to base-10 to reflect the actual limits. You can drop parseFloat for the time being and once the accounts changes are in, we will drop `binary: true` flag too and it should reflect correct value.

kwypchlo: ok this is already merged in https://github.com/SkynetLabs/skynet-accounts/pull/168

meeh0w: c964086f

peterjan (approved):

kwypchlo (approved):

PR #1913, opened 2022-03-23
Dashboard v2 auth pages login

meeh0w:
## Overview
* adds `AuthLayout` that will be re-used by `/auth/*` namespace, i.e. login / sign up / recover pages
* adds `ky` (`fetch` wrapper), `yup` (data validation) and `formik` (form builder) as dependencies
* implements `/auth/login` page

## 🎥  Visual overview

https://user-images.githubusercontent.com/3853033/159683485-78b0bb2d-f097-44f1-a7c1-8ee1ed54368e.mov


packages/dashboard-v2/README.md
```diff
@@ -8,3 +8,18 @@ This is a Gatsby application. To run it locally, all you need is:
 
 - `yarn install`
 - `yarn start`
+
+## Accessing remote APIs
+
+To be able to log in on a local environment with your production credentials, you'll need to make the browser believe you're actually on the same domain, otherwise browser will block the session cookie.
```
peterjan, line 14:
```suggestion
To be able to log in on a local environment with your production credentials, you'll need to make the browser believe you're actually on the same domain, otherwise the browser will block the session cookie.
```

packages/dashboard-v2/src/components/forms/LoginForm.js
```diff
@@ -0,0 +1,88 @@
...
+        });
+
+        onSuccess();
+      } catch (err) {
+        if (err.response) {
+          const data = await err.response.json();
+
+          if (INVALID_CREDENTIALS_ERRORS.includes(data.message)) {
+            setErrors({
+              email: "Invalid e-mail address or password",
```
peterjan, line 39: we should use consistent spelling, should be email

packages/dashboard-v2/src/components/forms/LoginForm.js
```diff
@@ -0,0 +1,88 @@
...
+              password: "Invalid e-mail address or password",
```
peterjan, line 40: wondering why the error says "or" here, can't it be 
```
email: "Invalid email address"
password: "Invalid password"
```

peterjan (approved):

packages/dashboard-v2/src/components/NavBar/NavBar.js
```diff
@@ -77,9 +78,12 @@ export const NavBar = () => (
               partiallyActive
             />
             <DropdownMenuLink
-              to="/logout"
-              as={Link}
+              onClick={async () => {
+                await accountsService.post("logout");
+                navigate("/auth/login");
```
kwypchlo, line 83: I would probably leave a todo here that says to add some kind of toast message if it fails

packages/dashboard-v2/src/components/forms/LoginForm.js
```diff
@@ -0,0 +1,88 @@
...
+    validationSchema={loginSchema}
+    onSubmit={async (values, { setErrors }) => {
+      try {
+        await accountsService.post("login", {
+          json: values,
+        });
+
+        onSuccess();
+      } catch (err) {
+        if (err.response) {
```
kwypchlo, line 34: we should handle a case where err.response is empty too, that's a timeout for example, we should display a toast or sth like that (can be in follow up)

packages/dashboard-v2/src/components/forms/LoginForm.js
kwypchlo: probably that's accounts not returning a specific error right? we should actually display the message from accounts I think since it can also say "registration is disabled at the moment" or other

kwypchlo (changes requested):

packages/dashboard-v2/src/components/forms/LoginForm.js
meeh0w: API actually does return separate messages for "user not found" and "password mismatch", but it's not a good practice to do so. API should only return something like "invalid credentials", without telling if the user exists in the database, so that it's not possible to extract a list of our users or check if somebody is using the service ([reference](https://tarunbatra.com/blog/security/Error-messages-in-login-process-Privacy-and-Security/#Protecting-user-base-information)). I posted [an issue](https://github.com/SkynetLabs/skynet-accounts/issues/171) in `accounts`.

As for the UI, I'm setting a generic message for that very reason -- but I guess it could be set on one of the fields only (or entirely below the form).

packages/dashboard-v2/src/components/forms/LoginForm.js
meeh0w: Good point. Here are some examples after fixes:

* [invalid credentials](https://user-images.githubusercontent.com/3853033/159941381-16f6e0b5-04ce-453f-936b-198b1b060b6c.png)
* [connectivity issues - offline, timeout](https://user-images.githubusercontent.com/3853033/159941511-c645ca96-d617-47a7-afbe-2939378e61f7.png)

packages/dashboard-v2/src/components/NavBar/NavBar.js
meeh0w: 👍

packages/dashboard-v2/package.json
```diff
@@ -10,6 +10,7 @@
   "scripts": {
     "develop": "gatsby develop",
     "start": "gatsby develop",
+    "secure": "gatsby develop --https --host=local.skynetpro.net -p=443",
```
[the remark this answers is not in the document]
meeh0w: 👍

packages/dashboard-v2/src/components/forms/LoginForm.js
meeh0w: Fixed 👍

packages/dashboard-v2/src/components/forms/LoginForm.js
meeh0w: Addressed in 731b1b6d

packages/dashboard-v2/src/components/forms/LoginForm.js
meeh0w: I addressed this and other comments in 731b1b6d

kwypchlo (approved):

peterjan (approved):
good stuff!

PR #1914, opened 2022-03-23
Dashboard v2 - registration and account recovery flows

meeh0w:
## Overview
* Adds account registration flow
* Adds account recovery flow

### Registration & recovery flows
https://user-images.githubusercontent.com/3853033/159711757-7c8a36e4-3d70-458e-b1c9-1e1a174d17c6.mov

### When email is occupied
Dashboard V2 - email in use


packages/dashboard-v2/src/components/forms/SignUpForm.js
```diff
@@ -0,0 +1,109 @@
...
+import { Formik, Form } from "formik";
+import * as Yup from "yup";
+
+import { TextField } from "../Form/TextField";
+import { Button } from "../Button";
+
+import accountsService from "../../services/accountsService";
+
+// TODO: better password strength validation
+const registrationSchema = Yup.object().shape({
```
meeh0w, line 11: I haven't introduced any specific password policy at this point, as I think it's worth discussing.

The password policy on our current dashboard is pretty loose (allows stuff like `123456`) -- I was wondering if we want to introduce something more strict. Maybe something like [this](https://stackoverflow.com/questions/55451304/formik-yup-password-strength-validation-with-react)?

packages/dashboard-v2/src/pages/auth/reset-password.js
```diff
@@ -0,0 +1,50 @@
...
+      <div className="mb-4 md:mb-16">
+        <img src="/images/logo-black-text.svg" alt="Skynet" />
+      </div>
+      {state !== State.Success && (
+        <RecoveryForm onSuccess={() => setState(State.Success)} onFailure={() => setState(State.Failure)} />
+      )}
+
+      {state === State.Success && (
+        <p className="text-primary text-center font-semibold">
+          Please check your email inbox for further instructions.
```
peterjan, line 28:
```suggestion
          Please check your inbox for further instructions.
```

packages/dashboard-v2/src/pages/auth/signup.js
```diff
@@ -0,0 +1,56 @@
...
+      <div className="mb-4 md:mb-16">
+        <img src="/images/logo-black-text.svg" alt="Skynet" />
+      </div>
+      {state !== State.Success && (
+        <SignUpForm onSuccess={() => setState(State.Success)} onFailure={() => setState(State.Failure)} />
+      )}
+
+      {state === State.Success && (
+        <div className="text-center">
+          <p className="text-primary font-semibold">Please check your email inbox and confirm your email address.</p>
```
peterjan, line 37:
```suggestion
          <p className="text-primary font-semibold">Please check your inbox and confirm your email address.</p>
```

packages/dashboard-v2/src/components/forms/ResetPasswordForm.js
```diff
@@ -0,0 +1,72 @@
...
+import { Formik, Form } from "formik";
+import * as Yup from "yup";
+
+import { TextField } from "../Form/TextField";
+import { Button } from "../Button";
+
+import accountsService from "../../services/accountsService";
+
+const resetPasswordSchema = Yup.object().shape({
+  password: Yup.string().required("Password is required").min(6, "Password has to be at least 6 characters long"),
```
peterjan, line 11: can't immediately find where, but I remember reviewing a  schema where password had no requirements, in any case this schema is probably duplicated in a couple of places so it would be best to deduplicate this and define the password requirements in a single place

peterjan (dismissed):

packages/dashboard-v2/src/pages/user/confirm.js
```diff
@@ -0,0 +1,68 @@
...
+      } catch {
+        setState(State.Failure);
+      }
+    }
+
+    if (token) {
+      confirm(token);
+    }
+
+    return () => clearTimeout(timer);
```
kwypchlo, line 41: This can run with timer being undefined if it rerenders with different token before `accountsService.get` resolves - probably totally unlikely but still a good practice to fix that I think.

I would make the hook return function async, assign `const confirming = await confirm()` and `await confirmation`  in the return function to make sure that `timer` is set before we clear it.

meeh0w: Yeah, extremely unlikely given the `token` comes from `location` :)

But better be on the safe side I guess.
I fixed it by adding an `AbortController`. Now if the token changes we'll cancel the previous request. Cancelling a request raises an error, so the `timer` will never be set (which means we don't need to clear it).

👉  be82050b

packages/dashboard-v2/src/pages/auth/signup.js
meeh0w: Fixed in c7b8c31d

packages/dashboard-v2/src/pages/auth/reset-password.js
meeh0w: Fixed in c7b8c31d

packages/dashboard-v2/src/components/forms/SignUpForm.js
meeh0w: No policy there on the server side, it just can't be empty.

I updated & unified the policies between `SignUpForm` and `ResetPasswordForm`: 6b733ab7

packages/dashboard-v2/src/components/forms/ResetPasswordForm.js
```diff
@@ -0,0 +1,72 @@
...
+  confirmPassword: Yup.string().oneOf([Yup.ref("password"), null], "Passwords must match"),
```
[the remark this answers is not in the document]
meeh0w: Yup.

kwypchlo (approved):

peterjan (approved):

PR #1915, opened 2022-03-24
Dashboard V2  - changing account information

meeh0w:
## Overview
* Adds `Alert` component (message box for success/error messages)
* Adds `Modal` component
* Implements data mutations on user account screen (changing email / password, deleting account)
* Removes avatar uploader and display name, as they're not available yet.

## 🎥  Visuals
https://user-images.githubusercontent.com/3853033/159890517-722aa1d1-3b2b-4a09-ab6e-ac1daab4bfd5.mov

### DIFF
--- a/packages/dashboard-v2/src/components/forms/AccountSettingsForm.js
+++ b/packages/dashboard-v2/src/components/forms/AccountSettingsForm.js
@@ -37,15 +37,17 @@ export const AccountSettingsForm = ({ user, onSuccess, onFailure }) => {
       validationSchema={emailUpdateSchema}
       onSubmit={async ({ email, password }, { resetForm }) => {
         try {
-          await accountsService.put("user", {
-            json: {
-              email: email || undefined,
-              password: password || undefined,
-            },
-          });
+          const user = await accountsService
+            .put("user", {
+              json: {
+                email: email || undefined,
+                password: password || undefined,
+              },
+            })
+            .json();
 
           resetForm();
-          onSuccess();
+          await onSuccess(user);
         } catch {
           onFailure();
         }
--- a/packages/dashboard-v2/src/components/forms/AccountSettingsForm.js
+++ b/packages/dashboard-v2/src/components/forms/AccountSettingsForm.js
@@ -1,0 +1,106 @@
+import * as Yup from "yup";
+import PropTypes from "prop-types";
+import { Formik, Form } from "formik";
+
+import { Button } from "../Button";
+import { TextField } from "../Form/TextField";
+import accountsService from "../../services/accountsService";
+
+const isPopulated = (value) => value?.length > 0;
+
+const emailUpdateSchema = Yup.object().shape({
+  email: Yup.string().email("Please provide a valid email address"),
+  confirmEmail: Yup.string()
+    .oneOf([Yup.ref("email"), null], "Emails must match")
+    .when("email", {
+      is: isPopulated,
+      then: (schema) => schema.required("Please confirm new email address"),
+    }),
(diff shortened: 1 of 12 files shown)


peterjan (approved):
LGTM 👍

packages/dashboard-v2/src/components/forms/AccountSettingsForm.js
```diff
@@ -0,0 +1,106 @@
...
+        confirmEmail: "",
+        password: "",
+        confirmPassword: "",
+      }}
+      validationSchema={emailUpdateSchema}
+      onSubmit={async ({ email, password }, { resetForm }) => {
+        try {
+          await accountsService.put("user", {
+            json: { email, password },
+          });
```
kwypchlo, line 42: should we only send email / password if it's supplied? it seems like it would send empty string instead of email if we don't supply it ?

packages/dashboard-v2/src/pages/settings/index.js
```diff
@@ -26,51 +34,38 @@ const AccountPage = () => {
...
-              </small>
-            </FormGroup>
+            {state === State.Failure && (
+              <Alert $variant="error">There was an error processing your request. Please try again later.</Alert>
+            )}
+            {state === State.Success && <Alert $variant="success">Changes saved successfully.</Alert>}
+            <AccountSettingsForm
+              user={user}
+              onSuccess={() => {
+                reloadUser();
```
kwypchlo, line 45: do we want to `await` reload ?

packages/dashboard-v2/src/components/forms/AccountSettingsForm.js
meeh0w: Yes, but blank values seem to be ignored by the API right now. That said, emails will become optional, so here's a bulletproofing patch: 52314130

packages/dashboard-v2/src/pages/settings/index.js
meeh0w: It won't hurt :) 2aa3437a

kwypchlo (approved):

packages/dashboard-v2/src/pages/settings/index.js
kwypchlo: we probably want to also `catch` potential bad reload since otherwise we'll never change to success state

meeh0w: Fair point. Should even feel a bit snappier now: 5bb1ed18.

kwypchlo (approved):

packages/dashboard-v2/src/components/forms/AccountSettingsForm.js
```diff
@@ -0,0 +1,111 @@
...
+      is: isPopulated,
+      then: (schema) => schema.required("Please confirm new email address"),
+    }),
+  password: Yup.string().min(1, "Password can't be blank"),
+  confirmPassword: Yup.string()
+    .oneOf([Yup.ref("password"), null], "Passwords must match")
+    .when("password", {
+      is: isPopulated,
+      then: (schema) => schema.required("Please confirm new password"),
+    }),
```
ro-tex, line 25: When updating the user's profile, we're fine with them not supplying a password - their password will stay the same. I'm not sure I'm reading this right but if I am, we are requiring a password change here.

Also, it should be fine for the user to not have an email.

ro-tex (approved):

packages/dashboard-v2/src/components/forms/AccountSettingsForm.js
kwypchlo: I created new issue with your comment to follow up.

PR #1916, opened 2022-03-24
insert snippet for docker image developement builds

kwypchlo:
With images for some of our services replacing local dockerfile builds, we lost an option to run a specific branch of that service. The use case of that is for example when skynet-accounts creates a branch with new feature and would like to have it deployed and tested on a dev server before merging in. Turns out docker compose [supports](https://github.com/docker/compose/pull/2430) git url in "build" key and it would pull specific branch of that repository and use it to build the image locally.

### DIFF
--- a/docker-compose.abuse-scanner.yml
+++ b/docker-compose.abuse-scanner.yml
@@ -8,6 +8,8 @@ x-logging: &default-logging
 
 services:
   abuse-scanner:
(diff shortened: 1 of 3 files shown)


ChrisSchinnerl (approved):

peterjan (approved):
this is such an awesome quick win, love it

PR #1917, opened 2022-03-24
use skynet-accounts image instead of local dockerfile

kwypchlo:
We want to use [skynetlabs/skynet-accounts](https://hub.docker.com/r/skynetlabs/skynet-accounts) image instead of building skynet-accounts locally.

@firyx this pull request also requires changes to https://github.com/SkynetLabs/ansible-playbooks because we no longer want to set skynet-accounts version in docker compose override or give it any special treatment since a portal will already set the correct image version when tagging a release, see example for blocker service on latest deploy tag: https://github.com/SkynetLabs/skynet-webportal/blob/deploy-2022-03-22/docker-compose.blocker.yml#L16

Please hold up with merging this pull request until https://github.com/SkynetLabs/ansible-playbooks/pull/276 is merged in.


firyx:
It can be merged when https://github.com/SkynetLabs/ansible-playbooks/pull/277 is merged.

firyx (approved):

peterjan (approved):

PR #1918, opened 2022-03-25
Dashboard v2 api keys mutations

meeh0w:
## Overview
* `info` variant for `Alert` component
* styling adoptions for `Modal` (setting max-width, wide modals look weird)
* integration with `/user/apikeys` endpoint - setting up general & public api keys, mutating `skylinks` list for public keys
* `name` is being sent to backend, but it's not deployed yet (has no effect), so in the videos I'm not even setting it.

### Happy path

https://user-images.githubusercontent.com/3853033/160085165-b8f6e22f-0b06-475f-b938-d661ab5a3536.mov

#### Error handling (timeouts, connection problems)

https://user-images.githubusercontent.com/3853033/160085226-d906c564-17bd-4a6f-be35-d5705be5ff0f.mov


packages/dashboard-v2/src/components/APIKeyList/useAPIKeyRemoval.js
```diff
@@ -0,0 +1,39 @@
...
+    try {
+      await accountsService.delete(`user/apikeys/${key.id}`);
+      setRemovalInitiated(false);
+      onSuccess();
+    } catch (err) {
+      let message = "There was an error processing your request. Please try again later.";
+
+      if (err.response) {
+        const response = await err.response.json();
+
```
peterjan, line 22: `nit` extra newline

packages/dashboard-v2/src/components/forms/AddAPIKeyForm.js
```diff
@@ -0,0 +1,114 @@
...
+      )}
+      <Formik
+        initialValues={{
+          name: "",
+        }}
+        validationSchema={newAPIKeySchema}
+        onSubmit={async ({ name }, { resetForm }) => {
+          try {
+            const { key } = await accountsService
+              .post("user/apikeys", {
```
peterjan, line 62: maybe it's worthwhile turning this into a hook, kinda weird seeing post requests like this in `onSubmit` handlers directly

packages/dashboard-v2/src/pages/settings/api-keys.js
```diff
@@ -1,63 +1,97 @@
...
-              <div className="flex mt-2 sm:mt-0 justify-center">
-                <Button onClick={() => console.info("TODO: generate ky")}>Generate</Button>
-              </div>
+
+          <section className="flex flex-col gap-2">
+            <h5>General keys</h5>
+            <p className="text-palette-500">
+              Give full access to <b>Accounts</b> service and are equivalent to using a JWT token.
+            </p>
+            <p className="underline">This type of API keys need to be kept secret and never shared with anyone.</p>
```
peterjan, line 72:
```suggestion
            <p className="underline">This type of API keys need to be kept secret and should never be shared with anyone.</p>
```
```suggestion
            <p className="underline">This type of API keys need to be kept secret and are never to be shared with anyone.</p>
```

packages/dashboard-v2/src/pages/settings/api-keys.js
```diff
@@ -1,63 +1,97 @@
...
-import { Button } from "../../components/Button";
-import { TrashIcon } from "../../components/Icons";
+import { AddAPIKeyForm, APIKeyType } from "../../components/forms/AddAPIKeyForm";
+import { APIKeyList } from "../../components/APIKeyList/APIKeyList";
+import { Alert } from "../../components/Alert";
 
 const APIKeysPage = () => {
-  const { data: apiKeys } = useSWR("user/apikeys");
+  const { data: apiKeys = [], mutate: reloadKeys, error } = useSWR("user/apikeys");
+  const generalKeys = apiKeys.filter(({ public: isPublic }) => isPublic === "false");
```
peterjan, line 12: why rename?

packages/dashboard-v2/src/components/forms/AddSkylinkToAPIKeyForm.js
```diff
@@ -0,0 +1,73 @@
...
+import { Formik, Form } from "formik";
+
+import accountsService from "../../services/accountsService";
+
+import { Button } from "../Button";
+import { TextField } from "../Form/TextField";
+import { CircledProgressIcon, PlusIcon } from "../Icons";
+
+const newSkylinkSchema = Yup.object().shape({
+  skylink: Yup.string().required("Provide a valid Skylink"), // TODO: Comprehensive Skylink validation
```
peterjan, line 12: I think a length check should suffice so it's probably easy to do here in this PR.  
Probably you also want to take into account the `sia://` prefix or the portal url prefixes.

peterjan (approved):

packages/dashboard-v2/src/pages/settings/api-keys.js
meeh0w: `public` is a reserved keyword :(

kwypchlo:
1. instead of api key id we should display either a name that the user assigned or "no name" if no name has been assigned (can be some sort of italic and lighter colour maybe)
Screenshot 2022-03-25 at 10 38 19

2. that border around skylink in the modal looks like it's an input, I think we can drop the border
Screenshot 2022-03-25 at 10 43 52

2. the input for pasting a skylink should have a placeholder ie. "paste skylink here" or sth of that sort (you might want to ask @dghelm for exact text, I'm usually pretty bad in copy :) )
Screenshot 2022-03-25 at 10 43 52

3. for public api keys, we should add a red note like "no skylinks set yet" because people will try to use them otherwise not noticing that those work with only skylinks specified, if some skylinks are specified, we should put a number of skylinks that the api key covers

4. @dghelm I'd like you to look at the copy in there in general, I think we could benefit from a bit more verbosity

packages/dashboard-v2/src/components/APIKeyList/useAPIKeyRemoval.js
```diff
@@ -0,0 +1,39 @@
+import { useCallback, useState } from "react";
+import accountsService from "../../services/accountsService";
+
+export const useAPIKeyRemoval = ({ key, onSuccess }) => {
+  const [removalInitiated, setRemovalInitiated] = useState(false);
+  const [removalError, setRemovalError] = useState(null);
+
+  const prompt = () => setRemovalInitiated(true);
```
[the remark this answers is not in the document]
meeh0w: Sure :)

packages/dashboard-v2/src/components/forms/AddAPIKeyForm.js
meeh0w: Good idea! 👉  614b7791

packages/dashboard-v2/src/components/forms/AddSkylinkToAPIKeyForm.js
meeh0w: I found `parseSkylink` in `skynet-js` :) It's gonna help both with validation and accepting `sia://` and portal URLs as skylinks to be pinned: 9dd4b8eb

meeh0w:
@kwypchlo 

1 - 4. Done! ✅  08bab146
https://user-images.githubusercontent.com/3853033/160114708-e9be64ad-7f50-45f1-ac8a-4ec1ee321d4b.mov

5. I'll wait for @dghelm to chime in with his thoughts on the copies here, I'm not an expert as well :)

meeh0w:
@kwypchlo I hope this is looking better ;)

https://user-images.githubusercontent.com/3853033/160161988-be4415e6-380d-4dab-abea-bc5a66933b16.mov

kwypchlo:
> @kwypchlo I hope this is looking better ;)
> 
>  Dashboard.V2.-.improved.public.keys.creation.mov

I was actually thinking you would move the creation of api key into the modal, the same that you make edits on. I think it looks a bit cluttered right now where you have a form for creating them on the page, then below you have list of api keys and then below you would need to have another form for general usage api keys if we want to be consistent and another list of api keys.

The text below the api key looks fantastic. Only think I would change is to rename `0 skylinks configured` to `No skylinks configured`. Also I think word "skylink" shouldn't be capitalised but I'll defer to @dghelm

@meeh0w we can also experiment with this in a follow up, from feature perspective it's all ok so I don't want to slow down the initial release - what do you think?

pcfreak30:
` Also I think word "skylink" shouldn't be capitalised but I'll defer to @dghelm` I agree on that :) @kwypchlo @meeh0w.

IMHO, it would make sense to create an api key, but then be able to just add in skylinks so its all one UI submission, not multiple steps (more friction/clicks), then have an apply/create button. After you can have a modal version for just the skylinks and a separate save button in the modal for the name if changeable after.

Thats my 2c, hope it helps :)

meeh0w:
@kwypchlo @pcfreak30 I changed `Skylinks` to `skylinks` :) About the modal, let's do it in the next phase of UI improvements 👍

kwypchlo (approved):

dghelm:
> @kwypchlo @pcfreak30 I changed `Skylinks` to `skylinks` :) About the modal, let's do it in the next phase of UI improvements 👍

I was going to look into the copy this morning, but I can wait until we have a full deploy on `dev`, if that works for everyone else. Then Ops can just do a full pass.

peterjan (approved):

Issue #1946, opened 2022-03-28
`skynet-portal-api` and `skynet-server-api` set to "https://" on files served using DNSLink


dghelm:
## Describe the bug
Sites using DNSLink on siasky.net (like siasky.tools and test.homescreen-beta.app) return incorrect headers for `skynet-portal-api` and `skynet-server-api`

This breaks `skynet-js` on these sites because the client looks to those values when initializing the client.

## To Reproduce

1. Open dev tools
2. Visit siasky.tools or homescreen-beta.app
3. see header values

![image](https://user-images.githubusercontent.com/4205837/160446905-4fafd822-4ff4-47c3-b8e9-4fc117c34255.png)

## Expected behavior
These should be `https://siasky.net` and `https://[sever].siasky.net`

PR #1947, opened 2022-03-29
ensure portal and server api headers on dnslink server

kwypchlo:
Fixes missing skynet-server-api and skynet-portal-api headers on dnslink server.

closes https://github.com/SkynetLabs/skynet-webportal/issues/1946

verification: https://homescreen-beta.app/ - skynet-server-api and skynet-portal-api headers are available

### DIFF
--- a/docker/nginx/conf.d.templates/server.dnslink.conf
+++ b/docker/nginx/conf.d.templates/server.dnslink.conf
@@ -12,5 +12,14 @@ server {
     ssl_certificate     /etc/ssl/local-certificate.crt;
     ssl_certificate_key /etc/ssl/local-certificate.key;
 
+    set_by_lua_block $skynet_portal_domain { return "{{PORTAL_DOMAIN}}" }
+    set_by_lua_block $skynet_server_domain {
+        -- fall back to portal domain if server domain is not defined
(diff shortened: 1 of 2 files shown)


meeh0w (approved):

docker/nginx/conf.d.templates/server.dnslink.conf
```diff
@@ -12,5 +12,14 @@ server {
     ssl_certificate     /etc/ssl/local-certificate.crt;
     ssl_certificate_key /etc/ssl/local-certificate.key;
 
+    set_by_lua_block $skynet_portal_domain { return "{{PORTAL_DOMAIN}}" }
+    set_by_lua_block $skynet_server_domain {
+        -- fall back to portal domain if server domain is not defined
```
peterjan, line 17: why do we want this fallback, seems like it just obfuscates the problem a bunch

peterjan (approved):

docker/nginx/conf.d.templates/server.dnslink.conf
kwypchlo: that's because we have a requirement that SERVER_DOMAIN should be empty if it's one-server setup, in this case PORTAL_DOMAIN is the SERVER_DOMAIN

PR #1952, opened 2022-03-31
Lint dockerfiles with hadolint

kwypchlo:
Lint dockerfile with https://github.com/hadolint/hadolint-action

ref. https://github.com/hadolint/hadolint

local scan: `docker run --rm -i hadolint/hadolint < Dockerfile`

### DIFF
--- a/packages/health-check/Dockerfile
+++ b/packages/health-check/Dockerfile
@@ -1,16 +1,15 @@
 FROM node:16.14.2-alpine
 
-RUN apk update && apk add dnsmasq
+RUN apk --no-cache add dnsmasq=2.86-r0 && rm -rf /var/cache/apk/*
 
 WORKDIR /usr/app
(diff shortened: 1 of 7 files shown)


packages/health-check/Dockerfile
```diff
@@ -1,16 +1,16 @@
 FROM node:16.14.2-alpine
 
-RUN apk update && apk add dnsmasq
+RUN apk add --no-cache dnsmasq=2.86-r0 && \
```
meeh0w, line 3: Isn't the point of using `--no-cache` _not_ having to remove the cache afterwards?

kwypchlo: right, fixed, thanks!

MSevey (approved):

meeh0w (approved):

PR #1956, opened 2022-03-31
Composite GitHub Actions

MSevey:
# PULL REQUEST

## Overview
Refactor GitHub actions into composite actions.  This is the first step towards DRY-ing up our GitHub actions organization wide.

These actions can either live in the repo that they were initially created, or they can be moved to the organization's `.github` repo.  

The goal would be to have one action or one GitHub action for each type to ensure consistency across repos. 

Any repo can now use the same prettier, python, or docker file linting without duplicating the core action code. 

Pending https://github.com/SkynetLabs/.github/pull/6 and https://github.com/SkynetLabs/.github/issues/7

## Checklist
 - [ ] Changelog file created


docker/sia/Dockerfile
```diff
@@ -12,5 +12,6 @@ FROM nebulouslabs/sia:1.5.6
 
 COPY --from=sia-builder /go/bin/ /usr/bin/
 
+# rename binaries to old siac siad naming
```
MSevey, line 15: this was to trigger the dockerfile lint

packages/dashboard/README.md
```diff
@@ -1,3 +1,5 @@
+# Dashboard
```
MSevey, line 1: This was to trigger the prettier lint

packages/health-check/src/checks/critical.js
```diff
@@ -195,6 +195,7 @@ async function blockerHealthCheck(done) {
   done({ name: "blocker", time: calculateElapsedTime(time), ...data });
 }
 
+// genericAccessCheck
```
MSevey, line 198: this was to trigger the dashboard lint

setup-scripts/health-checker.py
```diff
@@ -43,6 +43,7 @@
 setup()
 
 
+# run_checks runs the health checks
```
MSevey, line 46: this was to trigger the python lint

.github/actions/prettier/action.yml
```diff
@@ -0,0 +1,25 @@
...
+    - uses: actions/setup-node@v2
+      with:
+        node-version: 16.x
+    - run: |
+        yarn
+        yarn prettier --check .
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+    - if: inputs.working-directory == 'packages/dashboard-v2' || inputs.working-directory == 'packages/dashboard'
+      run: yarn lint
```
MSevey, line 23: @kwypchlo is this outdated? The action returns the error that no command exists for `yarn lint`. I get the same error locally.

ro-tex:
```
inovakov dashboard-v2 [master]% yarn lint
yarn run v1.22.18
$ eslint .
✨  Done in 7.99s.
```
I had to go into `skynet-webportal/packages/dashboard-v2` and install npm modules but it worked like a charm.

ro-tex (commented):
Looks good, overall. We need to figure out this linting issue.

.github/actions/prettier/action.yml
```diff
@@ -0,0 +1,29 @@
+# Define action name, description, and inputs
+name: "Prettier Lint"
+description: "Lint code with Prettier"
```
meeh0w, line 3: _nit_ - we're using both Prettier and ESLint here, but action name and directory name suggest otherwise :)

.github/workflows/lint.yml
```diff
@@ -0,0 +1,106 @@
...
+    steps:
+      # Use paths-filter to determine which paths have changes
+      # https://github.com/dorny/paths-filter
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          # Format of key: value. When there is a change detected in value, the
+          # key will be returned in the output.
+          filters: |
+            packages/dashboard: packages/dashboard/**
```
meeh0w, line 21: `packages/dashboard-v2` is missing here

meeh0w (commented):
I really like the approach here 👍  Left a couple of small comments.

ro-tex (dismissed):

.github/actions/js-lint/action.yml
```diff
@@ -0,0 +1,29 @@
...
+  working-directory:
+    required: false
+    description: "Working directory"
+    default: "."
+
+# Define action steps
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/setup-node@v2
```
kwypchlo, line 14:
```suggestion
    - uses: actions/setup-node@v3
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,107 @@
...
+          filters: |
+            py: "**.py"
+
+  # Run the hadolint dockefile lint if there are changes to any of the
+  # dockerfiles
+  python-lint:
+    needs: python-changes
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
```
kwypchlo, line 101:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,107 @@
...
+  # Run the js lint if there are changes to any of the js packages
+  js-lint:
+    needs: js-changes
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        working-directory: ${{ fromJSON(needs.js-changes.outputs.packages) }}
+    steps:
+      - uses: actions/checkout@v2
```
kwypchlo, line 59:
```suggestion
      - uses: actions/checkout@v3
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,107 @@
...
+  hadolint:
+    needs: dockerfile-changes
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        dockerfile: ${{ fromJSON(needs.dockerfile-changes.outputs.packages) }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: hadolint/hadolint-action@v2.0.0
```
kwypchlo, line 76:
```suggestion
      - uses: hadolint/hadolint-action@v2.1.0
```

.github/workflows/lint.yml
```diff
@@ -0,0 +1,107 @@
...
+          filters: |
+            packages/dashboard: packages/dashboard/**
+            packages/dashboard-v2: packages/dashboard-v2/**
+            packages/dnslink-api: packages/dnslink-api/**
+            packages/handshake-api: packages/handshake-api/**
+            packages/health-check: packages/health-check/**
+            packages/website: packages/website/**
+
+  # Determine what the dockerfile changes are
+  dockerfile-changes:
```
kwypchlo, line 29: shouldn't we have just one "changes" job that contains multiple filters? Sounds a bit easier to maintain

example: https://github.com/GoogleChrome/web.dev/blob/e1f0c28964e99ce6a996c1e3fd3ee1985a7a04f6/.github/workflows/lint-and-test-workflow.yml#L33-L48

.github/workflows/lint.yml
```diff
@@ -0,0 +1,107 @@
...
+      - uses: hadolint/hadolint-action@v2.0.0
+        with:
+          dockerfile: ${{ matrix.dockerfile }}
+
+  # Determine if there are python changes
+  python-changes:
+    runs-on: ubuntu-latest
+    outputs:
+      # Check if the changed status of py is true
+      updates: steps.changes.outputs.py == 'true'
```
kwypchlo, line 85: I'd like to improve some consistency, sometimes the output is called "packages" sometimes "updates". The id of the step is sometimes called "changes" and sometimes "filter". While those are all valid, I think it makes it easier to read the file if we keep consistent naming, I actually had to read up on some docs to find out whether those are different for a reason.

setup-scripts/health-checker.py
kwypchlo, line 46: Let's remove this from the PR once this is ready to be merged

```suggestion
```

packages/health-check/src/checks/critical.js
kwypchlo, line 198: Let's remove this from the PR once this is ready to be merged

```suggestion
```

packages/dashboard-v2/yarn.lock
```diff
@@ -14090,7 +14090,7 @@ sjcl@^1.0.8:
   resolved "https://registry.yarnpkg.com/sjcl/-/sjcl-1.0.8.tgz#f2ec8d7dc1f0f21b069b8914a41a8f236b0e252a"
   integrity sha512-LzIjEQ0S0DpIgnxMEayM1rq9aGwGRG4OnZhCdjx7glTaJtf4zRfpg87ImfjSJjoW9vKpagd82McDOwbRT5kQKQ==
 
-skynet-js@^4.0.27-beta:
+skynet-js@4.0.27-beta:
```
kwypchlo, line 14093: Let's revert this once this is ready to be merged

```suggestion
skynet-js@^4.0.27-beta:
```

packages/dashboard/README.md
```diff
@@ -1,3 +1,5 @@
...
+
```
kwypchlo, line 2: Let's remove this from the PR once this is ready to be merged, we can put it back later if we need to update the md file but I'd like to keep the changes limited to only required files.

```suggestion
```

docker/sia/Dockerfile
kwypchlo, line 15: Let's remove this from the PR once this is ready to be merged

```suggestion
```

.github/actions/python-flake8-lint/action.yml
```diff
@@ -0,0 +1,22 @@
...
+
+    # E203: https://www.flake8rules.com/rules/E203.html - Whitespace before ':'
+    # E501: https://www.flake8rules.com/rules/E501.html - Line too long
+    # W503: https://www.flake8rules.com/rules/W503.html - Line break occurred before a binary operator
+    # W605: https://www.flake8rules.com/rules/W605.html - Invalid escape sequence
+    # E722: https://www.flake8rules.com/rules/E722.html - Do not use bare except, specify exception instead
+    - run: |
+        pip install flake8
+        flake8 --max-line-length 88 --ignore E203,E501,W503,W605,E722
+      shell: bash
```
kwypchlo, line 22: Let's split the job into separate "run" keys, there's not benefit in keeping this in one (unlike Dockerfile where each layer is cached and costs disk space) and it's easier to catch which step failed.

.github/actions/js-lint/action.yml
```diff
@@ -0,0 +1,29 @@
...
+      with:
+        node-version: 16.x
+    - run: |
+        yarn
+        yarn prettier --check .
+      shell: bash
```
kwypchlo, line 20: Let's split the job into separate "run" keys, there's not benefit in keeping this in one (unlike Dockerfile where each layer is cached and costs disk space) and it's easier to catch which step failed.

.github/actions/python-black-lint/action.yml
```diff
@@ -0,0 +1,17 @@
...
+  steps:
+    - uses: actions/setup-python@v2
+      with:
+        python-version: "3.x"
+        architecture: x64
+
+    - run: |
+        pip install black
+        black --check .
+      shell: bash
```
kwypchlo, line 17: Let's split the job into separate "run" keys, there's not benefit in keeping this in one (unlike Dockerfile where each layer is cached and costs disk space) and it's easier to catch which step failed.

kwypchlo (changes requested):
I've left some comments. Not sure whether this adds readability over separate files but if the motivation is to have single "lint" actions for example for python files, I would put those actions in separate repository like SkynetLabs/github-actions or something because it would be a bit unintuitive for skynet-website (that would probably like to call the js lint job) to reference action from skynet-webportal repo.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,107 @@
...
+        with:
+          # Format of key: value. When there is a change detected in value, the
+          # key will be returned in the output.
+          filters: |
+            packages/dashboard: packages/dashboard/**
+            packages/dashboard-v2: packages/dashboard-v2/**
+            packages/dnslink-api: packages/dnslink-api/**
+            packages/handshake-api: packages/handshake-api/**
+            packages/health-check: packages/health-check/**
+            packages/website: packages/website/**
```
ro-tex, line 26: Is there anything in `packages` that we don't want to check? Would it make sense to have the lint run on `packages/**`, so when we add something we don't need to update this, too?

.github/workflows/lint.yml
```diff
@@ -0,0 +1,107 @@
...
+          # Format of key: value. When there is a change detected in value, the
+          # key will be returned in the output.
+          filters: |
+            docker/nginx/Dockerfile: docker/nginx/Dockerfile
+            docker/sia/Dockerfile: docker/sia/Dockerfile
+            packages/dashboard/Dockerfile: packages/dashboard/Dockerfile
+            packages/dnslink-api/Dockerfile: packages/dnslink-api/Dockerfile
+            packages/handshake-api/Dockerfile: packages/handshake-api/Dockerfile
+            packages/health-check/Dockerfile: packages/health-check/Dockerfile
+            packages/website/Dockerfile: packages/website/Dockerfile
```
ro-tex, line 48: Can we refactor this to
```
dockerfile:
 - '**/Dockerfile'
```
per Karol's example above?

ro-tex (commented):
I agree with Karol's remarks. Otherwise looks good.

.github/workflows/lint.yml
MSevey: that is a question for @kwypchlo 
I simply copied this from the previous linting.
I assume the intent here was to speed up the CI by only running the linting/testing when files were touched.

I would be fine for simplicity to just always run all the linters. It doesn't take long and protects against not updating this in the future. 

Although this lint is specifically prettier and eslint so only really applies to JS. 

The other consideration is that the dashboard packages are the only ones that perform the eslint.

.github/workflows/lint.yml
MSevey: no because the hadolint accepts the path to the dockerfile.

.github/workflows/lint.yml
MSevey: The reason we need the different change blocks is because the linters are expecting different values. If the change block was all one we would need to add more complicated filtering to the action calls.

The JS linters need the package paths for the working directory.

The hadolint needs the dockerfile path.

The python check is the only one that we would be able to mimic that functionality.

If we combined them all we would get an output like this
```
['packages/health-check', 'docker/nginx//Dockerfile', 'py']
```

So we would then need to filter the filter results to properly pass the right arguments to the callers. 

Keeping the changes separate is much simpler.

setup-scripts/health-checker.py
MSevey: These were added to prove that the linters are properly triggered.

packages/health-check/src/checks/critical.js
MSevey: These were added to prove that the linters are properly triggered.

packages/dashboard/README.md
MSevey: These were added to prove that the linters are properly triggered.

docker/sia/Dockerfile
MSevey: These were added to prove that the linters are properly triggered.

meeh0w (dismissed):

.github/workflows/ci_release.yml
```diff
@@ -0,0 +1,174 @@
...
+  release:
+    needs: changes
+    runs-on: ubuntu-latest
+    if: ${{ (needs.changes.outputs.updates == 'true' && github.event_name == 'schedule') || github.event_name == 'workflow_dispatch' }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Version Release
+        uses: SkynetLabs/.github/.github/actions/version-release@master
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
```
kwypchlo, line 174: I don't believe this is a good idea to tag master every night, we should only tag it when we have meaningful changes, I can see bunch of tags that have "fix typo" in the release notes. We also don't have enough code coverage, both unit and integration, to be certain that the release we're tagging is not broken - that's why before deploying to production we're checking the release manually on dev servers and we're supposed to even improve that process to deploy to dev servers even couple of days early to test the potential release. Additionally, if we tag master every night, any code that we will want to revert from master will be a breaking change and accordingly to semver require major version bump. I believe this will harm our processes more that it will help them, my stand is that making a release should be triggered by a human.

kwypchlo (changes requested):

PR #1974, opened 2022-04-06
improve lua stability and test coverage

kwypchlo:
- run lua tests with busted in resty environment
- included codecov coverage report on nginx lua files
- account.lua will be tested in a different pull request


docker/nginx/conf.d/server/server.api
```diff
@@ -228,12 +227,26 @@ location /skynet/skyfile {
 
     # proxy this call to siad endpoint (make sure the ip is correct)
     proxy_pass http://sia:9980/skynet/skyfile/$dir1/$dir2/$dir3$is_args$args;
+
+    log_by_lua_block {
```
peterjan, line 231: worthwhile to dedupe this block through an include?

docker/nginx/libs/utils.spec.lua
```diff
@@ -77,3 +79,125 @@ describe("extract_cookie_value", function()
...
+                assert.is_nil(utils.getenv("foo", "integer"))
+            end)
+
+            it("should parse '0' string as 0", function()
+                os.getenv.on_call_with("foo").returns("0")
+
+                assert.equals(utils.getenv("foo", "integer"), 0)
+            end)
+
+            it("should parse '1' string as 1", function()
```
peterjan, line 184: not sure whether this is valuable to have as a unit test, seems like these are a bit redundant

docker/nginx/libs/utils.spec.lua
```diff
@@ -77,3 +79,125 @@ describe("extract_cookie_value", function()
...
+            end)
+
+            it("should parse '0' string as false", function()
+                os.getenv.on_call_with("foo").returns("0")
+
+                assert.is_false(utils.getenv("foo", "boolean"))
+            end)
+
+            it("should throw an error for not supported string", function()
+                os.getenv.on_call_with("foo").returns("mandalorian")
```
peterjan, line 157: 😂  👌🏻

docker/nginx/libs/utils.lua
```diff
@@ -42,4 +42,42 @@ function _M.extract_cookie_value(cookie_string, name_matcher)
...
+    end
+
+    -- do not parse the value
+    if parse == nil then
+        return value
+    end
+
+    -- try to parse as boolean
+    if parse == "boolean" then
+        if value == "true" or value == "1" then
```
peterjan, line 64: can we lowercase value here, I think we should

docker/nginx/conf.d/server/server.api
kwypchlo: I deliberately squashed the include file because it hides the functionality and the lua block used. It's easier to maintain when I know exactly what blocks are used on a route. Ideally we would have no include files and move most of the business logic to separate lua files that can be unit tested. The idea is that if there is a functionality that can be encapsulated in include file, it should be possible to also encapsulate it in a testable lua function.

docker/nginx/libs/utils.lua
kwypchlo: good catch, we should support "True" and "TRUE"

peterjan (approved):
LGTM 👍  nice effort

docker/nginx/libs/utils.lua
kwypchlo: added in https://github.com/SkynetLabs/skynet-webportal/pull/1974/commits/4e64ddd798fdb629b577a0ade16004ce9acb0c46

docker/nginx/libs/utils.spec.lua
kwypchlo: what I wanted to achieve is to test positive/zero/negative examples

MSevey: if only there was a way to write these types of explanations in the code...

Something like a comment 😜

MSevey:
Nice work 👍🏼

Would love some comments though. It is hard to review 1000+ lines of code that has basically not comments to tell me what I'm looking at.

MSevey (approved):

docker/nginx/libs/utils.spec.lua
kwypchlo: > if only there was a way to write these types of explanations in the code...
> 
> Something like a comment 😜

I'll add a comment on some tests that need explanation in a follow up when writing tests for account.lua

PR #1975, opened 2022-04-06
add skynet-sponsor-key functionality

kwypchlo:
Includes an option to pass a sponsor key for portals requiring authentication to work with dnslink. The sponsor token is a special version of an api key that you can generate in your account dashboard that allows to list specific skylinks that it covers. Once sponsor key is generated, you should add new TXT _dnslink (you will have 2 TXT _dnslink entries now) dns entry to your existing dnslink enabled domain that looks like this `skynet-sponsor-key=<your-sponsor-key-here>`.

Note: exposing sponsor key in dns entry means that it is publicly available, anyone will be able to query dns records and look it up but it will only work with the set of skylinks that you enabled when creating it - also you can always revoke it in your user dashboard panel

### DIFF
--- a/packages/dnslink-api/src/index.js
+++ b/packages/dnslink-api/src/index.js
@@ -8,12 +8,14 @@ const port = Number(process.env.DNSLINK_API_PORT) || 3100;
 const server = express();
 
 const dnslinkNamespace = "skynet-ns";
+const sponsorNamespace = "skynet-sponsor-key";
 const dnslinkRegExp = new RegExp(`^dnslink=/${dnslinkNamespace}/.+$`);
+const sponsorRegExp = new RegExp(`^${sponsorNamespace}=[a-zA-Z0-9]+$`);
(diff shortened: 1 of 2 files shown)


packages/dnslink-api/src/index.js
```diff
@@ -58,9 +61,25 @@ server.get("/dnslink/:name", async (req, res) => {
 
     const skylink = matchSkylink[1];
 
+    // check if _dnslink records contain skynet-sponsor-key entries
+    const sponsors = records.filter((record) => sponsorRegExp.test(record));
+
+    if (sponsors.length > 1) {
```
peterjan, line 67: don't we need to handle the case where the length is 0

peterjan (approved):

packages/dnslink-api/src/index.js
ro-tex: if the length is zero, we just have a non-sponsored skylink. that's perfectly fine, if the user is using a portal that allows anonymous downloads or they are logged in in their portal.

ro-tex (approved):

PR #1981, opened 2022-04-11
refactor(dashboard-v2): move access guards outside of SWRConfig

meeh0w:
## Overview
* Moves guest/logged-in redirection to `UserProvider` which is the only place that keeps the state of the current `user`. It's also the first place to know when the user has logged in or logged out.
* Also fixes a few CSS issues.


peterjan (approved):
LGTM 👍

ro-tex (approved):

PR #1995, opened 2022-04-11
Invite siasky.net users to manage their account via skynetfree.net

meeh0w:
* I realize this isn't perfect, but this version of dashboard will be obsolete soon and we only need to fix it on siasky.net, as currently users get confused, because they are not able to manage their subscription on the portal they subscribed on.

### DIFF
--- a/packages/dashboard/src/pages/auth/recovery.js
+++ b/packages/dashboard/src/pages/auth/recovery.js
@@ -23,6 +23,8 @@ export default function Recovery() {
   useAnonRoute(); // ensure user is not logged in
 
   const [success, setSuccess] = React.useState(false);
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/pages/auth/recovery.js
```diff
@@ -72,6 +84,9 @@ export default function Recovery() {
           validationSchema={validationSchema}
           onSubmit={onSubmit}
           button="Send recovery link"
+          onError={(errorMessage) =>
+            setSkynetFreeInviteVisible(errorMessage === "registrations are currently disabled")
```
meeh0w, line 88: This is the message returned by our API.

dghelm (approved):
LGTM

MSevey (approved):

PR #2010, opened 2022-04-15
pass skylink from request context to /user/limits

kwypchlo:
When `skylink` variable is set in context of request, use it when requesting limits from /user/limits endpoint.

TODO & DISCLAIMER: that logic only works because the skylink is only set in context where skylink is accessed, unfortunately this assumption is be easy to break if someone without knowledge of this changes an endpoint that otherwise doesn't operate in context of skylink (like registry endpoint) and uses skylink variable for something else - this functionality should be rewritten when refactoring nginx account pieces with tests. (issue here https://github.com/SkynetLabs/skynet-webportal/issues/2011 )

### DIFF
--- a/docker/nginx/libs/skynet/account.lua
+++ b/docker/nginx/libs/skynet/account.lua
@@ -76,9 +76,16 @@ function _M.get_account_limits()
 
     if ngx.var.account_limits == "" then
         local httpc = require("resty.http").new()
+        local uri = "http://10.10.10.70:3000/user/limits"
+
+        -- include skylink if it is available in the context of request
+        -- todo: this should not rely on skylink variable to be defined
+        if ngx.var.skylink ~= nil and ngx.var.skylink ~= "" then
+            uri = uri .. "/" .. ngx.var.skylink
+        end
 
         -- 10.10.10.70 points to accounts service (alias not available when using resty-http)
-        local res, err = httpc:request_uri("http://10.10.10.70:3000/user/limits?unit=byte", {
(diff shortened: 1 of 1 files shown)


peterjan (approved):

docker/nginx/libs/skynet/account.lua
```diff
@@ -76,9 +76,16 @@ function _M.get_account_limits()
 
     if ngx.var.account_limits == "" then
         local httpc = require("resty.http").new()
+        local uri = "http://10.10.10.70:3000/user/limits"
+
+        -- include skylink if it is available in the context of request
+        -- todo: this should not rely on skylink variable to be defined
+        if ngx.var.skylink ~= nil and ngx.var.skylink ~= "" then
+            uri = uri .. "/" .. ngx.var.skylink
```
ro-tex, line 84: 1. What will happen if `ngx.var.skylink` is not defined?
2. Can't we just initialise it with "" if it's not initialised? The trailing slash will be stripped in accounts and the result will be the same.
3. You have very interesting notes in the description of this PR. It would be nice to have them in the code as well, so they are readily available to any dev who wants to change this.

Feel free to resolve this and merge right away - none of this is blocking.

ro-tex (approved):

docker/nginx/libs/skynet/account.lua
kwypchlo: I'll follow up with a pull request refactoring this function so skylink is passed as an arg and not implicitly assumed it's defined in a variable.

PR #2012, opened 2022-04-15
Add skylink health check to critical health checks

MSevey:
# PULL REQUEST

## Overview
This PR adds a check to the critical health checks that ensures that the uploaded file reaches full redundancy.

A follow up will be to log these skylinks to a json file or something so that we can do some checks that files are still healthy.

## Checklist
 - [x] Changelog file created

## Issues Closed

Closes SKY-13


packages/health-check/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM node:16.14.2-alpine
 
-RUN apk add --no-cache dnsmasq=2.86-r0
+RUN apk add --no-cache dnsmasq=2.86-r1
```
MSevey, line 3: When testing, I was getting an error trying to build the docker imagine. Bumping this version resolved it.

packages/health-check/src/checks/critical.js
```diff
@@ -63,6 +76,25 @@ async function uploadCheck(done) {
   done({ name: "upload_file", time: calculateElapsedTime(time), ...data });
 }
 
+// skylinkHealthCheck checks if the skylink has reached full redundancy
+async function skylinkHealthCheck(skylink, numRetries, authCookie) {
+  try {
+    const response = await got(`https://${process.env.PORTAL_DOMAIN}/skynet/health/skylink/${skylink}`, {
+      headers: { cookie: authCookie },
```
MSevey, line 83: @kwypchlo how do I pass in a no cache option for this endpoint?

I tested that this code works with I add a manual sleep to get around the caching.

kwypchlo: I don't think there's a nocache option any more with skyd but cc @ChrisSchinnerl

packages/health-check/src/checks/critical.js
```diff
@@ -63,6 +80,25 @@ async function uploadCheck(done) {
...
+    const healthData = getResponseContent(response);
+    if (healthData.basesectorredundancy !== 10 && numRetries > 0) {
+      return skylinkHealthCheck(skylink, numRetries - 1);
+    }
+    if (healthData.basesectorredundancy !== 10 && numRetries === 0) {
+      throw "Skylink did not reach full redundancy";
+    }
+    return response;
+  } catch (error) {
+    throw error;
```
kwypchlo, line 98: if you are just rethrowing the error, you can skip the try-catch block altogether

packages/health-check/src/checks/critical.js
```diff
@@ -63,6 +80,25 @@ async function uploadCheck(done) {
...
 
+// skylinkHealthCheck checks if the skylink has reached full redundancy
+async function skylinkHealthCheck(skylink, numRetries, authCookie) {
+  try {
+    const response = await got(`https://${process.env.PORTAL_DOMAIN}/skynet/health/skylink/${skylink}`, {
+      headers: { cookie: authCookie },
+    });
+    const healthData = getResponseContent(response);
+    if (healthData.basesectorredundancy !== 10 && numRetries > 0) {
+      return skylinkHealthCheck(skylink, numRetries - 1);
```
kwypchlo, line 91:
```suggestion
      return skylinkHealthCheck(skylink, numRetries - 1, authCookie);
```

kwypchlo, line 91: This will run X times without any delay between the calls, we probably want to add some sleep time between calls. You'll have to assume some time constraint we want to give the server for a file to reach full redundancy.

packages/health-check/src/checks/critical.js
```diff
@@ -53,6 +54,22 @@ async function uploadCheck(done) {
...
+    } catch (error) {
+      // Reset the up status as the previous successful file upload would have
+      // set this to true.
+      data.up = false;
+      data.statusCode = error.response?.statusCode || error.statusCode || error.status;
+      // Default to the error itself if the message or response are null since
+      // the error can be a simple error string.
+      data.errorMessage = error.message || error;
+      data.errorResponseContent = getResponseContent(error.response) || error;
+    }
```
kwypchlo, line 72: if you move the `await skylinkHealthCheck(...)` call just below `const response = await got.post(...)` then you can skip all that try-catch block since that's duplicating the outer try-catch block.

kwypchlo (changes requested):

packages/health-check/src/checks/critical.js
MSevey: yea this is what I had done, but I removed it because I was hoping there was a nocache option.

packages/health-check/src/checks/critical.js
MSevey: good to know 👍

packages/health-check/src/checks/critical.js
MSevey: I'll just go with the adding the sleep back in for now.

packages/health-check/src/checks/critical.js
MSevey: adding sleep back in

packages/health-check/src/checks/critical.js
MSevey: so I tested it and I can't target 10.10.10.10 directly. I get 

```js
  {
    name: 'upload_file',
    time: 2321,
    up: false,
    statusCode: 400,
    errorMessage: 'Response code 400 (Bad Request)',
    errorResponseContent: {
      message: 'Browser access disabled due to security vulnerability. Use Sia-UI or siac.'
    },
    ip: '10.10.10.10',
    errors: [
      {
        message: 'Response ip was different than current server ip - possibly there was an error with routing request',
        data: { response: '10.10.10.10', server: '66.29.147.77' }
      }
    ]
  },
```

I think if we can do your option of local traffic avoiding any default caching, that would be good

MSevey: I think we need to figure out the caching issue as it will lead to NDFs in the health checks.

I just tested these on my server and the large file upload passed but the small file upload failed. 

The small file upload got to 9x redundancy so was only sleeping for 1s in between requests so to me it seems like the cache wasn't updating and the test failed.

packages/health-check/src/checks/critical.js
```diff
@@ -60,7 +86,44 @@ async function uploadCheck(done) {
...
+  const response = await got(`https://${process.env.PORTAL_DOMAIN}/skynet/health/skylink/${skylink}`, {
+    headers: { cookie: authCookie },
+  });
+  const healthData = getResponseContent(response);
+
+  // Check Basesectorredundancy first
+  if (healthData.basesectorredundancy !== 10 && numRetries > 0) {
+    // Semi-smart sleep before retrying. Sleep longer if the redundancy is
+    // lower.
+    await new Promise((r) => setTimeout(r, (10 - healthData.basesectorredundancy) * 1000));
```
peterjan, line 104: `nit` it's worth extracting it into a  `sleep` helper imo

packages/health-check/src/checks/critical.js
```diff
@@ -60,7 +86,44 @@ async function uploadCheck(done) {
...
+    return skylinkHealthCheck(skylink, numRetries - 1, authCookie, isLarge);
+  }
+
+  // Check the Fanout redundancy if it is a large file
+  if (isLarge && healthData.fanoutredundancy != 3 && numRetries > 0) {
```
peterjan, line 109: `nit` maybe `3` and `10` could be constants with a meaningful name like `DEFAULT_FANOUT_REDUNDANCY` or whatever

packages/health-check/src/checks/critical.js
```diff
@@ -60,7 +86,44 @@ async function uploadCheck(done) {
...
+    // Semi-smart sleep before retrying. Sleep longer if the redundancy is
+    // lower.
+    await new Promise((r) => setTimeout(r, (3 - healthData.fanoutredundancy) * 10000));
```
peterjan, line 112: did you notice the `10_000` here? maybe extract it into a `SECOND` const

packages/health-check/src/checks/critical.js
```diff
@@ -60,7 +86,44 @@ async function uploadCheck(done) {
...
+async function skylinkHealthCheck(skylink, numRetries = 30, authCookie, isLarge = false) {
+  // Get the health of the skylink
+  const response = await got(`https://${process.env.PORTAL_DOMAIN}/skynet/health/skylink/${skylink}`, {
+    headers: { cookie: authCookie },
+  });
+  const healthData = getResponseContent(response);
+
+  // Check Basesectorredundancy first
+  if (healthData.basesectorredundancy !== 10 && numRetries > 0) {
+    // Semi-smart sleep before retrying. Sleep longer if the redundancy is
```
peterjan, line 102: so 10 factorial is (n^2 + n)/2 or whatever so that's ~55s, is that enough? I guess it should be but it's not a huge amount of time either right

peterjan (commented):
looks good, just think the 10_000 thing might be a mistake so holding off on an approval

packages/health-check/src/checks/critical.js
kwypchlo: > ```js
> Browser access disabled due to security vulnerability. Use Sia-UI or siac.
> ```

This error means that you need to pass "Sia-Agent" in "User-Agent" header. Not sure got allows that but that's what you should try.

packages/health-check/src/checks/critical.js
```diff
@@ -9,6 +10,8 @@ const MODULE_BLOCKER = "b";
 const skynetClient = new SkynetClient(`https://${process.env.PORTAL_DOMAIN}`);
 const exampleSkylink = "AACogzrAimYPG42tDOKhS3lXZD8YvlF8Q8R17afe95iV2Q";
 
+const sectorSize = 1 << 22;
```
kwypchlo, line 13:
```suggestion
const sectorSize = 1 << 22; // 40 MiB
```

packages/health-check/src/checks/critical.js
kwypchlo, line 112: For the large file it is possible that this can take way more than 5 minutes with all the retries and also waiting for the endpoint to return skylink data - the idea for critical checks is that the checks are executed every 5 minutes and should be almost instant so we know how the system operates at the specific point of time and we can pull it out of load balancer right away if there's an issue. If we're considering that a skylink can take 5+ minutes to reach full redundancy and still succeed then we should move this to extended tests probably that run once an hour. Otherwise we should assume a shorter (< 1 minute) total time for the critical test suite to either fail or succeed.

packages/health-check/src/checks/critical.js
```diff
@@ -60,7 +86,44 @@ async function uploadCheck(done) {
...
+  }
+
+  // Throw error if the basesectorredundancy never reached 10x
+  if (healthData.basesectorredundancy !== 10 && numRetries === 0) {
+    throw new Error(`Basesector did not reach full redundancy: ${healthData.basesectorredundancy}`);
+  }
+
+  // Throw error if the fanoutredundancy never reached 3x
+  if (isLarge && healthData.fanoutredundancy !== 3 && numRetries === 0) {
+    throw new Error(`Fanout did not reach full redundancy: ${healthData.fanoutredundancy}`);
```
kwypchlo, line 123:
```suggestion
    throw new Error(`File uploaded but fanout did not reach full redundancy: ${healthData.fanoutredundancy}`);
```

packages/health-check/src/checks/critical.js
```diff
@@ -60,7 +86,44 @@ async function uploadCheck(done) {
...
+  if (isLarge && healthData.fanoutredundancy != 3 && numRetries > 0) {
+    // Semi-smart sleep before retrying. Sleep longer if the redundancy is
+    // lower.
+    await new Promise((r) => setTimeout(r, (3 - healthData.fanoutredundancy) * 10000));
+    return skylinkHealthCheck(skylink, numRetries - 1, authCookie, isLarge);
+  }
+
+  // Throw error if the basesectorredundancy never reached 10x
+  if (healthData.basesectorredundancy !== 10 && numRetries === 0) {
+    throw new Error(`Basesector did not reach full redundancy: ${healthData.basesectorredundancy}`);
```
kwypchlo, line 118:
```suggestion
    throw new Error(`File uploaded but basesector did not reach full redundancy: ${healthData.basesectorredundancy}`);
```

kwypchlo (changes requested):

packages/health-check/src/checks/critical.js
kwypchlo: is the redundancy configurable in skyd? should we pull the configuration values off of skyd and compare to those?

MSevey: it is configurable but we use the defaults for the portal and I can't think of a reason that we would ever not use the defaults on our portals.

applications might choose to use different defaults, but that doesn't apply here. 

I'll update to constants

packages/health-check/src/checks/critical.js
MSevey: I intentionally made it 10,000 since it was going to take longer.

MSevey: a 40MB file shouldn't take more than 5 minutes to upload. If it does, I would argue that the portal isn't operating as expected and should be pulled. 

but i think its find to move this to the extended tests.

kwypchlo: We can leave it in the critical test but we should set the timeout to something like 2 minutes. We should be able to set a time threshold for uploading a 40MB file directly from the server so with no network latency - a minimum performance threshold of some sort.

packages/health-check/src/checks/critical.js
MSevey: It should be. If we are failing this I think is it just cause for us to investigate. Our servers have 1gbps connections and streaming upload should be priority 1 for workers so there shouldn't be anything blocking this from uploading 40MB in 55s.

packages/health-check/src/checks/critical.js
```diff
@@ -36,20 +39,43 @@ async function skydConfigCheck(done) {
...
+
+// uploadLargeFileCheck returns the result of uploading a large file
+async function uploadLargeFileCheck(done) {
+  const payload = Buffer.from(crypto.randomBytes(sectorSize));
+
+  return uploadFunc(done, payload, "upload_large_file", true);
+}
+
+// uploadFunc handles the upload and health check for the upload checks
+async function uploadFunc(done, payload, name, isLarge = false) {
```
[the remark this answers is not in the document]
MSevey: I disagree.  there are 2 redundancy timeouts. so that would be be adding a timeout field. 

the fanout timeout should be larger than the basesector timeout because it is more data to upload. So I don't think we should use the same timeout.

I think it is also good practice to be explicit like these instead of inferring based on the value of something.

packages/health-check/src/utils.js
```diff
@@ -132,6 +154,92 @@ function isPortalModuleEnabled(module) {
...
+  if (healthData.basesectorredundancy !== defaultBaseSectorRedundancy && numRetries === 0) {
+    throw new Error(`File uploaded but basesector did not reach full redundancy: ${healthData.basesectorredundancy}`);
+  }
+
+  // Throw error if the fanoutredundancy never reached 3x
+  if (isLarge && healthData.fanoutredundancy !== defaultFanoutRedundancy && numRetries === 0) {
+    throw new Error(`File uploaded but fanout did not reach full redundancy: ${healthData.fanoutredundancy}`);
+  }
+
+  return response;
```
peterjan, line 199: just a suggestion, rewriting it this way makes it easy to ensure you've covered all possible outcomes

```javascript
const { basesectorredundancy, fanoutredundancy} = healthData
const baseSectorRedundancyReached = basesectorredundancy >= DEFAULT_BASE_SECTOR_REDUNDANCY
const fanoutSectorRedundancyReached = fanoutredundancy >= DEFAULT_FANOUT_REDUNDANCY
const sleepForInS = isLarge
    ? (DEFAULT_FANOUT_REDUNDANCY - fanoutredundancy) * 10;
    : (DEFAULT_BASE_SECTOR_REDUNDANCY - basesectorredundancy) * 10;

const healthy = baseSectorRedundancyReached && (!isLarge || fanoutSectorRedundancyReached)
if (!healthy && attemptsLeft > 0 ) {
    await sleep(sleepForInS)
    return skylinkHealthCheck(skylink, attemptsLeft - 1, authCookie, isLarge);
}

if (!baseSectorRedundancyReached) { throw new Error(...) }
if (!fanoutSectorRedundancyReached) { throw new Error(...) }
```

peterjan (commented):
looks good! almost ready I think

packages/health-check/src/utils.js
```diff
@@ -1,8 +1,30 @@
...
+// sectorSize is the skyd sector size
+const sectorSize = 1 << 22; // 40 MiB
+
+// SECOND is a helper constant for defining the number of milliseconds in a
+// second
+const SECOND = 1000;
+
+// defaultBaseSectorRedundancy is the default baseSectorRedundancy defined by
+// skyd
+const defaultBaseSectorRedundancy = 10;
```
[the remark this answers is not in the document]
ro-tex: should we set up a common eslint configuration which we can reuse in all repos?

packages/health-check/src/utils.js
```diff
@@ -1,8 +1,30 @@
...
+
+// defaultFanoutRedundancy is the default fanout redundancy defined by skyd
+const defaultFanoutRedundancy = 3;
+
+// siaDockerContainerIP is the local IP of the sia docker container
+const siaDockerContainerIP = "10.10.10.10";
+
+// defaultSiaPort is the default port the sia runs on.
+const defaultSiaPort = "9980";
```
ro-tex, line 26: TL;DR: Naming consistency.

Why is one of them "default" and the other one not? When I encountered it in the code it made me think that the `siaContainerIP` could be overridden somehow and that if I want to use the similarly overridden Sia port, I should use not `defaultSiaPort` but `siaPort` or something like that.

packages/health-check/src/utils.js
```diff
@@ -132,6 +154,92 @@ function isPortalModuleEnabled(module) {
...
+  const response = await got(`http://${siaDockerContainerIP}:${defaultSiaPort}/skynet/health/skylink/${skylink}`, {
+    headers: { "User-Agent": "Sia-Agent", cookie: authCookie },
+  });
+  const healthData = getResponseContent(response);
+
+  // Check Basesectorredundancy first
+  if (healthData.basesectorredundancy !== defaultBaseSectorRedundancy && numRetries > 0) {
+    // Semi-smart sleep before retrying. Sleep longer if the redundancy is
+    // lower.
+    let sleepTime = defaultBaseSectorRedundancy - healthData.basesectorredundancy;
```
ro-tex, line 175: nit: 9 seconds sleep when the current redundancy is 1 seems a little excessive.

ChrisSchinnerl (dismissed):

packages/health-check/src/utils.js
```diff
@@ -132,6 +154,92 @@ function isPortalModuleEnabled(module) {
...
 }
 
+// sleep is a helper method of sleeping for for given time. The input time is
+// expected in seconds
+async function sleep(seconds) {
+  return new Promise((r) => setTimeout(r, seconds * SECOND));
+}
+
+// skylinkHealthCheck checks if the skylink has reached full redundancy
+async function skylinkHealthCheck(skylink, numRetries = 30, authCookie, isLarge = false) {
```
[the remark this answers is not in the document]
ChrisSchinnerl: I like `attemptsLeft`. That makes it clearer to understand `attemptsLeft == 0` later because `numRetries == 0` makes it seem like we are in the first iteration.

packages/health-check/src/utils.js
```diff
@@ -132,6 +154,92 @@ function isPortalModuleEnabled(module) {
...
+  if (isLarge && healthData.fanoutredundancy != defaultFanoutRedundancy && numRetries > 0) {
+    // Semi-smart sleep before retrying. Sleep longer if the redundancy is
+    // lower.
+    let sleepTime = (defaultFanoutRedundancy - healthData.fanoutredundancy) * 10;
+    await sleep(sleepTime);
+    return skylinkHealthCheck(skylink, numRetries - 1, authCookie, isLarge);
+  }
+
+  // Throw error if the basesectorredundancy never reached 10x
+  if (healthData.basesectorredundancy !== defaultBaseSectorRedundancy && numRetries === 0) {
```
ChrisSchinnerl, line 190: Same here. I'd check for the redundancy being smaller than the default.

MSevey:
need to revisit and move to new health-checks repo

kwypchlo:
moved this to a new health check repo

PR #2038, opened 2022-04-19
Mongo: ShardingTaskExecutorPoolMinSize=10

ro-tex:
# PULL REQUEST

## Overview

We are experiencing a large number of these errors, referring to different servers:
```json
{
    "t": {
        "$date": "2022-04-19T13:54:53.216+00:00"
    },
    "s": "I",
    "c": "NETWORK",
    "id": 4712102,
    "ctx": "ReplicaSetMonitor-TaskExecutor",
    "msg": "Host failed in replica set",
    "attr": {
        "replicaSet": "skynet",
        "host": "eu-fin-5.siasky.net:27017",
        "error": {
            "code": 202,
            "codeName": "NetworkInterfaceExceededTimeLimit",
            "errmsg": "Couldn't get a connection within the time limit of 649ms"
        },
        "action": {
            "dropConnections": false,
            "requestImmediateCheck": false,
            "outcome": {
                "host": "eu-fin-5.siasky.net:27017",
                "success": false,
                "errorMessage": "NetworkInterfaceExceededTimeLimit: Couldn't get a connection within the time limit of 649ms"
            }
        }
    }
}
```

Potential solution: https://www.mongodb.com/community/forums/t/mongos-couldnt-get-a-connection-within-the-time-limit/3382/5
Official docs: https://www.mongodb.com/docs/manual/reference/parameters/#mongodb-parameter-param.ShardingTaskExecutorPoolMinSize

## Checklist
 - [ ] Changelog file created

### DIFF
--- a/docker-compose.mongodb.yml
+++ b/docker-compose.mongodb.yml
@@ -15,7 +15,7 @@ services:
 
   mongo:
     image: mongo:4.4.1
-    command: --keyFile=/data/mgkey --replSet=${SKYNET_DB_REPLICASET:-skynet}
+    command: --keyFile=/data/mgkey --replSet=${SKYNET_DB_REPLICASET:-skynet} --setParameter ShardingTaskExecutorPoolMinSize=10
     container_name: mongo
     restart: unless-stopped
     logging: *default-logging


docker-compose.mongodb.yml
```diff
@@ -15,7 +15,7 @@ services:
 
   mongo:
     image: mongo:4.4.1
-    command: --keyFile=/data/mgkey --replSet=${SKYNET_DB_REPLICASET:-skynet}
+    command: --keyFile=/data/mgkey --replSet=${SKYNET_DB_REPLICASET:-skynet} --setParameter ShardingTaskExecutorPoolMinSize=10
```
peterjan, line 18: Hm, why did you only add `ShardingTaskExecutorPoolMinSize`, doesn't the thread also suggest to update `ShardingTaskExecutorPoolMaxConnecting`. It feels a bit "dartsy", like we're just trying something to see what sticks.   

In any case I think changes like this warrant a comment in-line to explain why the parameter is set.

ro-tex: Re: `ShardingTaskExecutorPoolMaxConnecting` - I don't quite understand what that option does, to be honest.  

The docs say the following:
```
Maximum number of simultaneous initiating connections (including pending connections in setup/refresh state) each TaskExecutor connection pool can have to a [mongod](https://www.mongodb.com/docs/manual/reference/program/mongod/#mongodb-binary-bin.mongod) instance.
```
That sounds to me as a measure to counter connection depletion and services waiting for initial connection to the cluster. I'm not sure this is relevant? It might be.

> It feels a bit "dartsy", like we're just trying something to see what sticks.
I am afraid that you are spot on. I cannot say with certainty that this will solve the problem.

kwypchlo (approved):

ro-tex:
> I think we should at least make an issue as a follow-up to see whether this change took effect.

Good call. F/U: https://github.com/SkynetLabs/skynet-webportal/issues/2050

> It's undocumented

What do you mean? I linked to the official documentation.  
If you mean that it's undocumented on our end then we should document it. :) What would be a good way to do that?

PR #2040, opened 2022-04-21
use dashboard dockerhub image

kwypchlo:
- use dashboard dockerhub image
- prepare dashboard-v2 Dockerfile for image release
  - we will release it as skynetlabs/dashboard but with major version bumped when we're ready to make a switch

### DIFF
--- a/docker-compose.accounts.yml
+++ b/docker-compose.accounts.yml
@@ -79,7 +79,9 @@ services:
       - mongo
 
   dashboard-v2:
-    build: packages/dashboard-v2/Dockerfile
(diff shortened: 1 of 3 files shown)


peterjan (approved):
should we add semver versioning here like Matt added for all abuse related repos as well?

kwypchlo:
> should we add semver versioning here like Matt added for all abuse related repos as well?

yeah but we'll have to do it when we tag skynet-webportal for next release

ro-tex (approved):

PR #2045, opened 2022-04-22
use multi stage build for website to slim down the image

kwypchlo:
use multi stage build for website to slim down the image

reference: https://github.com/SkynetLabs/skynet-websites/blob/main/Dockerfile


peterjan (approved):
lookin' good

ro-tex (approved):
Nice!

Issue #1870, opened 2022-03-16
Dynamic Report Email Addresses


peterjan:
## Overview

Now that our terms and conditions are webpages, we should make the abuse report email dynamic based on the environment. 

## Design or Proposal

Every mention of `report@siasky.net` should become `report@[environment]`.
Then we should update the LastPass configuration to point at the new inboxes, I will take care of that.

PR #2056, opened 2022-04-29
website: dynamic emails and site metadata

kwypchlo:
- configure website url in site metadata to use PORTAL_DOMAIN
- introduce new env variables for website emails
  - WEBSITE_CONTACT_EMAIL (defaults to hello@PORTAL_DOMAIN)
  - WEBSITE_ABUSE_EMAIL (defaults to report@PORTAL_DOMAIN)
- added a bit of context to emails in the website footer
- necessary change: because we need the website to be aware of env variables, we need to switch from building the static files in docker build step to building and serving when running the service (so it's built with the env vars context), it means we had to move away from multi stage builds and the image will be slightly larger

closes https://github.com/SkynetLabs/skynet-webportal/issues/1870


peterjan (approved):

packages/website/gatsby-config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     description: `Skynet is a decentralized file sharing and content distribution protocol`,
     author: `Skynet Labs`,
     siteUrl: `https://siasky.net`,
```
ro-tex, line 8: Should this be configurable as well?

packages/website/src/services/emails.js
```diff
@@ -0,0 +1,8 @@
+const hostname = typeof window !== "undefined" ? window.location.hostname : "";
+const domain = hostname.substring(hostname.lastIndexOf(".", hostname.lastIndexOf(".") - 1) + 1);
+const emails = {
+  contact: `hello@${domain}`, // general contact email
+  report: `report@${domain}`, // report abuse email
```
ro-tex, line 5: We are making a pretty heavy assumption here that the portal operator will have a certain set of emails configured for their domain. This might or might not be true. While certainly an improvement over the previously hardcoded email address, we can do better. Let's introduce an option for the portal op to customise these emails completely.

ro-tex (changes requested):
The implementation LGTM. The idea doesn't go far enough and if we're touching it at all we should make the final step and deliver a complete solution to the problem.

packages/website/gatsby-config.js
kwypchlo: well yeah but let me make some changes then

kwypchlo: I had to switch back from building static files in image and service them to building locally to use env vars to substitute the site url.

packages/website/src/services/emails.js
kwypchlo: redesigned to use WEBSITE_CONTACT_EMAIL and WEBSITE_ABUSE_EMAIL but default to hello@PORTAL_DOMAIN and report@PORTAL_DOMAIN respectively

ro-tex (approved):

peterjan (approved):

PR #2096, opened 2022-05-10
Add Pinner to the stack.

ro-tex:
# PULL REQUEST

## Overview

This PR adds Pinner to the webportal stack. Its module code is `p`.

## Checklist
 - [ ] Changelog file created

## Issues Closed
Closes SKY-151

### DIFF
--- a/docker-compose.pinner.yml
+++ b/docker-compose.pinner.yml
@@ -17,7 +17,7 @@ services:
     env_file:
       - .env
     environment:
-      - PINNER_LOG_LEVEL=${PINNER_LOG_LEVEL:-trace}
+      - PINNER_LOG_LEVEL=${PINNER_LOG_LEVEL:-info}
     expose:
       - 4000
     networks:
--- a/docker-compose.pinner.yml
+++ b/docker-compose.pinner.yml
@@ -1,0 +1,31 @@
+version: "3.7"
+
+x-logging: &default-logging
+  driver: json-file
+  options:
+    max-size: "10m"
+    max-file: "3"
+
+services:
+  health-check:
+    environment:
+      - BLOCKER_HOST=10.10.10.110
+      - BLOCKER_PORT=4000
+
+  blocker:
+    # uncomment "build" and comment out "image" to build from sources
+    # build: https://github.com/SkynetLabs/pinner.git#main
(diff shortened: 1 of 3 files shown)


docker-compose.pinner.yml
```diff
@@ -0,0 +1,26 @@
...
+  driver: json-file
+  options:
+    max-size: "10m"
+    max-file: "3"
+
+services:
+  pinner:
+    # uncomment "build" and comment out "image" to build from sources
+    # build: https://github.com/SkynetLabs/pinner.git#main
+    image: skynetlabs/pinner
```
kwypchlo, line 13: we should use a tagged version of pinner from the beginning

kwypchlo (changes requested):

docker-compose.pinner.yml
ro-tex: That's great but we don't have one. It seems we'll block on this until Pinner is ready for prod.

docker-compose.pinner.yml
```diff
@@ -0,0 +1,28 @@
...
+    # uncomment "build" and comment out "image" to build from sources
+    # build: https://github.com/SkynetLabs/pinner.git#main
+    image: skynetlabs/pinner:0.1.0
+    container_name: pinner
+    restart: unless-stopped
+    logging: *default-logging
+    env_file:
+      - .env
+    environment:
+      - PINNER_LOG_LEVEL=${PINNER_LOG_LEVEL:-trace}
```
kwypchlo, line 20: is there any specific reason why we're overriding a default log level with trace here?

ro-tex: Yes. I want to default to a lot of output for this service, so we can monitor the way it works. It's both important, untested, and capable of causing a lot of damage, should there be a nasty bug. We can always override this with an env var.

kwypchlo: I would think we would want to do it other way around - set the env var initially to trace instead of defaulting to it for everyone but if that's temporary then I think it's ok. Maybe we can leave a note in the code about it so we remember to tune it down in some time?

kwypchlo (approved):

MSevey (approved):

PR #2099, opened 2022-05-16
update skynet-accounts version

kwypchlo:
Closes SKY-32

### DIFF
--- a/docker-compose.accounts.yml
+++ b/docker-compose.accounts.yml
@@ -22,7 +22,7 @@ services:
   accounts:
     # uncomment "build" and comment out "image" to build from sources
     # build: https://github.com/SkynetLabs/skynet-accounts.git#main
-    image: skynetlabs/skynet-accounts
+    image: skynetlabs/skynet-accounts:1.1.0
     container_name: accounts
     restart: unless-stopped
     logging: *default-logging


ro-tex (approved):

docker-compose.accounts.yml
```diff
@@ -22,7 +22,7 @@ services:
   accounts:
     # uncomment "build" and comment out "image" to build from sources
     # build: https://github.com/SkynetLabs/skynet-accounts.git#main
-    image: skynetlabs/skynet-accounts
+    image: skynetlabs/skynet-accounts:1.1.0
```
MSevey, line 25: Didn't we want to just use the Major and Minor version?

Wasn't that one of the main reasons for the docker action?

MSevey:
Closes SKY-32

docker-compose.accounts.yml
ro-tex: I'm open to using major.minor. This will allow us to fix bugs in patch versions, which is why they exist in the first place. And those will get auto-deployed on restart. There is some risk involved but also a lot of flexibility. This will allow us to fix critical issues without cutting a new deployment.

kwypchlo: I'm just afraid that we're gonna end up with broken release at some point and we're gonna have to either wait for a bug fix with broken servers or revert to a specific patch version. We use specific version of go and node images in our dockerfiles, I think the motivation for that is similar.

kwypchlo: In any case, we're already setting full version on other repos like blocker or malware-scanner so let's go ahead with full version here and follow up with a discussion on whether we should use just major.minor version or major.minor.path version.

ro-tex: SGTM.

MSevey (approved):

PR #2166, opened 2022-07-05
Define a log file for pinner.

ro-tex:
# PULL REQUEST

## Checklist

- [ ] All git commits are signed. (REQUIRED)
- [ ] All new methods or updated methods have clear docstrings.
- [ ] Testing added or updated for new methods.
- [ ] Verified if any changes impact the WebPortal Health Checks.
- [ ] Appropriate documentation updated.
- [ ] Changelog file created.

### DIFF
--- a/docker-compose.pinner.yml
+++ b/docker-compose.pinner.yml
@@ -16,8 +16,11 @@ services:
     logging: *default-logging
     env_file:
       - .env
+    volumes:
+      - ./docker/data/pinner/logs:/logs
     environment:
       - PINNER_LOG_LEVEL=${PINNER_LOG_LEVEL:-info}
+      - PINNER_LOG_FILE=${PINNER_LOG_FILE}
     expose:
       - 4000
     networks:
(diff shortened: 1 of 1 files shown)


docker-compose.pinner.yml
```diff
@@ -16,8 +16,11 @@ services:
     logging: *default-logging
     env_file:
       - .env
+    volumes:
+      - ./docker/data/pinner/logs:/logs
     environment:
       - PINNER_LOG_LEVEL=${PINNER_LOG_LEVEL:-info}
+      - PINNER_LOG_FILE=${PINNER_LOG_FILE}
```
kwypchlo, line 23: if you're not assigning any defaults then this line is redundant as this service loads `.env` anyway

kwypchlo (approved):

docker-compose.pinner.yml
ro-tex: you're right, I could have left it out. anyway, I'll just let it be, even if it's not strictly necessary.

ro-tex: you're right. I thought about leaving it there but I'd rather start cleaning these files up. accounts has a whole bunch of these sitting around for no reason. we shouldn't do the same here. removed. :)

kwypchlo (approved):

meeh0w (approved):

PR #2171, opened 2022-07-06
use remote sia dockerfile until ansible is ready to use image

kwypchlo:
Before we switch to using actual dockerhub image for skyd, we need ansible adjustments but that work that might not be done for a while due to priorities. In the meantime, instead of using a local dockerfile to build sia, we can start using the one from docker-skyd repository that will eventually become the official docker skyd image.


docker-compose.yml
```diff
@@ -16,16 +16,16 @@ networks:
 services:
   sia:
     build:
-      context: ./docker/sia
-      dockerfile: Dockerfile
+      context: https://github.com/SkynetLabs/docker-skyd.git#main
+      dockerfile: scratch/Dockerfile
       args:
         branch: portal-latest
+    command: --disable-api-security --api-addr :9980 --modules gctwra
```
meeh0w, line 23: What is `a`? It's not documented [here](https://github.com/NebulousLabs/Sia/blob/master/cmd/siad/main.go#L73-L131)

meeh0w (approved):

docker-compose.yml
ro-tex: It's accounting. Documented here: https://gitlab.com/SkynetLabs/skyd/-/blob/master/cmd/skyd/main.go#L148

The repo on GitHub is out of date, the main repo is on GitLab. We need to setup auto-sync.

PR #2179, opened 2022-07-28
upgrade to handshakeorg/hsd:4.0.2

kwypchlo:
- switch to official [handshakeorg/hsd](https://hub.docker.com/r/handshakeorg/hsd) image
- bumps hsd to v4 and adds chain migration option
- disables wallet module since we're not using it
- disables api key auth (we do not expose the api publicly)
- enables outdated tree data compaction on startup (saves around 15G disk space)

tested on siasky.xyz

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -122,17 +122,11 @@ services:
       - 9000
 
   handshake:
-    image: skynetlabs/hsd:3.0.1
-    command: --chain-migrate=2 --wallet-migrate=1
+    image: handshakeorg/hsd:4.0.2
+    command: --chain-migrate=3 --no-wallet --compact-tree-on-init --network=main --http-host=0.0.0.0
     container_name: handshake
     restart: unless-stopped
     logging: *default-logging
-    environment:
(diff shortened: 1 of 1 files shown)


meeh0w:
@kwypchlo I think it entered a crashloop on siasky.xyz and keeps reporting these:

```
handshake           | Error: Database version mismatch for database: "chain". Please run a data migration before opening.
```

or is there something else going on on that server that could cause this?

kwypchlo:
> siasky.xyz seems to be back to v3 and is reporting migration errors right now, I guess because chain is now on a version higher than it was before:
> 
> ```
> handshake           | Error: Database version mismatch for database: "chain". Please run a data migration before opening.
> ```
> 
> So I built it locally and tested with latest version of `handshake-api` and it is working fine 🎉

yeah I forgot I deployed it there and tried to run current deploy and it downgraded and failed, it's important to note that once we do this upgrade we can't roll back without resyncing the handshake blockchain

peterjan (approved):

PR #2192, opened 2022-08-09
remove outdated blocklist scripts

kwypchlo:
Now that there is a [blocker](https://github.com/SkynetLabs/blocker) service integrated in portal stack, we do not have use for outdated airtable script and all hashes are already in skyd and if portal includes blocker module, they are in blocker database too.

- ~~drop manual blocklist script~~ keeping those
- drop airtable blocklist script


scripts/README.md
```diff
@@ -4,10 +4,6 @@ This package contains useful scripts for managing a Skynet Webportal.
 
 ## Available Scripts
 
-**blocklist-skylink.sh**\
-The `blocklist-skylink.sh` script adds a skylink to the blocklist on all
-servers.
-
```
ro-tex, line 10: I don't think we want to remove this. I know we can use blocker but looking at this from the perspective of someone unfamiliar with the stack and working under pressure, I'd rather give them more than one tool, so they have a higher chance of finding it quickly.

kwypchlo: reverted changes and removed only the airtable scripts

ro-tex (approved):

ChrisSchinnerl (approved):

PR #2202, opened 2022-08-11
allow relative call to dc script

kwypchlo:
allows calling `dc` from any directory, not only from the root directory of project

**before**
`cd /home/user/skynet-webportal && ./dc` - works
`cd /home/user && ./skynet-webportal/dc` - does not work
`cd /home/user && /home/user/skynet-webportal/dc` - does not work

**after**
`cd /home/user/skynet-webportal && ./dc` - works
`cd /home/user && ./skynet-webportal/dc` - works
`cd /home/user && /home/user/skynet-webportal/dc` - works

this fix is useful for future ansible development

### DIFF
--- a/dc
+++ b/dc
@@ -5,56 +5,59 @@
 # would use docker-compose with the only difference being that you don't need to specify compose files. For more
 # information you can run `./dc` or `./dc help`.
 
-if [ -f .env ]; then
+# get current working directory of this script
+cwd="$(dirname -- "$0";)";
+
+if [ -f ${cwd}/.env ]; then
   OLD_IFS=$IFS
   IFS=$'\n'
-  for x in $(grep -v '^#.*' .env); do export $x; done
+  for x in $(grep -v '^#.*' ${cwd}/.env); do export $x; done
   IFS=$OLD_IFS
 fi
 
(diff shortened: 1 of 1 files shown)


dc
```diff
@@ -5,56 +5,59 @@
...
 
-if [ -f .env ]; then
+# get current working directory of this script
+cwd="$(dirname -- "$0";)";
+
+if [ -f ${cwd}/.env ]; then
   OLD_IFS=$IFS
   IFS=$'\n'
-  for x in $(grep -v '^#.*' .env); do export $x; done
+  for x in $(grep -v '^#.*' ${cwd}/.env); do export $x; done
```
ro-tex, line 14: We need a comment explaining what exactly is happening here. I, personally, have no idea and future maintainers might not have either.

MSevey (dismissed):

dc
kwypchlo: added more info in https://github.com/SkynetLabs/skynet-webportal/pull/2202/commits/063f0b0b7bcf483553fa2dc88e6e351ea631598b

ro-tex (approved):

MSevey (approved):

PR #2206, opened 2022-08-16
do not report from dc when env file does not exist

kwypchlo:
dc should not report error when .env does not exist (it just logs error, doesn't really break the script anyway), docker-compose will report on that and exit with proper error code

before
```
➜  skynet-webportal git:(master) ✗ ./dc down
grep: ./.env: No such file or directory
open /Users/karol/sia/skynet-webportal/.env: no such file or directory
```

after
```
➜  skynet-webportal git:(master) ✗ ./dc down
open /Users/karol/sia/skynet-webportal/.env: no such file or directory
```

### DIFF
--- a/dc
+++ b/dc
@@ -11,7 +11,9 @@
 cwd="$(dirname -- "$0";)";
 
 # get portal modules configuration from .env file (if defined more than once, the last one is used)
-PORTAL_MODULES=$(grep -e "^PORTAL_MODULES=" ${cwd}/.env | tail -1 | sed "s/PORTAL_MODULES=//")
+if [[ -f "${cwd}/.env" ]]; then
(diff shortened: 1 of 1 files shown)


ro-tex:
Is that a good idea? I would definitely prefer it to alert me if that file is not there. If `dc` runs with default setup instead of the one I expect from it that can cause a bunch of problems.

kwypchlo:
> Is that a good idea? I would definitely prefer it to alert me if that file is not there. If `dc` runs with default setup instead of the one I expect from it that can cause a bunch of problems.

the error is thrown anyway, the stack will not start/stop since the services rely on the .env file (see the after error)

the only change is that grep in dc will not print the error, docker-compose (bottom of the file) will exit with code 14 and print the proper error

ChrisSchinnerl (approved):

ro-tex (approved):

